PR #7, opened 2015-01-15
kola: an automated test framework for CoreOS

marineam:
Work in progress: This is merely a snapshot of current development and is not yet a proper test suite but I want to submit this before I get too far along. The biggest change that must be made is refactor platform/local and platform/qemu to move to a cluster based abstraction, similar to what is provided in the fleet functional test suite that inspired this code to begin with.
    
https://github.com/coreos/fleet/blob/v0.9.0/functional/platform/cluster.go
    
Error handling and unit tests are also incomplete.

At this point this is all it does:
```
sudo ./kola qemu --qemu.image=$HOME/latest-usr/coreos_production_image.bin
Password: 
dnsmasq[15261]: started, version 2.72 cachesize 150
dnsmasq[15261]: compile time options: IPv6 GNU-getopt DBus i18n IDN DHCP DHCPv6 no-scripts TFTP no-conntrack ipset no-auth no-DNSSEC loop-detect
dnsmasq[15261]: warning: no upstream servers configured
dnsmasq-dhcp[15261]: DHCP, static leases only on 10.2.0.1, lease time 1h
dnsmasq-dhcp[15261]: DHCP, static leases only on 10.1.0.1, lease time 1h
dnsmasq-dhcp[15261]: DHCP, static leases only on 10.0.0.1, lease time 1h
dnsmasq-dhcp[15261]: DHCPv4-derived IPv6 names on fd02::
dnsmasq-dhcp[15261]: router advertisement on fd02::
dnsmasq-dhcp[15261]: DHCPv4-derived IPv6 names on fd01::
dnsmasq-dhcp[15261]: router advertisement on fd01::
dnsmasq-dhcp[15261]: DHCPv4-derived IPv6 names on fd00::
dnsmasq-dhcp[15261]: router advertisement on fd00::
dnsmasq-dhcp[15261]: IPv6 router advertisement enabled
dnsmasq[15261]: cleared cache
SSH: Linux localhost 3.18.2 #2 SMP Thu Jan 8 21:23:05 PST 2015 x86_64 Intel(R) Core(TM) i7-4960HQ CPU @ 2.60GHz GenuineIntel GNU/Linux

QEMU successful!
```


marineam:
Major update, does the exact same thing but has introduced a cluster level interface.

network/ssh.go
```diff
@@ -0,0 +1,137 @@
...
+// An interface for anything compatible with net.Dialer
+type Dialer interface {
+	Dial(network, address string) (net.Conn, error)
+}
+
+// SSHAgent can manage keys, updates cloud config, and loves ponies.
+// The embedded dialer is used for establishing new SSH connections.
+type SSHAgent struct {
+	agent.Agent
+	Dialer
```
crawford: Why network.Dialer instead of net.Dialer?

network/ssh.go
```diff
@@ -0,0 +1,137 @@
...
+
+	for _, key := range keys {
+		cfg.SSHAuthorizedKeys = append(cfg.SSHAuthorizedKeys, key.String())
+	}
+
+	return nil
+}
+
+// Add port to host if not already set.
+func addDefaultPort(host string, port int) string {
```
crawford: This is a strange name given that it doesn't use a default.

network/ssh.go
```diff
@@ -0,0 +1,137 @@
...
+	if host[0] == '[' && host[i-1] == ']' {
+		return host
+	}
+	if strings.Contains(host[:i], ":") {
+		return fmt.Sprintf("[%s]:%d", host, port)
+	}
+
+	// Address seems invalid, just leave it alone.
+	return host
+}
```
crawford: What about something like:

```go
switch {
  case !strings.Contains(host, ":"):
    return fmt.Sprintf("%s:%d", host, port)
  case strings.HasPrefix(host, ":"):
    return host
  case strings.HasPrefix(host, "[") && strings.Contains(host, "]:"):
    return host
  case strings.Count(host, ":") > 1:
    return fmt.Sprintf("[%s]:%d", host, port)
  default:
    return host
}
```

network/ssh.go
```diff
@@ -0,0 +1,137 @@
...
+// agent forwarding but it must also be enabled per-session.
+func (a *SSHAgent) NewClient(host string) (*ssh.Client, error) {
+	sshcfg := ssh.ClientConfig{
+		User: a.User,
+		Auth: []ssh.AuthMethod{
+			ssh.PublicKeysCallback(a.Signers),
+		},
+	}
+
+	addr := addDefaultPort(host, 22)
```
crawford: const defaultPort = 22?

network/ssh.go
marineam: network.Dialer adds support for retrying on failure. A helpful property when waiting for a machine to boot.

network/ssh.go
marineam: Well, meaning "set this default if a port isn't already specified in the address". Got a better name?

platform/local/cluster.go
```diff
@@ -0,0 +1,146 @@
...
+
+	origns, err := netns.Get()
+	if err != nil {
+		return nil, err
+	}
+	defer netns.Set(origns)
+
+	lc := &LocalCluster{}
+	lc.nshandle, err = netns.New()
+	if err != nil {
```
crawford: these `if`s can be inlined. it's your call though. I don't feel strongly about it.

network/ssh.go
marineam: Not quite, in the current logic it is significant that i is the index of the last `:` char but I will see what something similar to your suggestion would look like.

marineam: Updated, slightly harder to read due to the extra text but the actual logic is clearer this way.

network/ssh.go
marineam: done

platform/qemu.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	defer session.Close()
+
+	session.Stderr = os.Stderr
+	out, err := session.Output(cmd)
+	out = bytes.TrimSpace(out)
+	return out, err
+}
+
+func (qm *qemuMachine) Destroy() error {
```
crawford: Is it safe to remove the return type? It doesn't seem to be used anywhere.

platform/local/cluster.go
marineam: Generally I didn't inline multi-value assignments in order to keep the usage of `=` and `:=` consistent.

network/ssh.go
crawford: But its the exact same interface as `net.Dialer`. I think it makes sense to either embed a `RetryDialer` (if you want to enforce that) or `net.Dialer`.

platform/qemu.go
marineam: I intend to eventually return an error, it just requires extra logic to figure out what error to return. Kill failing may be ok if the process is already dead, Wait reporting "killed" as an error shouldn't be returned since we meant to kill it blah blah blah. I plan on figuring out some reasonable logging scheme and improved error handling in a second pass.

network/ssh.go
crawford: ensurePortSuffix?

network/ssh.go
marineam: There is also a NsDialer that may be used so an interface must be embedded. net.Dialer is not an interface despite the name seeming like it should be, it is a struct.

network/ssh.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return nil
+}
+
+// Add port to host if not already set.
+func addDefaultPort(host string, port int) string {
+	switch {
+	case !strings.Contains(host, ":"):
+		return fmt.Sprintf("%s:%d", host, port)
+	case strings.HasPrefix(host, "[") && strings.HasSuffix(host, "]"):
```
crawford: nice. that's way cleaner.

network/ssh.go
marineam: hm, ensureHasPort?

network/ssh.go
crawford: oh wow. yeah, my eyes skipped right over that...

crawford:
LGTM

network/ssh.go
marineam: went with ensurePortSuffix

PR #10, opened 2015-02-13
platform/local: consolidate namespace entering code

marineam:
Creating and switching namespaces, which are local to the OS thread, is
rather awkward in Go. Implement the goo once and fix a file descriptor
leak in the process, previously origns was never being closed.

### DIFF
--- a/platform/local/nsenter.go
+++ b/platform/local/nsenter.go
@@ -1,0 +1,64 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


platform/local/nsenter.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil, err
+	}
+
+	err = netns.Set(ns)
+	if err != nil {
+		origns.Close()
+		return nil, err
+	}
+
+	return func() error {
```
crawford: That's a clever way to defer from another scope.

crawford:
LGTM

pbx0:
LGTM

PR #15, opened 2015-02-18
Bootchart command for kola

marineam:
Find out what slows down coreos:

```sh
sudo ./kola bootchart --qemu.image=$HOME/latest/coreos_production_image.bin > chart.svg
firefox ./chart.svg
```


crawford:
LGTM

PR #22, opened 2015-03-06
Support for running many etcd tests in kola

pbx0:
/cc @marineam

### DIFF
--- a/cmd/kola/etcd.go
+++ b/cmd/kola/etcd.go
@@ -1,0 +1,108 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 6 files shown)


marineam:
The `"$public_ipv4"` goo is included in the StartJournal commit but the change log doesn't mention it.

pbx0:
Thanks for the feedback, I updated the branch with the changes. I seem to have squashed most of the pull request's comments in the process. 

/cc @marineam @crawford

marineam:
[LGTM](http://www.lgtm.in/i/8Y28PluXp)

marineam:
btw, on the comments, that's just one awkward bit about PRs, if comments are made on individual commits like we did then replacing those commits similarly drops the comments from the PR. Commenting on the big diff in the PR will keep the comments associated but then the reviewer cannot view the diff split into the logical chunks commits provide. No particular rule to do things one way or the other but something to keep in mind.

pbx0:
Updated branch with @crawford 's comments. The go-routine responsible for journal output will be exit when the machine does. A stop mechanism could be added in the future.

crawford:
LGTM

PR #23, opened 2015-03-11
kola: add design document

pbx0:
/cc @marineam


marineam:
[LGTM](http://www.lgtm.in/i/EWhMNV4Qp)

PR #24, opened 2015-03-18
kola: add basic NFS server/client test

mischief:
cc @pbx0 @marineam

### DIFF
--- a/cmd/kola/nfs.go
+++ b/cmd/kola/nfs.go
@@ -1,0 +1,195 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"path"
+	"time"
+
+	"github.com/coreos/mantle/cli"
+	"github.com/coreos/mantle/platform"
+
+	"github.com/coreos/mantle/Godeps/_workspace/src/github.com/coreos/coreos-cloudinit/config"
+)
+
+func init() {
+	cli.Register(cmdNfs)
+}
+
+var cmdNfs = &cli.Command{
+	Run:     runNfs,
+	Name:    "nfs",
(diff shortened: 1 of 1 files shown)


cmd/kola/nfs.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+	}()
+
+	log.Printf("NFS client booted.")
+
+	var lsmnt []byte
+
+	log.Printf("Waiting for NFS mount on client...")
+
+	/* there's probably a better wait to check the mount */
```
pbx0: Looks fine to me. Qemu platform works similarly: https://github.com/coreos/mantle/blob/master/platform/qemu.go#L158

pbx0:
LGTM

cmd/kola/nfs.go
```diff
@@ -0,0 +1,205 @@
...
+		if err := m1.Destroy(); err != nil {
+			log.Printf("Machine.Destroy: %s", err)
+			return
+		}
+	}()
+
+	log.Printf("NFS server booted.")
+
+	/* poke a file in /tmp */
+	rnd := make([]byte, 8)
```
crawford: Should this be `make([]byte, 0, 8)`?

cmd/kola/nfs.go
```diff
@@ -0,0 +1,205 @@
...
+	"bytes"
+	"crypto/rand"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/coreos/mantle/cli"
+	"github.com/coreos/mantle/platform"
+
+	"github.com/coreos/coreos-cloudinit/config"
```
marineam: This should refer to the Godep copy of cloudinit

mischief: Ah, but of course. I will fix it shortly.

cmd/kola/nfs.go
marineam: At some point I would like to add support to go-systemd to interact with the dbus API tunneled over SSH (systemctl supports this in combination with systemd-stdio-bridge). But that is a fair amount of work and not a priority. So busy waits will work just dandy for now. Later we can watch the status of units instead.

cmd/kola/nfs.go
```diff
@@ -0,0 +1,205 @@
...
+
+	/* poke a file in /tmp */
+	rnd := make([]byte, 8)
+	_, err = rand.Read(rnd)
+	if err != nil {
+		log.Printf("rand.Read: %s", err)
+		return 2
+	}
+
+	_, err = m1.SSH(fmt.Sprintf("sudo touch /tmp/%X", rnd))
```
marineam: Why not just call mktemp? Also, this is /tmp, sudo is not needed here.

mischief:
PTAL. i think i addressed all concerns. `[]byte(path.Base(string(tmp)))` is sure ugly though.

marineam:
[LGTM](http://www.lgtm.in/i/gyhTHfL3q)

PR #25, opened 2015-03-21
cmd/kola: add multiple etcd discovery tests and better vm cleanup

pbx0:
Add multiple discovery tests for different etcd versions and configurations. Requires image build with both versions: export ETCD_PROTOCOLS="1 2"

/cc @marineam

### DIFF
--- a/cmd/kola/etcd.go
+++ b/cmd/kola/etcd.go
@@ -33,7 +33,8 @@ var cmdEtcd = &cli.Command{
 
 Work in progress: the code sets up an etcd cluster and then calls
 integration tests to see how etcd interacts with new CoreOS images.
-Currently discovery using another etcd cluster is supported for 0.4.7
+Currently, a local discovery service is available through an embedded
+etcd server.
 
 This must run as root!
 `}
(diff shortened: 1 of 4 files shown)


cmd/kola/etcd.go
```diff
@@ -48,51 +48,74 @@ func runEtcd(args []string) int {
 		fmt.Fprintf(os.Stderr, "No args accepted\n")
 		return 2
 	}
+
+	var exit int
+	var machines []platform.Machine
+TestLoop:
 	for _, t := range etcdtests.Tests {
+		// kill any machines from the last test instead of waiting for
```
marineam: How about splitting the body of this loop into a new function. That way you can leave it to a `defer cluster.Destroy()` and clean up a lot of the complexity of this code in general.

cmd/kola/etcdtests/etcdDiscovery.go
```diff
@@ -22,29 +24,40 @@ func etcdDiscovery(cluster platform.Cluster) error {
 	for i, m := range cluster.Machines() {
 		// start etcd instance
 		etcdStart := "sudo systemctl start etcd.service"
-		_, err := m.SSH(etcdStart)
+		//etcdStart := "etcd version"
```
marineam: left over debugging?

platform/local/cluster.go
```diff
@@ -102,7 +102,7 @@ func (lc *LocalCluster) GetDiscoveryURL(size int) (string, error) {
 	client := &http.Client{Transport: tr}
 
 	body := strings.NewReader(url.Values{"value": {strconv.Itoa(size)}}.Encode())
-	req, err := http.NewRequest("PUT", URL+"_config/size", body)
+	req, err := http.NewRequest("PUT", URL+"/_config/size", body)
```
marineam: Use `path.Join` should be used to avoid similar errors in the future.

cmd/kola/etcdtests/etcdDiscovery.go
```diff
@@ -22,29 +24,40 @@ func etcdDiscovery(cluster platform.Cluster) error {
...
+	for i := 0; i < curlRetries; i++ {
+		time.Sleep(time.Second)
+		cmd := cluster.NewCommand("curl", "http://10.0.0.2:4001/v2/machines")
+		b, err := cmd.Output()
+		if err != nil {
+			continue
+		}
+		members = strings.Split(string(b), ",")
+		// etcd 2.0 listens on 2 ports for clients and returne 2*csize
+		if len(members) == csize || len(members) == csize*2 {
```
marineam: It would be more robust to compare the list of members to the list of machines.

marineam:
Overall this is better but I think it deserves another cleanup pass. As a general trend loops seem more complicated than they need to be. Breaking up into functions more should help.

platform/local/cluster.go
pbx0: path.Join mangles the double-slash in the URL though.

marineam: Ah, yeah, and then more complicated forms of URL construction are likely harder to read. So this is fine then.

cmd/kola/etcdtests/etcdDiscovery.go
pbx0: Curling `/v2/machines` seemed to be the only version agnostic way to easily see that the cluster had formed correctly, but is slightly ugly in that 2.0.x normally returns a list that is double the cluster size.  curling `/v2/members` is 2.0.x only.

I'll check that `etcdctl cluster-health` returns 3 healthy members instead.

marineam: I was thinking along the lines of parsing the output of machines to verify the hosts you expected were there. The check could just ignore duplicated hosts on different ports because it is skimming for IP or host id.

pbx0:
PTAL

I did some cleanup in etcd.go and etcdDiscovery.go Checking for cluster success now utilizes `etcdctl cluster-health` which should be more robust.

platform/local/cluster.go
philips: Uppercase variable names are generally reserved for exported variables, not local function variables, please call it something like `base` or something.

cmd/kola/etcd.go
```diff
@@ -33,7 +33,7 @@ var cmdEtcd = &cli.Command{
 
 Work in progress: the code sets up an etcd cluster and then calls
 integration tests to see how etcd interacts with new CoreOS images.
-Currently discovery using another etcd cluster is supported for 0.4.7
+Currently discovery using an embedded etcd server is supported.
```
philips: This sentence doesn't parse for me. I think it is OK if "is supported" is removed and a comma is added after "Currently".

cmd/kola/etcd.go
```diff
@@ -48,51 +48,53 @@ func runEtcd(args []string) int {
...
-		cluster, err := platform.NewQemuCluster()
+		err := runTest(t)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "Cluster failed: %v\n", err)
+			fmt.Fprintf(os.Stderr, "%v failed: %v\n", t.Name, err)
 			return 1
 		}
+		fmt.Printf("Test %v ran successfully\n", t.Name)
+	}
+	fmt.Fprintf(os.Stderr, "Etcd tests ran successfully!\n")
```
philips: Etcd should always be lowercase

cmd/kola/etcdtests/etcdDiscovery.go
```diff
@@ -22,30 +22,47 @@ func etcdDiscovery(cluster platform.Cluster) error {
 	for i, m := range cluster.Machines() {
 		// start etcd instance
 		etcdStart := "sudo systemctl start etcd.service"
-		_, err := m.SSH(etcdStart)
+		b, err := m.SSH(etcdStart)
 		if err != nil {
 			return fmt.Errorf("SSH cmd to %v failed: %s", m.IP(), err)
 		}
-		fmt.Fprintf(os.Stderr, "etcd instance%d started\n", i)
+		fmt.Fprintf(os.Stderr, "etcd instance%d started%s\n", i, b)
```
philips: space after started?

pbx0:
PTAL

Fixed comments from @philips and squashed commits into one. The second commit re-fixed the vm cleanup loop code (ugly) in the first commit which doesn't need to be in the history.

cmd/kola/etcd.go
```diff
@@ -48,51 +49,53 @@ func runEtcd(args []string) int {
...
+			return fmt.Errorf("Cluster failed starting machine: %v", err)
 		}
-
-		for _, m := range cluster.Machines() {
-			err = m.Destroy()
-			if err != nil {
-				fmt.Fprintf(os.Stderr, "Destroy failed: %v\n", err)
-				return 1
+		defer func() {
+			if err := m.Destroy(); err != nil {
```
marineam: cluster.Destroy cleans up machines so this shouldn't be required.

marineam:
Code is looking much better now, once you address the two remaining comments this should be good to go.

pbx0:
Fixed up last two comments and merging

PR #26, opened 2015-03-23
dnsmasq: advertise NTP via DHCP

marineam:
mantle doesn't actually provide a ntp server but this is still useful
for checking whether the DHCP options are getting plumbed through
systemd-networkd to systemd-timesyncd.


mischief:
lgtm

PR #28, opened 2015-03-24
Add etcd fallback and migration test

pbx0:
/cc @philips @marineam

### DIFF
--- a/cmd/kola/etcdtests/util.go
+++ b/cmd/kola/etcdtests/util.go
@@ -85,7 +85,6 @@ func replaceEtcdBin(cluster platform.Cluster, oldv, newv string) error {
 	// escape file path forward-slashes for sed command
 	oldv = strings.Replace(oldv, "/", `\/`, -1)
 	newv = strings.Replace(newv, "/", `\/`, -1)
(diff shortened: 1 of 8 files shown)


cmd/kola/etcdtests/util.go
```diff
@@ -0,0 +1,158 @@
...
+// replace default binary for etcd.service with internal version v
+func replaceEtcdBin(cluster platform.Cluster, oldv, newv string) error {
+	// escape file path forward-slashes for sed command
+	oldv = strings.Replace(oldv, "/", `\/`, -1)
+	newv = strings.Replace(newv, "/", `\/`, -1)
+
+	for _, m := range cluster.Machines() {
+		// first err collector
+		var err error
+		firstErr := func(b []byte, e error) {
```
marineam: I've used this pattern in order to perform all cleanup operations even if one of them failed instead of coming up with a scheme to return all the errors if more than one thing failed.

In this case there isn't any reason to continue execution if something fails so the standard pattern of wrapping each in an if is more appropriate.

philips:
this should be closed, right?

pbx0:
Closing since these tests don't represent current migration process.

PR #30, opened 2015-03-25
kola: output logs from all etcd nodes

marineam:
Logs from a single node aren't enough to identify why intermittent
failures are happening in the cluster as a whole.


pbx0:
LGTM

PR #32, opened 2015-03-31
auth: fix license comment so it is not displayed in generated documentation

mischief:


mischief:
see http://godoc.org/github.com/coreos/mantle/auth

pbx0:
LGTM

PR #34, opened 2015-04-01
remove deprecated goauth2 library and use latest oauth2

pbx0:


auth/google.go
```diff
@@ -15,49 +15,101 @@
...
+	"encoding/json"
 	"fmt"
+	"log"
 	"net/http"
+	"os"
 	"os/user"
 	"path/filepath"
 
-	"github.com/coreos/mantle/Godeps/_workspace/src/code.google.com/p/goauth2/oauth"
+	"golang.org/x/oauth2"
```
marineam: You need to run `godep save -r ./...` to rewrite this path and vendor the new oauth2 library under the Godeps directory.

mischief:
lgtm

marineam:
@pbx0 there must be something crufty in your environment, old built library installed in your gopath maybe? I re-ran godep and got this: https://github.com/coreos/mantle/compare/coreos:master...marineam:pbx0-plume?expand=1

pbx0:
@marineam PTAL

auth/google.go
```diff
@@ -15,49 +15,99 @@
...
 	if err != nil {
 		return nil, err
 	}
 
-	cachePath := filepath.Join(userInfo.HomeDir, ".gsextra-google.json")
-	transport := oauth.Transport{Config: &config}
-	transport.TokenCache = oauth.CacheFile(cachePath)
-	if debug {
-		transport.Transport = &logTransport{http.DefaultTransport}
+	cachePath := filepath.Join(userInfo.HomeDir, ".tokencache-google.json")
```
marineam: `.mantle-google.json` is probably better so someone has a better chance of figuring out where that file came from.

marineam:
LGTM, only request is to rename from gsextra to mantle instead of tokencache. (gsextra was the name of the project I originally wrote this code in)

PR #36, opened 2015-04-03
cmd/plume: upload cmd also creates GCE image

pbx0:
This unifies uploading to Google Storage and creating an image in GCE
into one step. The Google Storage filename is mapped into an image name
for simplicity.

### DIFF
--- a/auth/google.go
+++ b/auth/google.go
@@ -35,7 +35,8 @@ var conf = oauth2.Config{
 		TokenURL: "https://accounts.google.com/o/oauth2/token",
 	},
 	RedirectURL: "urn:ietf:wg:oauth:2.0:oob",
-	Scopes:      []string{"https://www.googleapis.com/auth/devstorage.full_control"},
+	Scopes: []string{"https://www.googleapis.com/auth/devstorage.full_control",
+		"https://www.googleapis.com/auth/compute"},
 }
 
 func writeCache(cachePath string, tok *oauth2.Token) error {
@@ -76,9 +77,10 @@ func getToken() (*oauth2.Token, error) {
 	if err != nil {
 		log.Printf("Error reading google token cache file: %v", err)
 	}
-	if tok == nil {
+	if tok == nil || !tok.Valid() {
 		url := conf.AuthCodeURL("state", oauth2.AccessTypeOffline)
 		fmt.Printf("Visit the URL for the auth dialog: %v\n", url)
+		fmt.Print("Enter token: ")
 
 		var code string
 		if _, err := fmt.Scan(&code); err != nil {
--- a/cmd/plume/upload.go
+++ b/cmd/plume/upload.go
@@ -29,6 +29,7 @@ import (
 	"github.com/coreos/mantle/Godeps/_workspace/src/google.golang.org/cloud/storage"
 	"github.com/coreos/mantle/auth"
 	"github.com/coreos/mantle/cli"
+	"google.golang.org/api/compute/v1"
 )
 
 var (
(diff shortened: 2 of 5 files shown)


auth/google.go
```diff
@@ -76,9 +77,10 @@ func getToken() (*oauth2.Token, error) {
 	if err != nil {
 		log.Printf("Error reading google token cache file: %v", err)
 	}
-	if tok == nil {
+	if tok == nil || !tok.Valid() {
 		url := conf.AuthCodeURL("state", oauth2.AccessTypeOffline)
 		fmt.Printf("Visit the URL for the auth dialog: %v\n", url)
+		fmt.Println("Enter token: ")
```
marineam: Should this be `Print` so the input remains on the same line?

cmd/plume/upload.go
```diff
@@ -86,8 +86,13 @@ func runUpload(args []string) int {
 		fmt.Fprintf(os.Stderr, "URL missing bucket name %v\n", bucket)
 		return 1
 	}
+
 	bucket = gsURL.Host
+	// add prefix
```
marineam: This comment seems a bit pointless.

cmd/plume/upload.go
```diff
@@ -86,8 +86,13 @@ func runUpload(args []string) int {
...
 	imageName = strings.TrimPrefix(gsURL.Path+"/"+imageName, "/")
+	// create equivalent image names for GS and GCE
+	imageNameGCE := gceSanitize(imageName)
```
marineam: Although this scheme including the full gsURL is simple for now something like the following would be cleaner: `$USER-$BOARD-$imageName`

cmd/plume/upload.go
```diff
@@ -50,8 +51,8 @@ var (
 func init() {
 	cmdUpload.Flags.StringVar(&bucket, "bucket", "gs://coreos-plume", "gs://bucket/prefix/")
```
marineam: For reference the existing default for `--upload` in build_image and similar is: `gs://users.developer.core-os.net/$USER/boards/$BOARD/$VERSION` which may be a useful pattern to stick with.

pbx0: If we make the default `gs://users.developer.core-os.net/$USER/$BOARD/$VERSION` instead then the translation to a GCE name doesn't have to handle the special case for default paths vs user-specified paths.

pbx0:
PTAL. GS now defaults to `gs://users.developer.core-os.net/$USER/$BOARD/$VERSION` and the image name is `$USER-$BOARD-$imageName`

cmd/plume/upload.go
```diff
@@ -95,17 +110,76 @@ func runUpload(args []string) int {
 		return 1
 	}
 
-	fmt.Printf("Writing %v to %v\n", imageName, bucket)
+	fmt.Printf("Writing %v to gs://%v ...\n", imageNameGS, bucket)
+	fmt.Printf("(Sometimes this takes a few mintues)\n")
+
+	/*
```
marineam: Left over debugging?

cmd/plume/upload.go
```diff
@@ -126,17 +200,20 @@ func getImageVersion(imagePath string) string {
 	return version
 }
 
-func writeFile(client *http.Client, filename string) error {
+// Write file to Google Storage
+func writeFile(client *http.Client, filename, destname, projec, projectID string) error {
 	ctx := cloud.NewContext(projectID, client)
```
marineam: Why is projectID needed? The upload API only needs a project ID to create a new bucket. Is this a quirk of the 'cloud' library? All of my work has been on the direct API bindings.

pbx0: Yeah, just seems to be a quick in the cloud library: https://godoc.org/google.golang.org/cloud/storage

cmd/plume/upload.go
pbx0: Recently uploading to GS has gone from going quite fast to taking several minutes. It eventually succeeds.I figured it is important that user's know its not broken when its taking 5 or more minutes to upload ~100-200MB.

marineam: Well, ok. but it shouldn't be commented out in the PR...

cmd/plume/upload.go
marineam: So it gets embedded in the context object but nothing in the storage api uses it since cloud/storage doesn't implement creating buckets. I'd say re-use the gce project name here or use an empty string or whatever dummy value. No point in passing around unused data, could lead to surprises as things change over time.

cmd/plume/upload.go
pbx0: Doh! Yes, this is left over debugging. I was looking at the wrong line. I thought you were talking about the taking a few minute line for some reason.

pbx0:
PTAL. Removed the debugging that slipped in and removed need to specify GS project.

marineam:
LGTM

PR #37, opened 2015-04-08
kola: remove migration from roadmap

philips:
We are no longer doing the etcd2 migration, remove it from the roadmap.


marineam:
LGTM

PR #38, opened 2015-04-08
Gce plume

pbx0:
This adds the all the basic commands necessary to create and destroy clusters on GCE. 

Most commands have extra optional parameters, but relying on default flag usage looks like
`plume upload`
`plume create-instances -image=<imagename> -n 3`
`plume list-instances`
`plume destroy-instances -prefix=<$USER>`
`plume list-images`

`create-instances` would in most cases specify a cloud-config filepath.

I prefixed the subcommand flag variables to avoid confusion and collision. Alternative approaches would be moving each subcommand into a separate package (but still having to register them in `main`) or reworking the way the CLI works.

gce.go will eventually be migrated into platform for use with kola as well.

### DIFF
--- a/cmd/plume/create.go
+++ b/cmd/plume/create.go
@@ -1,0 +1,110 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/coreos/mantle/auth"
+	"github.com/coreos/mantle/cli"
+)
+
+var (
+	cmdCreate = &cli.Command{
+
+		Name:        "create-instances",
+		Summary:     "Create cluster on GCE",
+		Usage:       "-image <gce image name> -n <number of instances>",
+		Description: "Upload os image to Google Storage bucket and create image in GCE",
+		Flags:       *flag.NewFlagSet("create", flag.ExitOnError),
+		Run:         runCreate,
+	}
+	createProject      string
+	createZone         string
+	createMachine      string
+	createBaseName     string
+	createImageName    string
+	createConfig       string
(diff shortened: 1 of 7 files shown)


cmd/plume/create.go
```diff
@@ -0,0 +1,136 @@
...
+	for _, vm := range vms {
+		fmt.Printf("To access %v use cmd:\n", vm.name)
+		fmt.Printf("ssh -o UserKnownHostsFile=/dev/null -o CheckHostIP=no -o StrictHostKeyChecking=no core@%v\n", vm.extIP)
+	}
+
+	return 0
+}
+
+// nextName returns the next available numbered name or the given base name.
+// snipped from: https://github.com/golang/build/blob/master/cmd/gomote/list.go
```
marineam: As such perhaps fair to include the origional copyright in the header?

cmd/plume/gce.go
```diff
@@ -0,0 +1,232 @@
...
+		if !strings.HasPrefix(image.Name, prefix) {
+			continue
+		}
+		images = append(images, image.Name)
+	}
+	return images, nil
+}
+
+// create gce VM and wait for creation to succeed. Some code snipped from:
+// https://github.com/golang/build/blob/master/buildlet/gce.go#L323
```
marineam: This is certainly big enough to deserve preserving the copyright notice.

marineam:
Looks reasonable, my only concern is that there is enough copied code that the original copyright notice should be preserved, as required by the BSD license the code was released under.

philips:
cc @kelseyhightower

cmd/plume/create.go
pbx0: So just add the Go author copyright up with ours? Should I still mention which functions have the code and a link to where it is from?

marineam: Yes, add the same go author copyright to the top. Also keep the existing
comments/links. That is useful to know.

pbx0:
Added the Go author copyright header and BSD license under the the file LICENSE-BSD next to our apache license file.

marineam:
Relicensing BSD code under Apache-2 is perfectly fine so you just need to preserve the copyright line. Having both license headers in the file is just confusing.

pbx0:
BSD license removed.

cmd/plume/gce.go
```diff
@@ -0,0 +1,260 @@
+// Copyright 2015 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
```
marineam: This is the bit I was referring to about having two license headers. Just need something like:

```
// Copyright 2015 CoreOS, Inc.
// Copyright 2015 The Go Authors.
//
// Licensed under the Apache License.....
```

pbx0: Makes more sense, fixed.

marineam:
LGTM

PR #39, opened 2015-04-13
platform: factor plume GCE functionality into platform

pbx0:
The platform package interface has been updated to accomodate an options
struct to handle all the options that cloud providers will need when
creating a new machine.


mischief:
since you have changes the signature of Cluster.NewMachine, can you update the other callers in kola?

pbx0:
Yes, fixed the signature for kola nfs and bootchart tests.

marineam:
What about instead of MachineOpts the machine startup gets split from object initialization. That could be useful for a few things:

 - In tests that need to manually configure an etcd cluster or similar with IP addresses it can create three machines, then update eachotehr's cloud config appropriately with the other's IPs.
 - Add additional network interfaces before starting (for QEMU)
 - Add cloud specific options before starting (for GCE)

That way there doesn't need to be a global structure that can do everything ever. This probably means that each cluster and machine struct should become public. Then the platform specific APIs become available via a type assertion.

Open to other ideas but that seems like a more flexible API

pbx0:
The platform interface is reverted for now and default options are set in flags within gce.go. Offline discussion decided to punt on the api for handling platform specific options for the time being (and use flags). The option struct is kept around for now as well since its useful when using platform from plume which has its own defaults set from flags.

In the future, its possible some of these options may go into the cluster creation commands since they are already platform specific.

marineam:
[LGTM](http://www.lgtm.in/i/Y6usWGDln)

PR #40, opened 2015-04-16
Full gce support for kola

pbx0:
Kola now has a platform flag that the etcd2 tests are aware of.

`sudo kola etcd`
and
`kola etcd -platform=gce -gce.image=pb-amd64-usr-653-0-0-2015-04-15-1702`

Will successfully run etcd (1 and 2) discovery on qemu and gce.


marineam:
nice!! LGTM

PR #42, opened 2015-04-21
README: Add docs for plume gce commands

pbx0:


README.md
```diff
@@ -7,3 +7,51 @@ This repository is a collection of utilities for the CoreOS SDK.
...
+
+`plume list-instances`
+
+### plume destroy-instances
+
+Destroy instances on gce based by prefix or name. Images created with
+`create-instances` use a common basename as a prefix that can also be
+used to tear down the cluster. Common usage:
+
+`plume destroy-instances -prefix=<$USER>`
```
marineam: I'd leave off the < and > here since you are using a var instead of a text description. Otherwise < and > start to look like part of the command.

README.md
```diff
@@ -7,3 +7,51 @@ This repository is a collection of utilities for the CoreOS SDK.
...
+Print out gce images from a project. Common usage:
+
+`plume list-images`
+
+### plume create-instances
+
+Launch instances on gce. SSH keys should be added to the metadata
+section of the gce developers console. Common usage:
+
+`plume create-instances -n=3 -image=<gce image name>`
```
marineam: Do you need to specify prefix here as well?

pbx0: No, in this case the basename is just the image name unless an alternative basename is supplied. If the default image naming is also used, then all instances from that image will have a common prefix of `$USER`

pbx0:
Added `-config=<cloud config file>` as a common usage of `create-instances` since, though its not required, i'm sure it represents the most common usage. Also, `<$USER>` -> `$USER`

marineam:
LGTM

marineam:
@robszumski Can you help point @pbx0 in the right direction for how to import this into our SDK docs on the website?

robszumski:
Does this need to go _into_ the SDK guide or is this a new document?

marineam:
@robszumski new doc, with time I probably want to re-org the sdk docs a bit too but for starters this is just something new.

robszumski:
@pbx0 Create a PR against https://github.com/coreos/docs/tree/master/sdk-distributors/sdk and follow the same style as the rest of the docs -- its probably easiest to start from a copy of an existing doc and change the title, etc.

pbx0:
@robszumski https://github.com/coreos/docs/pull/486

PR #43, opened 2015-04-23
flag cleanup

pbx0:
Global flags in platform were polluting the help options in plume since plume sets its own flags in subcommands. To fix this for now, the global flags go into the main kola package itself. The platform package gets access to the flags from an options struct already used by plume. 

Kola tests nfs and bootchart are also platform aware and run on both without issue.


marineam:
LGTM

PR #46, opened 2015-05-01
Harness

pbx0:
Factor the test harness out into a separate package.


kola/harness.go
```diff
@@ -20,53 +20,57 @@ import (
...
+	Platforms   []string // whitelist of platforms to run test against -- defaults to all
 }
 
 // runs and sets up environment for tests specified in etcdtests pkg
-func runEtcd(args []string) int {
-	if len(args) != 0 {
-		fmt.Fprintf(os.Stderr, "No args accepted\n")
+func RunTests(args []string) int {
+	if len(args) != 1 {
+		fmt.Fprintf(os.Stderr, "No test group specified. Use 'all' to run all tests\n")
```
marineam: Why not just make 'all' the default?

kola/harness.go
```diff
@@ -20,53 +20,57 @@ import (
...
-
-This must run as root!
-`}
-
-func init() {
-	cli.Register(cmdEtcd)
+type Test struct {
+	Run         func(platform.Cluster) error
+	Name        string
+	Group       string // used to manually group related tests
```
marineam: Other test suites often use name pattern matching and/or some sort of tag scheme such as one indicating the execution time like 'long' vs 'short'. We could probably combine Group and Platforms into a generic "tags" scheme. If you want to simplify this PR you can just drop Group for now but add pattern matching on test names.

kola/tests/exampleTest
```diff
@@ -0,0 +1,46 @@
...
+package example
+
+import "github.com/coreos/mantle/harness"
+
+// This is necessary somewhere in the package to register all tests you wish to make available
+// for kola to run.
+func init() {
+	harness.Tests = append(harness.Tests, []harness.Test{
+		Test{
+			Name:        "versionCheck",
```
marineam: Lets name 'Name' practical as a global name, useful to the end user. Something like "EtcdVersionCheck" would work.

kola/tests/exampleTest
```diff
@@ -0,0 +1,46 @@
...
+			Group:    "version",
+			// This will group sets of tests together to be run by a
+			// kola subcommand. It'd be nice to do it automatically by
+			// packages, but I think that would require codegen.
+		},
+
+		// Multiple tests configuations can be specified for a single
+		// run function. Each trial gets a new cluster.
+		Test{
+			Name:        "versionCheck",
```
marineam: Names should be unique, otherwise what's the purpose of a name?

kola/tests/exampleTest
```diff
@@ -0,0 +1,46 @@
...
+			Run:         versionCheck,
+			Discovery:   true,
+			ClusterSize: 5,
+			CloudConfig: `#cloud-config...`,
+			Group:    "version",
```
marineam: I don't see how a group name like this is useful. Just drop it please.

kola/etcdregistry.go
```diff
@@ -12,37 +12,36 @@
...
-		Discovery:   true,
-		ClusterSize: 3,
-		Name:        "etcd2 discovery",
-		EtcdVersion: 2,
-		CloudConfig: `#cloud-config
+		// test etcd discovery with 2.0 with new cloud config
+		Test{
+			Run:         etcd.DiscoveryV2,
+			ClusterSize: 3,
+			Name:        "etcd2 discovery",
```
marineam: Don't use spaces in test names, that'll make it easier to specify them as a command line option.

kola/harness.go
pbx0: Sounds good for now. Can pattern matching be as simple as `kola run name-substring` that matches any test whose name contains it? Or would something like `kola run regex-pattern`be better?

marineam: Lets do globbing, that is easier to do on command line than regex. So kola run 'Etcd*' to do etcd tests. Also lets make test names start with uppercase. Later if we add tags like 'long' or 'short' they can be lower case so kola run short isn't ambiguous.

pbx0:
PTAL:
All tests run by default. Groups are gone and names are patterned matched for the time being. Test names should be unique and uppercase. Deleted example test for now. Examples should just go in documentation.

kola/flags.go
```diff
@@ -0,0 +1,33 @@
+package kola
```
marineam: Missing license header

marineam:
LGTM

PR #47, opened 2015-05-07
minor updates to test harness and build fix

pbx0:


cmd/kola/bootchart.go
```diff
@@ -37,6 +39,8 @@ init process.
 This must run as root!
 `}
 
+var kolaPlatform = flag.String("platform", "qemu", "compute platform for bootchart")
```
marineam: Seems like `kola.go` is more appropriate for common options.

marineam:
LGTM

cmd/kola/bootchart.go
pbx0: Sounds good. (Bootchart is just happens to be the only command that uses the flag anymore)

PR #48, opened 2015-05-10
Build fixes and enable Travis

marineam:
Need to start catching our stupid build errors :-P

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: go
+sudo: false
+
+# Gracefully handle developer forks
+install: |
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: go
```
crawford: Do you want to pin this to a particular version of golang?

pbx0:
LGTM

.travis.yml
marineam: Eh, maybe. Will find out when 1.5 comes out :-P

Issue #6, opened 2015-01-14
kola: run coretest equivilant


marineam:

marineam:
Note: coretest itself does things like check external network connectivity which is not applicable to kola, instead we just need to make sure we cover anything useful that coretest does.

mischief:
@marineam can this be closed since we have coretests ported now?

philips:
next step: we need to re-add the fleet tests.

mischief:
what's the current problem here?

what's blocking re-enabling of the fleet tests?

```
$ sudo ./bin/kola --qemu-image /home/mischief/work/coreos-dist/808.0.0/coreos_production_image.bin run coretestsCluster
2015-09-18T20:54:59Z kola: === RUN coretestsCluster on qemu
2015-09-18T20:55:00Z etcdserver: added local member f3620cbbbe7cd162 [http://localhost:0] to cluster 3bd8c6452814d712
2015-09-18T20:55:02Z etcdserver: set the initial cluster version to 2.1.0
2015-09-18T20:56:10Z kola/tests/coretest: running EtcdUpdateValue...
2015-09-18T20:56:11Z kola/tests/coretest: running FleetctlListMachines...
2015-09-18T20:56:11Z kola/tests/coretest: running FleetctlRunService...
2015-09-18T20:56:17Z kola: --- PASS: coretestsCluster on qemu (78.571s)
2015-09-18T20:56:17Z kola: 1 passed 0 failed out of 1 total
```

mischief:
tests also passed on aws against alpha 808 pv in us-west-1.

pbx0:
Still had tests fail on kam, fix is here: https://github.com/coreos/mantle/pull/115

PR #49, opened 2015-05-18
cmd/kolalet: add support for running native go test code

pbx0:
Kolalet binaries are dropped onto newly created machines by the test
harness. Tests are now registered as a Group. Each group gets its own
cluster and will sequentially run a list of cluster wide tests that have
access to the platform.Cluster interface. A list of native test
functions are then run on each node of the cluster via the kolalet
binary.

This commit will be used to port the coretest series of tests. https://github.com/coreos/mantle/issues/6

Right now kolalet binaries are just expected to be in the working dir from which kola is run. After the coretest work I'll followup with a PR into overlay to get the kolalet binary built into the SDK.

### DIFF
--- a/cmd/kolalet/kolalet.go
+++ b/cmd/kolalet/kolalet.go
@@ -1,0 +1,71 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/coreos/mantle/cli"
+	"github.com/coreos/mantle/kola"
+)
+
+const (
+	cliName        = "kolalet"
+	cliDescription = "Native code runner for kola"
+	// http://en.wikipedia.org/wiki/Kola_Superdeep_Borehole
+)
+
+// main test harness
+var cmdRun = &cli.Command{
+	Name:    "run",
+	Summary: "Run native tests a group at a time",
+	Run:     Run,
+}
+
+func init() {
+	cli.Register(cmdRun)
+}
+
+func main() {
+	cli.Run(cliName, cliDescription)
+}
+
+// test runner
+func Run(args []string) int {
+	if len(args) != 1 {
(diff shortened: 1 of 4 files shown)


cmd/kolalet/kolalet.go
```diff
@@ -0,0 +1,71 @@
...
+	"os"
+
+	"github.com/coreos/mantle/cli"
+	"github.com/coreos/mantle/kola"
+)
+
+const (
+	cliName        = "kolalet"
+	cliDescription = "Native code runner for kola"
+	// http://en.wikipedia.org/wiki/Kola_Superdeep_Borehole
```
marineam: Heh, don't need this comment here.

kola/harness.go
```diff
@@ -24,18 +24,32 @@ import (
...
-	Run         func(platform.Cluster) error
-	Name        string //Should be uppercase and unique
-	CloudConfig string
-	ClusterSize int
-	Platforms   []string // whitelist of platforms to run test against -- defaults to all
+// Defines a group of tests to run on a cluster instance. These tests are
+// repeated on a new cluster for each platform. Cluster test functions are run
+// first using the platform.Cluster interface. Then the series of native code
+// tests are run on each member of the cluster.
+type TestGroup struct {
```
marineam: Could you split adding TestGroup and adding NativeTests to it into two commits please?

marineam:
So what about the case of a cluster wide test needing a local helper? My thinking would be "tests" would still be defined as they were but could optionally invoke kolalet on one or more machines do do some or all of the dirty work, possibly with some arguments. That would be more flexible in the long term.

Also, kolalet sounds awkward, what about kolet? :)

pbx0:
In the case of a cluster wide test needing a local helper I was thinking that adding something like `KoletRun(funcname) error` to the platform interface would work, but I didn't need it yet.

While porting coretest I used the `TestGroup` concept because it seemed the tests fit well into 3 groups (https://gist.github.com/pbx0/a0cca9d05f3a1425799c) whose organization and contents is very visible in the registration file. This way small tests could all be run on a 1-node cluster without spinning up a new cluster on each platform for each test.

So the same abstraction provided  by `TestGroup` could also be achieved with a `Test` that simply calls `KoletRun()`on group of helpers or small tests. This maybe isn't quite as nice as a TestGroup in the use case of simply porting coretest, but maybe its better then having both abstractions and some overlap in functionality.

So I think the cluster wide tests having a local helper makes sense and could replace TestGroups like you suggested, but I just want to confirm.

Also, kolet sounds good.

marineam:
I like the idea of a Test that calls KoletRun(), it may be verbose but it will look the same as more complicated tests that use Kolet. Whatever the API is complicated tests will want to run Kolet on a specific host with specific args while the simple case should have some helper so tests that only want to run one command on one machine are a single line in the function body.

marineam:
And again, things will get less verbose if we assemble some `go test`-like introspection scheme so simple tests can literally be a single function definition and nothing more. So I'm not too concerned about boilerplate glue right now.

pbx0:
Moving this over to: https://github.com/coreos/mantle/pull/50

PR #50, opened 2015-05-22
kolet: add support for native code in kola tests

pbx0:
Kolet binaries are dropped onto newly created machines by the test harness and allow for running native helper or test functions within a test.

Tests can now define native functions it would like to have available. These tests can be run through the ugly closure passed along with the cluster. This closure is used over a struct + method to prevent an import cycle.

This functionality is useful in porting coretest into kola in future PRs.

Right now kolalet binaries are just expected to be in the working dir from which kola is run. Once kolet is in overlay kolet binaries will be in a standard location.

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -24,9 +24,6 @@ import (
 	"github.com/coreos/mantle/platform"
 )
 
-// tests should register by adding init() functions to this package
-var Tests []Test
-
 type Test struct {
 	Run         func(platform.Cluster) error
 	Name        string //Should be uppercase and unique
@@ -35,7 +32,19 @@ type Test struct {
 	Platforms   []string // whitelist of platforms to run test against -- defaults to all
 }
 
-// runs and sets up environment for tests specified in etcdtests pkg
+// maps names to tests
+var Tests map[string]*Test
+
+// panic if existing name is registered
+func Register(t *Test) {
+	_, ok := Tests[t.Name]
+	if ok {
+		panic("test already registered with same name")
+	}
+	Tests[t.Name] = t
+}
+
+// test runner
 func RunTests(args []string) int {
 	if len(args) > 1 {
 		fmt.Fprintf(os.Stderr, "Extra arguements specified. Usage: 'kola run [glob pattern]'\n")
@@ -48,7 +57,7 @@ func RunTests(args []string) int {
 		pattern = "*" // run all tests by default
 	}
 
-	var ranTests int //count sucessful tests
+	var ranTests int //count successful tests
 	for _, t := range Tests {
 		match, err := filepath.Match(pattern, t.Name)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


kola/harness.go
```diff
@@ -112,11 +128,60 @@ func runTest(t Test, pltfrm string) (err error) {
 		fmt.Fprintf(os.Stderr, "%v instance up\n", pltfrm)
 	}
 
+	// drop kolet binary on machines
+	for _, m := range cluster.Machines() {
```
marineam: Perhaps skip if NativeFuncs is empty?

kola/harness.go
```diff
@@ -112,11 +128,60 @@ func runTest(t Test, pltfrm string) (err error) {
...
+		return fmt.Errorf("Error establishing ssh session: %v", err)
+	}
+	defer session.Close()
+
+	// machine reads file from stdin
+	session.Stdin = in
+
+	// cat file to fs
+	_, filename := filepath.Split(src)
+	_, err = session.CombinedOutput(fmt.Sprintf("cat > ./%s", filename))
```
marineam: One trick I've used before is `"install -m 0755 /dev/stdin /dest/path"` and similar so you can set the mode at the same time.

kola/tests/etcd/discovery.go
```diff
@@ -23,12 +23,12 @@ import (
 	"github.com/coreos/mantle/platform"
 )
 
-func DiscoveryV2(cluster platform.Cluster) error {
-	return discovery(cluster, 2)
+func DiscoveryV2(c platform.Cluster, _ func(string, platform.Machine) error) error {
```
marineam: How about define that func as a type to minimize the copy/pasting of this signature all over the place.

pbx0: That'd be much nicer but I can't figure out how to do that without causing an import cycle.

The kola package requires tests call `Register()` inside the kola package to glue in the separate test packages. If I pass a named type from kola in as an argument in the Test.Run function then the tests will have a signature like `func Tester(c platform.Cluster, nr kola.NativeRunner) error {}` when defined outside the kola package causing the import cycle.

I could name the type as a platform.NativeRunner. Does that sound reasonable?

marineam: Or a small kola/native package?

pbx0:
PTAL -- Integrated comments. The function to run a native function now lives in platform and is passed along to a Test as a  TestCluster wrapping cluster. I think going back to a named closure but in kola/native would be fine too. I left it in platform because I thought TestCluster might have other uses for Cluster helper functions that don't have any platform specific implementation aspects to it.

platform/platform.go
```diff
@@ -38,3 +40,20 @@ type Cluster interface {
...
+// methods.
+type TestCluster struct {
+	Name string
+	Cluster
+}
+
+// run a registered NativeFunc on a remote machine
+func (t *TestCluster) RunNative(funcName string, m Machine) error {
+	// scp and execute kolet on remote machine
+	b, err := m.SSH(fmt.Sprintf("./kolet run \"%v\" \"%v\"", t.Name, funcName))
```
marineam: how about `%q` instead of manually escaping?

marineam:
Just one nit, and looks good to me! :)

platform/platform.go
pbx0: Nifty!

vcaputo: btw, instead of escaping quotes in Go you can use ``` `grave-enclosed "string literals"` ```

PR #51, opened 2015-05-27
port coretest suite into kola

pbx0:


kola/tests/coretest/cloudinit.go
```diff
@@ -0,0 +1,125 @@
...
+
+const cloudinitBinPath = "/usr/bin/coreos-cloudinit"
+
+func read(filename string) (string, error) {
+	bytes, err := ioutil.ReadFile(filename)
+	return string(bytes), err
+}
+
+func rmdir(path string) error {
+	cmd := exec.Command("sudo", "rm", "-rf", path)
```
mischief: why not `os.RemoveAll` here?

kola/tests/coretest/cloudinit.go
```diff
@@ -0,0 +1,125 @@
...
+ssh_authorized_keys:
+    - %s
+    - %s
+`
+	configData := fmt.Sprintf(configTmpl, keyOne, keyTwo)
+	configFile, err := ioutil.TempFile(os.TempDir(), "coretest-")
+	if err != nil {
+		return fmt.Errorf("Failed creating tempfile: %v", err)
+	}
+	defer syscall.Unlink(configFile.Name())
```
mischief: why not `os.Remove` here?

kola/tests/coretest/cloudinit.go
pbx0: Sounds reasonable but this file was a directly taken from github.com/coreos/coretest. The only thing I changed when porting coretest files was to change the function signatures of the tests to match that expected by the remote function runner. Then I added some glue to group the tests into 3 different kola tests that each call a series of native functions.

If coretest needs some reorganizing maybe we can do another PR?

mischief: if this is mostly reorganization, another pr would probably be fine. it seems there are a lot of instances of exec'ing commands or calling into `syscall` directly, which is suspicious.

marineam:
Hm, I'm not convinced importing coretest directly is really what we want... but if we do do that it would be slick to preserve the git history.

marineam:
Here is a branch that pulls in a filtered history of coretest: https://github.com/marineam/mantle/commits/coretest

it excludes extra bits like third_party or the README, and adjusts commit messages so github autolinking of pull requests point to the right location. So if we merge that branch and rebase ef9bc81 would then be dropped from this PR

pbx0:
Very nice. I'll rebase this PR once we merge that branch then.

marineam:
@pbx0 pushed that branch to master. Go ahead and rebase. And travis is going to be grumpy with me until your test porting commit lands. I would say in this first pass in this PR just get the basics plumbed together and the build working, we can clean up things in followup PRs.

pbx0:
PTAL I rebased it with the new history and dropped ef9bc81 in the process. Also commented out the fleet tests in coretest for the time being since they fail more often then not. Will followup and enable once I can get them working consistently.

marineam:
[LGTM](http://lgtm.in/i/UyXVZo0dq)

PR #53, opened 2015-05-28
add test and cover scripts

pbx0:


cover
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash -e
+#
+# Generate coverage HTML for a package
+# e.g. PKG=./unit ./cover
```
marineam: Ugg, I know this comes from our standard goo scripts but I find it completely irritating to the point that I never use it.

 - PKG should be provided as an argument, not environment.
 - The default behavior should be to scan all packages, if the go coverage tool is really incapable of reporting multiple packages at once then loop over the packages running go test once for each.

test
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash -e
+#
+# Run all tests (not including functional)
+#   ./test
+#   ./test -v
+#
+# Run tests for one package
+#   PKG=./foo ./test
```
marineam: Just no... PKG should be an argument and multiple packages should be accepted.

test
```diff
@@ -0,0 +1,56 @@
...
+#   PKG=bar ./test
+#
+
+# Invoke ./cover for HTML output
+COVER=${COVER:-"-cover"}
+
+source ./build
+
+TESTABLE="util network"
+FORMATTABLE="$TESTABLE auth kola cli cmd index platform version"
```
marineam: Why do these have to be listed?

marineam:
Go ahead and toss the debug.go fix in a new PR... I intentionally never added these standard scripts to mantle because I always thought they were crap and just used `go test` directly but if we could create not-crap versions of the scripts I would be game for that :)

pbx0:
Sounds good to me.  I'll revisit these scripts when we think we need something like it. Only the build script was requested.

PR #55, opened 2015-05-28
First pass at a NTP implementation

marineam:
Started hacking on this over the weekend, not integrated into the local platform code yet but once it is it should help with coretest and enable more comprehensive NTP client testing.

### DIFF
--- a/network/ntp/protocol.go
+++ b/network/ntp/protocol.go
@@ -1,0 +1,250 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// NTP v4 (RFC 5905)
+// http://tools.ietf.org/html/rfc5905
+package ntp
+
+import (
+	"encoding/binary"
+	"fmt"
+	"time"
+)
+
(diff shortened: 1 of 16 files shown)


network/ntp/protocol.go
```diff
@@ -0,0 +1,250 @@
...
+	"encoding/binary"
+	"fmt"
+	"time"
+)
+
+// JAN_1970 is the Unix Epoch in NTP Seconds (1970 - 1900)
+const JAN_1970 = 2208988800
+
+// short-hand to make the marshal functions less tedious
+var be binary.ByteOrder = binary.BigEndian
```
mischief: the type `binary.ByteOrder` can be elided here

network/ntp/protocol.go
```diff
@@ -0,0 +1,250 @@
...
+//  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+//  |          Seconds              |           Fraction            |
+//  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+//
+type Short struct {
+	Seconds  uint16
+	Fraction uint16
+}
+
+func (s *Short) marshalBinary(b []byte) {
```
mischief: why not use `binary.Read`/`binary.Write` here? or implement `encoding.BinaryMarshaller`/`encoding.BinaryUnmarshaller`? same for Timestamp below

marineam: Because I wanted a little helper for BinaryMarshaller that filled an existing slice instead of returning a new one that had to be copied. binary.Read/Write could be used if I switched BinaryMarshaller/BinaryUnmarshaller to use bytes.Buffer instead of a pre-allocated slice but these two types are so simple that it isn't a big deal to spell out manually.

I absolutely would use binary.Read/Write if it could handle the whole Header struct but without bitfield support in go it cannot so... meh.

network/ntp/protocol.go
marineam: Fixed

mischief:
lgtm

marineam:
Rebased and travis is happy now. :)

PR #56, opened 2015-06-01
Misc updates for mantle/kola

marineam:
Depends on https://github.com/coreos/pkg/pull/20


pbx0:
LGTM -- For my own understanding though, how would the rest of the kola code write to this logger?

marineam:
@pbx0 each package creates its own PackageLogger

marineam:
will start converting the rest shortly

PR #59, opened 2015-06-03
coretest: wait for the etcd cluster before beginning etcd tests

marineam:
The cluster health function is ripped off from the EtcdDiscovery tests
but modified to run locally instead of remotely. This whole thing needs
to be gutted to not use curl and/or moved out of coretest, there isn't
any reason for it to be local except for its history.

Until then at least it doesn't fail now :)


pbx0:
lgtm

PR #60, opened 2015-06-03
kola: check etcd1 cluster health reliably

marineam:
etcd1 clusters can report healthy with one or two hosts so continue
looping if not enough hosts report as healthy.


pbx0:
This is the right thing I've meant to make that change before. 

LGTM

PR #63, opened 2015-06-17
added gce-upload and gce-create from ore's upload cmd

pbx0:
Added `force` and `set-retries` flag to both commands. Without `--force` the commands will return success but skip any uploading for files that already exist.

The one piece left of these commands that may not be suitable for scripting is the authentication. Its still interactive when you have a token that doesn't work.


mischief:
lgtm

PR #64, opened 2015-06-21
Extend NTP server for testing leap seconds

marineam:


pbx0:
Not really familiar with how leap-seconds work, but LGTM. Yay go generate.

PR #66, opened 2015-07-01
add rolling upgrade test to kola

pbx0:
/cc @marineam @yichengq 

The test works by starting a 2.0.12 etcd cluster. Then a rolling upgrade is started by bringing down and replacing the binary of one node at a time. During this , random writes are being made and checked to try and ensure the cluster remains available the entire time. The specific versions are configurable, but right now it tests 2.0.12 -> 2.1.0-alpha.1 using precompiled binaries. As far as this test can tell, the reliable works and the final cluster version is updated.

@marineam I added a bit of a hack to get some flags into the rolling upgrade test itself.

### DIFF
--- a/kola/tests/etcd/rolling.go
+++ b/kola/tests/etcd/rolling.go
@@ -1,0 +1,157 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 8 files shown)


marineam:
I'd rather not check in binaries into git

pbx0:
Removed the checked in binaries and changed the default location to look for them to just be `etcd` and `etcd2` in the working dir. 

In making this work more generically across any etcd binaries that support rolling updates, I added a skipVersionCheck flag so you don't have to specify the  versions you expect out of etcd. Just the rolling update, health checks, and availability are checked.

marineam:
LGTM

kola/tests/etcd/rolling.go
```diff
@@ -0,0 +1,184 @@
...
+		for _, m := range cluster.Machines() {
+			if err := m.StartJournal(); err != nil {
+				return fmt.Errorf("failed to start journal: %v", err)
+			}
+		}
+	}
+
+	// drop in starting etcd binary
+	plog.Debug("adding files to cluster")
+	if err := cluster.DropFile(etcdBin); err != nil {
```
marineam: perhaps add a TODO to gracefully skip tests if the binaries aren't available once there is a meaningful way to actually do that.

pbx0: done

PR #72, opened 2015-08-01
dnsmasq: add more detail in error messages

crawford:
```
$ sudo kola qemu
Cluster failed: Network setup failed: LinkAdd() failed: operation not supported
```

Apparently, I didn't build my kernel with CONFIG_BRIDGE.


marineam:
LGTM

marineam:
Since this will run on random dev boxes should we add explicit checks for kernel features?

crawford:
That would be nice to have, but probably not necessary. Developers should be able to interpret the error messages.

PR #73, opened 2015-08-06
Start implementing an omaha server

marineam:
This just covers the protocol, my server works but needs some cleanup before it is practical to merge.

Note: this may get split back out into a new version of go-omaha but for now keeping it in-tree to simplify iterating on it and getting it into kola tests.


marineam:
@kelseyhightower this may be of interest to you.

@pbx0 this includes the generic download/verification code I mentioned earlier.

pbx0:
LGTM

PR #75, opened 2015-08-16
Misc updates for manatle's omaha

marineam:
- capnslog has seen a number of updates, and now always has logging enabled by default.
- add support for computing hashes for omaha packages
- a few functions for navigating the SDK

### DIFF
--- a/network/omaha/package.go
+++ b/network/omaha/package.go
@@ -1,0 +1,112 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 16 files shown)


network/omaha/package.go
```diff
@@ -0,0 +1,112 @@
...
+// is not a standard part of the Omaha protocol which only uses Sha1.
+type Package struct {
+	Name     string `xml:"name,attr"`
+	Sha1     string `xml:"hash,attr"`
+	Sha256   string `xml:"sha256,attr,omitempty"`
+	Size     uint64 `xml:"size,attr"`
+	Required bool   `xml:"required,attr"`
+}
+
+func (p *Package) FromFile(name string) error {
```
vcaputo: nit: name implies the name of a dentry, this is a path.

vcaputo:
minor nit, otherwise lgtm

network/omaha/package.go
marineam: Changed to `FromPath`

PR #77, opened 2015-08-25
platform: set format=raw for disk fds passed to qemu

mischief:
gets rid of an annoying message:

WARNING: Image format was not specified for '/dev/fdset/1' and probing guessed raw.
         Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
         Specify the 'raw' format explicitly to remove the restrictions.


marineam:
LGTM

PR #78, opened 2015-08-25
kola/tests: add test for starting a fleet unit through etcd2 proxy

mischief:


marineam:
LGTM, feel free to do the copyfile rework as a follow up or amend this PR.

PR #79, opened 2015-08-30
aws platform support

mischief:
need to figure out if we want to use goamz now, or just fix this pr to use aws-sdk-go. we will likely need the more complete features of aws-sdk-go later.


mischief:
switched to aws-sdk-go. this pr depends on #81.

mischief:
rebased after merging #81

pbx0:
LGTM

PR #81, opened 2015-08-31
fix godeps

mischief:
godep not behaving very well with code.google.com repos, so i manually disentangled it.

i tested godep restore and godep save -r ./... cycle, and it seems to work ok now, except for github.com/coreos/yaml:

```
$ godep restore
godep save -r ./...# cd /tmp/gopath2/src/github.com/coreos/yaml; git checkout master
error: pathspec 'master' did not match any file(s) known to git.
package github.com/coreos/yaml: exit status 1
godep: restore: exit status 1
```

but that's another matter.


pbx0:
Looks like github.com/coreos/etcd/pkg/metrics and pkg/flags were dropped in the 2nd commit according to the godep json file. Just checking that this was intentional, if so, LGTM.

mischief:
@pbx0 are those packages needed somewhere? mantle seemed to build fine with the updates in this pr.

pbx0:
I didn't check if they were used but it sounds like its fine. Just checking because godeps

marineam:
LGTM as far as I can tell ;-)

mischief:
well, i merged it, but github doesn't think i did.

PR #82, opened 2015-08-31
sdk: update build signing subkey

marineam:


pbx0:
Was this key just truncated somehow in the first commit? LGTM

marineam:
@pbx0 no, I added a new subkey. the old subkeys and master key remain unchanged.

pbx0:
That makes sense now. I was thinking one-key per pgp block.

Issue #70, opened 2015-07-15
qemu.image and other flags broken


mischief:
since c9c26119a43fceedaee831b5857c296799dc31f2 flags defined in https://github.com/coreos/mantle/blob/master/kola/flags.go no longer work because `flag.Parse` is no longer called.

mischief:
fixed by #83

PR #83, opened 2015-08-31
make options more consistent

mischief:
currently ore is broken by the removal of the old GCEOpts.


mischief:
fixes #70

pbx0:
Barring reorganizing the location of the options structs and the 2 line nits I made, LGTM. Nice refactor

mischief:
@pbx0 @marineam PTAL. sorry if i broke anything, but it all builds now.

cmd/kola/kola.go
```diff
@@ -47,18 +46,19 @@ var (
 )
 
 func init() {
-	cmdRun.Flags().StringVar(&kolaPlatform,
-		"platform", "qemu", "VM platform: qemu or gce")
-	cmdRun.Flags().StringVar(&EtcdRollingVersion,
+	sv := cmdRun.Flags().StringVar
+
+	sv(&EtcdRollingVersion,
```
marineam: Unless you are going to further reformat this and make all the sv calls a single line I don't think there is much value in using sv instead of the full name.

marineam:
LGTM

cmd/kola/kola.go
pbx0: I have plans to rip these options out today once this PR goes in anyway since its all a hack for that one test.

mischief: fixed the formatting.

mischief:
top level options are broken. please don't merge.

mischief:
global flags fixed, sv calls turned to single lines.

PR #86, opened 2015-09-01
platform/local: fix etcd server configuration

mischief:
etcd needs the member to be in InitialPeerURLsMap.


pbx0:
LGTM -- A link to the etcd documentation about this being necessary might be informative if you have have one.

PR #87, opened 2015-09-01
kola/tests/coretest: add useradd test

mischief:
this wasn't originally in coretest, but it seemed ok for this test to live here until we reorganize them.

should this test userdel too?


marineam:
LGTM

PR #89, opened 2015-09-02
Options

pbx0:
We needed to move the rolling upgrade to a subcommand because of its reliance on pre-compiled binaries making it ill-suited for automatic OS testing. 

Additionally I wanted a better way for these types of one-off tests that need special options, to be implemented as kola subcommands without polluting the kola package itself with all these options. I choose a way to pass test options that works for now, but i'm open to improving it as well.

These subcommand tests can just register themselves to avoid them being when running the automated tests. The test harness exports RunTest now so that subcommands don't need to write any kola cluster creation code.


kola/harness.go
```diff
@@ -151,8 +164,20 @@ func runTest(t *Test, pltfrm string) error {
 	for k := range t.NativeFuncs {
 		names = append(names, k)
 	}
+
+	// prevent unsafe access if tests ever become parallel and access
+	tempTestOptions := make(map[string]string, 0)
```
mischief: i'm not sure that this will ever be a problem in practice, because the options are written when the program starts, and (hopefully!) never written again. concurrent readers should be OK without locking.

mischief:
we will probably have to rework option passing in the future, but it works for now. lgtm

kola/harness.go
pbx0: Yeah, I figured it was unlikely, but just being careful. It would be odd for a test to do any writes, but, the api doesn't prevent it either I guess.

pbx0:
I agree, I'm leaving the option passing as it for now for lack of clearly better methods. The current way keeps more of the code out of the main kola package and its deficits (only string types) are made obvious.

pbx0:
retest this please

pbx0:
retest this please

pbx0:
retest this please

pbx0:
retest this please

PR #91, opened 2015-09-06
kola/tests/coretest: run id instead of calling user.Lookup to verify useradd

mischief:
the golang package for fedora does not link to libc, so when kolet is
compiled on fedora, user.Lookup tests will fail at runtime.


marineam:
LGTM

PR #92, opened 2015-09-06
adjust etcd health checks

mischief:
etcdctl output format has changed recently to include two instances of 'healthy' per member.

etcd also takes a bit more time to come up on aws, so raise the retry count and sleep interval.


marineam:
LGTM

PR #94, opened 2015-09-11
extract retrying function into util

mischief:


kola/tests/fleet/fleet.go
```diff
@@ -136,19 +137,21 @@ func Proxy(c platform.TestCluster) error {
 	var status []byte
 
 	// 5 seconds seems like a reasonable amount of time to make this happen.
```
pbx0: I think this comment is no longer relevant here. Deleting or moving down to the util.Retry call is probably fine.

mischief:
@pbx0 PTAL

pbx0:
I was wondering about the scope of the util package since all files affected are under kola, but, this could also be used under platform/gce.go and qemu.go for ssh retries so i think mantle/util is fine. Up to you if you want to add this to platform or not right now. LGTM

PR #95, opened 2015-09-11
kola/tests/misc: use systemd to check nfs mount; increase timeout

mischief:


mischief:
retest this please

pbx0:
LGTM

PR #96, opened 2015-09-12
platform: convert ssh checks to use util.Retry

mischief:


pbx0:
LGTM after build error fixed

mischief:
must have missed that when trying to split out all these patches.

mischief:
@pbx0 PTAL

mischief:
i merged this but i forgot to rebase. silly github :(

PR #97, opened 2015-09-12
platform: add Manhole

mischief:
this is useful when debugging the machines spawned by tests.


mischief:
i could be convinced to name this something more geologic, but i think the name fits.

pbx0:
I like the name as well. I might just add another doc sentence stating something like. "Manhole is intended for use within a failing kola test."

LGTM

PR #99, opened 2015-09-12
cmd/kola: add list command

mischief:


vcaputo:
lgtm except for the checks failing...

mischief:
@vcaputo the only check failure is jenkins running the kola tests. this is expected.

PR #104, opened 2015-09-16
platform: use a lock to prevent races around qemu cluster

mischief:


mischief:
cc @pbx0 @vcaputo

vcaputo:
won't be surprised if there's more to serialize but seems reasonable as a start, lgtm.

pbx0:
LGTM.
(although I'd still advocate parallelizing small chucks per platform of over coming back and adding murexes per platform)

PR #108, opened 2015-09-17
kola: add NFS version 4 test

mischief:


kola/tests/misc/nfs.go
```diff
@@ -61,7 +55,23 @@ func NFS(c platform.TestCluster) error {
...
+After=network-online.target
+Requires=network-online.target
+After=rpc-statd.service
+Requires=rpc-statd.service
+
+[Mount]
+What=%s:/tmp
+Where=/mnt
+Type=%s
+Options=defaults,noexec
```
marineam: You need to use the `nfs` type and toggle `nfsvers=4` vs `nfsvers=3` in options. The client and server are free to negotiate version 3 or 4 if with the `nfs` type.

marineam: So in short, you may have been testing v4 all along ;-)

mischief: i'm not so sure - NFSv4 is failing against current alpha, but not against my local copy with updated nfs-utils. :confounded:

pbx0:
LGTM

kola/tests/misc/nfs.go
marineam: may have been negotiating v3 then since v4 was broken. *shrug*

marineam: whatever the case, nfsvers= is required to be sure of the version in use.

mischief:
@marineam ptal

marineam:
LGTM

PR #109, opened 2015-09-17
Downloadimage

pbx0:
Adds a download image command to cork. This is intended to replace curling the image down each time kola runs. Now the image is verified and only updated if its different form the local copy.

### DIFF
--- a/cmd/cork/downloadimage.go
+++ b/cmd/cork/downloadimage.go
@@ -1,0 +1,116 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/coreos/mantle/Godeps/_workspace/src/github.com/spf13/cobra"
+	"github.com/coreos/mantle/sdk"
+)
+
+var (
+	downloadImageCmd = &cobra.Command{
+		Use:   "download-image",
+		Short: "Download and verify CoreOS images",
+		Long:  "Download and verify current CoreOS images to a local cache.",
+		Run:   runDownloadImage,
+	}
+	downloadImageRoot         string
+	downloadImageCacheDir     string
+	downloadImagePrefix       string
+	downloadImageVerify       bool
+	downloadImagePlatformList platformList
+)
+
+func init() {
+	downloadImageCmd.Flags().StringVar(&downloadImageRoot,
+		"root", "http://storage.core-os.net/coreos/amd64-usr/master/", "base URL of images")
+	downloadImageCmd.Flags().StringVar(&downloadImageCacheDir,
+		"cache-dir", "", "local dir for image cache")
+	downloadImageCmd.Flags().StringVar(&downloadImagePrefix,
+		"image-prefix", "coreos_production", "image filename prefix")
+	downloadImageCmd.Flags().BoolVar(&downloadImageVerify,
+		"verify", true, "verify")
+	downloadImageCmd.Flags().Var(&downloadImagePlatformList,
+		"platform", "Choose qemu, gce, or aws. Multiple platforms can be specified by repeating the flag")
+
+	root.AddCommand(downloadImageCmd)
+}
+
+type platformList []string // satisfies pflag.Value interface
+
+func (platforms *platformList) String() string {
+	return fmt.Sprintf("%v", *platforms)
+}
+
+// not sure what this is for, but won't compile without it
+func (platforms *platformList) Type() string {
+	return "platformList"
+}
+
+// Set will append additional platform for each flag set. Comma
+// separated flags without spaces will also be parsed correctly.
+func (platforms *platformList) Set(value string) error {
+
+	// maps names of platforms to filename prefix
+	platformMap := map[string]string{
+		"qemu": "_image.bin.bz2",
+		"gce":  "_gce.tar.gz",
+		"aws":  "_ami_image.bin.bz2",
+	}
+
+	values := strings.Split(value, ",")
+
+	for _, platform := range values {
+		prefix, ok := platformMap[platform]
+		if !ok {
+			plog.Fatalf("platform not supported: %v", platform)
+		}
+		*platforms = append(*platforms, prefix)
+	}
+	return nil
+}
+
+func runDownloadImage(cmd *cobra.Command, args []string) {
+	if len(args) != 0 {
+		plog.Fatalf("Unrecognized arguements: %v", args)
+	}
+
+	if downloadImageCacheDir == "" {
+		plog.Fatal("Missing --cache-dir=FILEPATH")
+	}
+	if len(downloadImagePlatformList) == 0 {
+		plog.Fatal("Must specify 1 or more platforms to download")
+	}
+	if downloadImageVerify == false {
+		plog.Notice("Warning: image verification turned off")
+	}
+
+	// download signed digest
+	for _, suffix := range downloadImagePlatformList {
+		file := downloadImagePrefix + suffix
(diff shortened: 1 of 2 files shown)


sdk/download.go
```diff
@@ -151,3 +153,40 @@ func DownloadSDK(version string) error {
...
+	// get local sig
+	localSig := []byte{}
+	if _, err := os.Stat(sigFile); err == nil {
+		localSig, err = ioutil.ReadFile(sigFile)
+		if err != nil {
+			return err
+		}
+	}
+
+	// overwrite local sig with latest
```
marineam: I'd be inclined to add a `UpdateFile` which downloads to memory or a temp file and only replaces the original if it changed. Not a big deal but it feels cleaner to not be needlessly changing the inode/mtime.

cmd/cork/downloadimage.go
```diff
@@ -0,0 +1,128 @@
...
+	downloadImagePrefix       string
+	downloadImageVerify       bool
+	downloadImagePlatformList platformList
+)
+
+func init() {
+	downloadImageCmd.Flags().StringVar(&downloadImageRoot,
+		"root", "http://storage.core-os.net/coreos/amd64-usr/master/", "base URL of images")
+	downloadImageCmd.Flags().StringVar(&downloadImageCacheDir,
+		"cache-dir", "", "local dir for image cache")
```
marineam: For the sake of consistency how about make this default to a directory under `sdk.RepoCache()`

sdk/download.go
pbx0: Sounds good. I'll do it to a temp file so as not to copy all the DownloadFile() code. UpdateFile will need some sort of compareFileBytes() function. Either by chunking and comparing bytes or hashing and comparing that.

pbx0:
@marineam PTAL -- CacheDir defaults to sdk.RepoCache() and added UpdateFile() function.

cmd/cork/downloadimage.go
```diff
@@ -0,0 +1,128 @@
...
+	downloadImagePrefix       string
+	downloadImageVerify       bool
+	downloadImagePlatformList platformList
+)
+
+func init() {
+	downloadImageCmd.Flags().StringVar(&downloadImageRoot,
+		"root", "http://storage.core-os.net/coreos/amd64-usr/master/", "base URL of images")
+	downloadImageCmd.Flags().StringVar(&downloadImageCacheDir,
+		"cache-dir", sdk.RepoCache(), "local dir for image cache")
```
marineam: lets use a sub dir like 'images' to make cleanup simple.

marineam: for reference the cache of sdk tarballs is "sdks", sources for portage is "distfiles" and so forth.

sdk/download.go
```diff
@@ -151,3 +154,93 @@ func DownloadSDK(version string) error {
...
+	if err != nil {
+		return false, err
+	}
+	return bytes.Equal(sum1, sum2), nil
+}
+
+// UpdateFile downloads a file to temp dir and replaces the file only if
+// contents have changed. If tempDir is "" default will be os.TempDir().
+func UpdateFile(file, url, tempDir string) error {
+	t, err := ioutil.TempFile(tempDir, "sdkUpdateCheck")
```
marineam: Lots of error conditions where the temp file will not be removed.

pbx0:
@marineam PTAL -- Should be all in order now. I wasn't thinking about tmp cleanup before. Also it no longer uses os.Rename to copy over the temp file since it will fail when tmp is a different partition.

sdk/download.go
```diff
@@ -151,3 +154,106 @@ func DownloadSDK(version string) error {
...
+	sum2, err := fileSum(fileName2)
+	if err != nil {
+		return false, err
+	}
+	return bytes.Equal(sum1, sum2), nil
+}
+
+// UpdateFile downloads a file to temp dir and replaces the file only if
+// contents have changed. If tempDir is "" default will be os.TempDir().
+func UpdateFile(file, url, tempDir string) error {
```
vcaputo: Why not just derive the temp file path from the file-to-update path so you're able to reliably use Rename()?   e.g.
```t, err := ioutil.TempFile(filepath.Dir(file), "sdkUpdateCheck")```

File: sdk/download.go [hunk 6]
```diff
@@ -151,3 +154,106 @@ func DownloadSDK(version string) error {
...
+	t, err := ioutil.TempFile(tempDir, "sdkUpdateCheck")
+	if err != nil {
+		return nil
+	}
+	t.Close()
+	tempFile := t.Name()
+	defer os.Remove(tempFile)
+
+	if err := DownloadFile(tempFile, url); err != nil {
```
vcaputo: It feels like a new Download function should be added which receives a simple WriterCloser interface instead of a string path for the file.  Then you could insert the sum/compare in-line with the download.  DownloadFile would probably become a caller of this new more generalized function.

sdk/download.go
pbx0: Yes definitely, thats much simpler.

sdk/download.go
vcaputo: Somewhat related:
https://github.com/vcaputo/coreos-install/blob/basic_installation/src/main.go#L206

We're doing quite similar things here, should probably share a package.

pbx0:
tempDir is now in the cacheDir so we can safely use os.Rename(). Added a note about possibly adding a Download () func some other time.

PTAL

sdk/download.go
```diff
@@ -151,3 +154,96 @@ func DownloadSDK(version string) error {
...
+	hash := sha512.New()
+	if _, err := io.Copy(hash, f); err != nil {
+		return nil, err
+	}
+
+	return hash.Sum(nil), nil
+}
+
+func compareFileBytes(fileName1, fileName2 string) (bool, error) {
+	sum1, err := fileSum(fileName1)
```
vcaputo: why are we checksumming both files here?

the checksum must read in each file entirely before it can be compared

why not compare the contents of the two and abort on the first difference?

Also, why not start this whole thing out with two stat calls which compare the sizes to see if they differ before doing any further IO?

the comparison loop should just be reading a block from file1, block from file2, compare the block, return false on first different block.

fall-through returns true

checksumming is just adding computational overhead, serializing the process, and requiring both files be read entirely always.

pbx0: Mostly because it was easier to write and easier to get correct. The better way has kinda tricky error conditions. 

Also, when using the verified updater we are only comparing .sig files so the optimization doesn't mean much for the common case. When using UpdateFile() we have to download the image first anyway which dwarfs the file comparison step.

That said after thinking about it more,  I couldn't help writing the optimized version.

pbx0:
@vc PTAL -- optimized the file compare function.

sdk/download.go
```diff
@@ -151,3 +153,128 @@ func DownloadSDK(version string) error {
...
+
+		if err1 == io.EOF && err2 == io.EOF {
+			return true, nil
+		} else if err1 == io.EOF || err2 == io.EOF {
+			return false, nil
+		}
+
+		if err1 == io.ErrUnexpectedEOF && err2 == io.ErrUnexpectedEOF {
+			if n1 == n2 {
+				return bytes.Equal(buf1, buf2), nil
```
vcaputo: I don't think this is correct due to: (from ReadFull() documentation)
> On return, n == len(buf) if and only if err == nil.

So you're comparing the entirety of buf1 and buf2 when the comparison should really be clipped to n1:
> return bytes.Equal(buf1[:n1], buf2[:n1])

So I think this could actually be simplified to just replacing ```if n1 == n2 { ... } else { ... }``` altogether with ```return bytes.Equal(buf1[:n1], buf2[:n2])```, since it will be false if the slice lengths differ.

pbx0: In the case in which n1 == n2 either it returned early on the first chunk and so the rest of the buffer is zero (and equal) or it successfully compared the last chunk and so, also, the rest of the buffer is equal. If I didn't check n1== n2 and checked the full buffer then this would create a weird and rare failure case.

Your way is probably more readable though so I'll change that.

sdk/download.go
```diff
@@ -151,3 +153,124 @@ func DownloadSDK(version string) error {
...
+		}
+
+		if err1 != nil {
+			return false, err1
+		}
+		if err2 != nil {
+			return false, err2
+		}
+
+		if !bytes.Equal(buf1, buf2) {
```
vcaputo: ```return bytes.Equal(buf1, buf2), nil```

vcaputo: nevermind, I'm crazy.

vcaputo:
lgtm

PR #111, opened 2015-09-18
kola/tests/misc: use private ip in nfs tests

mischief:
this will hopefully make the nfs tests pass on aws, where the security group disallows nfs connections to the public ip.


pbx0:
lgtm

PR #115, opened 2015-09-18
kola/tests/coretest: wrap coretest fleet tests in retries

pbx0:
This helps fix flakiness issues with this tests on certain platforms.

### DIFF
--- a/kola/tests/coretest/fleet.go
+++ b/kola/tests/coretest/fleet.go
@@ -8,6 +8,8 @@ import (
 	"strings"
 	"syscall"
 	"time"
+
+	"github.com/coreos/mantle/util"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pbx0:
@mischief

marineam:
Could test flakyness just be due to fleet's over-agressive 1-second timeout for every etcd request? Would just upping that to a reasonable value work?

pbx0:
@marineam The local default is 10 but its getting the 1 second from fleet. Should we just hard set it to ~10 seconds and delete the retries if that works?

marineam:
I'd say retrying to wait for the system to settle is fair, the rest of the time intermittent failures would actually cause problems for people. So upping the timeout in fleetd so it is more tolerant of etcd running on slow disks and such would be preferable.

pbx0:
Unsure what to do about this PR for the time being. Increasing the timeout only (https://github.com/coreos/mantle/pull/118) doesn't seem to help for me. Yet, @michief is unable to reproduce my errors on the same machine. PR 118 is fine to pass through but doesn't seem to solve anyone's problem. While this PR works for me, it also may be doing too much and covering up underlying bugs or future bugs.

pbx0:
retest this please

pbx0:
retest this please

pbx0:
retest this please

pbx0:
Similar errors also happen using the coretest scripts on amazon. So this seems to be an issue with fleet commands sometimes failing and not only occurring on kola. What do people think about filing a bug against fleet and a bug against mantle/kola. The fleet bug is about the error we commonly see touching unit-files. Mantle bug is to remove some of these retries and depends on the fleet bug finding a solution. 

This way we don't have to disable the test for fleet in the mean-time. In my tests, I can cause errors by taking away any of the retries added here so they seem necessary to make it pass for now. The fleet-proxy test will probably need a similar retry mechanism as I have seen that fail too.

@marineam @mischief

pbx0:
Replaced by https://github.com/coreos/mantle/pull/129,  closing.

PR #118, opened 2015-09-21
kola: increase fleet etcd request timeout

mischief:
this should help the tests pass on a system with slow io.


mischief:
this does not in fact fix the bug. let's go along the lines of #115 instead.

PR #121, opened 2015-09-23
plume aws

mischief:


mischief:
`plume aws register-image $(plume aws create-snapshot $(plume aws import-volume $(plume aws s3-manifest $(plume aws s3-upload /home/mischief/src/coreos/src/build/images/amd64-usr/latest/coreos_production_ami_vmdk_image.vmdk)))`

is what this can do right now.

pbx0:
Do we need to expose all these aws details to the command line? What about just making `plume aws upload [...]coreos_production_ami_vmdk_image.vmdk do the equivalent of the above command. Or maybe call it create-image or upload-image or something like that.

mischief:
@pbx0 there is a `one-shot` subcommand.

i broke out the subcommands for testing. having to go through the whole process to test one part was a hassle.

platform/aws_test.go
```diff
@@ -1,57 +0,0 @@
-// Copyright 2015 CoreOS, Inc.
```
pbx0: In moving aws to its own amazon package, was deleting this intentional? It looks like it wasn't copied over.

mischief: the test was broken, and i'm not sure it's worth fixing.

marineam:
There are basically two things that may be worth having two commands:

 - take a disk image and create a EBS snapshot, the user won't really care about the intermediate steps like s3 manifest, ebs volume, etc. If in the future for example we find out how to import directly to a snapshot, skipping the volume, we can do so without changing anything the user does.
 - take a snapshot and register an image. There is one snapshot but two different AMIs normally (PV vs HVM) so it is reasonable that a user may want to do only one or do the two differently. Or create more variations because yet another feature has been added that requires independent AMIs because amazon.
 - And then there should be a command that does the default of uploading and creating the two default AMIs

platform/aws_test.go
pbx0: sounds good

mischief:
@marineam point 1 is the subcommands `s3-upload`, `s3-manifest`, `import-volume`, `create-snapshot`. point 2 is `register-image`, which can do both pv/hvm easily. point 3 is implemented right now as `one-shot`, which does all 5 steps. right now, i do not have a publish command though.

PR #122, opened 2015-09-23
make gce's NewMachine() safe to run in parallel

pbx0:


platform/gce.go
```diff
@@ -44,10 +46,11 @@ type GCEOptions struct {
...
 type gceCluster struct {
-	api      *compute.Service
-	sshAgent *network.SSHAgent
-	conf     *GCEOptions
-	machines map[string]*gceMachine
+	api        *compute.Service
+	sshAgent   *network.SSHAgent
+	conf       *GCEOptions
+	machines   map[string]*gceMachine
+	sync.Mutex // protects only against NewMachine being called in parallel
```
vcaputo: Generally it's better to associate locks with the data they protect, not code.
At a glance, it looks like the mutex serializes access to the machines map.

I presume you left Destroy() alone because it needs to own the gceCluster instance to destroy it safely, but it's kind of an exception I'd avoid and throw the lock in anyway for consistency sake.

Machines() isn't being serialized, what's the rationale behind that?

pbx0: Yeah, I left Destroy() alone for the same reason as Machines(). I only wanted the minimum allow NewMachine() to be called in parallel since that is the only function we currently expect to work in parallel. I'm not sure why would expect any other functions on a cluster to work in parallel since the owner of that cluster can manage that. Listing out Machines() doesn't have an expectation that they will change out from under you.

As far as associating locks with the data, are you suggesting something like this:
`type machines struct {
machs map[string]*gceMachine
sync.Mutex
}
`

vcaputo: I don't know what's idiomatic for golang, but it's generally accepted to "lock data, not code", so in this case the lock protects the machines map, I'd expect the code to consistently wrap access to the machines map in a lock/unlock sequence.  The comment says the lock protects code, not data, which isn't a good sign.  The comment also doesn't clearly express which members of the structure are protected by the lock.  So at a least, the comment should be updated.

Imagine a future reader investigating a potential race condition/crash, and they suspect there's a data structure not protected with a lock where needed.  Since the code has inconsistently serialized access to this map, the exceptions where there's no locking look like obvious bugs at a glance, until further scrutiny may show that those code paths aren't entered concurrently like the ones with the lock.  What's the harm in holding the lock everywhere this shared data structure is accessed?  It's not performance-sensitive code, so just protect the shared data everywhere it's accessed.

pbx0: Hmm, thats fair I will lock all access to the map while we are here. 

I wanted to convey that NewMachine() was the only part of the interface expected to work in parallel since we essentially are locking code by going back and making this so in the first place. Its generally not expected for the cluster to be something that thread safe overall. I guess user's of a cluster can still assume that the cluster is not a thread safe interface.

mischief:
@pbx0 anything that could read or write shared data needs to be locked. practically this means that we have a mach map in the cluster struct, it is modified by Cluster.NewMachine, read by Cluster.Machines, and (unfortunately) modified by Machine.Destroy. so, those functions need to protect the mach map.

pbx0:
cluster.Machines() has locks and changed comment, PTAL

(Not sure why the sdk bug fix commit is showing up in this PR, its already in master. Github bug I guess)

mischief:
@pbx0 i believe if you rebase on master it will go away?

pbx0:
Oh, yes indeed, that fixes it.

platform/gce.go
```diff
@@ -44,10 +46,11 @@ type GCEOptions struct {
...
 type gceCluster struct {
-	api      *compute.Service
-	sshAgent *network.SSHAgent
-	conf     *GCEOptions
-	machines map[string]*gceMachine
+	api        *compute.Service
+	sshAgent   *network.SSHAgent
+	conf       *GCEOptions
+	machines   map[string]*gceMachine
+	sync.Mutex // protects concurrent access to machines
```
mischief: can you name this 'mu' instead of embedding it? i'd like to move each platform's code to its own package, and when that happens, gceCluster will need to be exported as GCECluster. let's prevent accidents by not exposing Lock/Unlock there.

mischief:
fix the last nit of embedded sync.Mutex and lgtm

platform/gce.go
vcaputo: +1

I noticed this but ignored it since gceCluster was private

vcaputo:
lgtm

PR #124, opened 2015-09-24
allow non-interactive auth when kola runs from a GCE vm

pbx0:


auth/google.go
```diff
@@ -123,3 +124,17 @@ func GoogleTokenSource() (oauth2.TokenSource, error) {
 	}
 	return conf.TokenSource(oauth2.NoContext, tok), nil
 }
+
+// Fetch from Google Compute Engine's metadata server to retrieve
+// an access token for the provided account.
+func GoogleServiceClient() *http.Client {
+	return &http.Client{
+		Transport: &oauth2.Transport{
+			Source: google.ComputeTokenSource(""),
```
marineam: What does the empty string mean?

pbx0: Empty string means it will use the "default" service account which seems to work fine. Should I expose that to the API?
https://godoc.org/golang.org/x/oauth2/google#ComputeTokenSource

mischief:
lgtm

marineam:
LGTM, just using the default service accound should be fine

PR #127, opened 2015-09-28
various changes to ore upload in support of kola gce script

pbx0:


cmd/ore/upload.go
```diff
@@ -66,8 +70,9 @@ func runUpload(cmd *cobra.Command, args []string) {
 
 	// if an image name is unspecified try to use version.txt
 	if uploadImageName == "" {
-		uploadImageName = getImageVersion(uploadFile)
-		if uploadImageName == "" {
+		var err error
+		uploadImageName, err = sdk.GetVersion(filepath.Dir(uploadFile))
+		if err != nil {
 			fmt.Fprintf(os.Stderr, "Unable to get version from image directory, provide a -name flag or include a version.txt in the image directory\n")
```
vcaputo: while you're here may as well get the err value into the printed message

vcaputo:
lgtm

cmd/ore/upload.go
pbx0: done

PR #128, opened 2015-09-29
platform/gce: wait for image creation to finish on request

pbx0:
When scripting image uploads before a kola run, ongoing image creation
operations would cause kola to fail. This change makes image uploads
wait for the image create operation to complete before returning.

### DIFF
--- a/platform/gce.go
+++ b/platform/gce.go
@@ -290,7 +290,8 @@ func GCEDestroyVM(api *compute.Service, proj, zone, name string) error {
 	return nil
 }
 
-// Create image on GCE and return. Will not overwrite existing image.
+// Create image on GCE and and wait for completion. Will not overwrite
+// existing image.
 func GCECreateImage(api *compute.Service, proj, name, source string) error {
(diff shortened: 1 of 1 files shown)


mischief:
lgtm

platform/gce.go
```diff
@@ -430,6 +440,35 @@ OpLoop:
...
+		if err != nil {
+			return fmt.Errorf("Failed to get op %s: %v", opname, err)
+		}
+		switch op.Status {
+		case "PENDING", "RUNNING":
+			continue
+		case "DONE":
+			if op.Error != nil {
+				for _, operr := range op.Error.Errors {
+					return fmt.Errorf("Error creating instance: %+v", operr)
```
vcaputo: drive-by observation: we don't typically capitalize returned error strings

pbx0: gah, just merged it. Noted though.

mischief: :+1: but we can worry about it later.

PR #134, opened 2015-10-02
kola/tests/coretest: bump fleetctl retries to *40*

mischief:
20 seems too low for our test host, and frequently bails immediately
after the service was started by fleetd but before the status is
reported back to fleetctl.


pbx0:
lgtm

pbx0:
Closing, I believe we worked around this amd found that this bump didn't stop the failures.

PR #139, opened 2015-10-13
etcd/discovery: do not block while starting etcd

marineam:
Setting etcd2 to Type=notify means etcd2.service will not finish
starting until the cluster is operational. The cluster will not be
operational until enough etcd nodes have been started. Therefore
blocking until the first node reports as started is a deadlock.

Broke with https://github.com/coreos/coreos-overlay/pull/1576 on Thursday but no one seems to have noticed until I did a release today :( :( :(


pbx0:
lgtm

PR #140, opened 2015-10-13
kola/tests/fleet: need to retry first command to wait for etcd to settle

pbx0:
Increasing EtcdRequestTimeout to 60+ seconds still causes the start
command to reliably fail. It seems to have no effect. Retrying seems to
work reliably. This is needed to not have this break a jenkins build
every other run.

### DIFF
--- a/kola/tests/fleet/fleet.go
+++ b/kola/tests/fleet/fleet.go
@@ -124,22 +124,26 @@ func Proxy(c platform.TestCluster) error {
 	}
 	defer proxy.Destroy()
 
-	// settling...
-	time.Sleep(5 * time.Second)
-
 	err = copyFile(proxy, strings.NewReader(fleetunit), "/home/core/hello.service")
 	if err != nil {
 		return fmt.Errorf("copyFile: %s", err)
 	}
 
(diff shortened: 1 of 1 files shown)


kola/tests/fleet/fleet.go
```diff
@@ -124,17 +124,21 @@ func Proxy(c platform.TestCluster) error {
...
 	err = copyFile(proxy, strings.NewReader(fleetunit), "/home/core/hello.service")
 	if err != nil {
 		return fmt.Errorf("copyFile: %s", err)
 	}
 
-	_, err = proxy.SSH("fleetctl start /home/core/hello.service")
-	if err != nil {
-		return fmt.Errorf("fleetctl start: %s", err)
+	// settling...
+	retryFunc := func() error {
```
vcaputo: retryFunc is a terrible name... it's already obviously the retry function, it's passed to util.Retry().  But the opportunity to document *what* is being retried in the name has been missed.

pbx0: Sure, something more like fleetStart?

vcaputo:
lgtm

PR #143, opened 2015-10-13
cmd/cork: add retry flag for downloading images

pbx0:
This should help all the connection resets happening on ale.


pbx0:
Not really needed for the time being now that the actual source of the resets has been solved.

marineam:
This would still be useful functionality to have though. Perhaps it would be best to push down to the HTTP download code though so everything can benefit. Since we already support resuming downloads we could make use of that to automatically resume for errors that are likely transient.

PR #150, opened 2015-10-19
sdk: get rid of float conversions in GetCoreOSAge

mischief:


sdk/version.go
```diff
@@ -68,5 +68,5 @@ const (
 
 // GetCoreOSAge returns the number of days since the CoreOS epoch.
 func GetCoreOSAge() int64 {
-	return int64(time.Since(time.Unix(CoreOSEpoch, 0)).Seconds()) / 86400
+	return int64(time.Since(time.Unix(CoreOSEpoch, 0)) / (86400 * time.Second))
```
marineam: Should be able to drop the int64 cast right?

mischief: only if we want to return time.Duration.

marineam: Ah, right. In that case the intent would be clearer if the whole expression was wrapped in the cast, not just time.Since.

marineam: Er, wait. nm

I got my parrens mismatched, you are doing that. this is why I don't write LISP

marineam:
LGTM

PR #151, opened 2015-10-19
kola/tests: add kubernetes smoke test

pbx0:
Starts a 3 node kubernetes cluster using offical kubernetes guides at: http://kubernetes.io/v1.0/docs/getting-started-guides/coreos/coreos_multinode_cluster.html

There are multiple ways of going about setting up a cluster, this i just the first one I choose that worked reasonably well for me. In the future I'd also like a test that sets up a test strictly following our docs and launches the same smoke tests.

### DIFF
--- a/kola/kuberegistry.go
+++ b/kola/kuberegistry.go
@@ -1,0 +1,26 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kola
+
+import "github.com/coreos/mantle/kola/tests/kubernetes"
+
+func init() {
+	Register(&Test{
+		Name:        "kubeMultiNodeSmoke",
+		Run:         kubernetes.MultiNodeSmoke,
+		ClusterSize: 0,
+		Platforms:   []string{"gce", "aws"},
+	})
+}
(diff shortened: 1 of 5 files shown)


kayrus:
Hm... These 
@pbx0 there are [offical kubernetes guides](http://kubernetes.io/v1.0/docs/getting-started-guides/coreos/coreos_multinode_cluster.html) and [offical coreos-kubernetes guides](https://github.com/coreos/coreos-kubernetes). Official kubernetes guides need to be maintained, because they are quite old and doesn't meet official CoreOS way. So if you would like to run kubernetes tests you have to use our doc. I've already made some draft which builds k8s cluster using our official CoreOS way, you can try to port it:
https://github.com/endocode/coreos-docs/tree/kubernetes/kubernetes/examples

* [k8s_master.yaml](https://github.com/endocode/coreos-docs/tree/kubernetes/kubernetes/examples/k8s_master.yaml)
* [k8s_node.yaml](https://github.com/endocode/coreos-docs/tree/kubernetes/kubernetes/examples/k8s_node.yaml)

pbx0:
@kayrus Sure, I'll definitely add the coreos guides as a setup method. Thanks for the examples with cloud configs, that is very useful. For now though, would it not be useful to include both methods but, in separate tests? What are your thoughts?

Right now this test is purely for the OS release CI, we want to prioritize testing regressions in OS image , and not have our CI fail because SkyDNS, for example, contains the regression. So, i'm not sure where to draw the line here. SkyDNS seems useful to test though because it will exercise more OS software like etcd. Anyone one else have opinions here?

In the future, I hope the expand the scope and usefulness of kola so it is useful for testing software in external projects and have a way of delineating tests for the OS release process and tests for external projects.

kayrus:
> our CI fail because SkyDNS, for example, contains the regression

If we use nginx tests, why don't use dns tests? Our official k8s doc supposes that user will install skydns, but not nginx. So we can replace nginx with skydns. As for regression, we can tag specific version.

pbx0:
That sounds good to me. I'll make another PR using those guides and SkyDNS. We can come back to this and decide to close it or not.

mischief:
lgtm for now, we can iterate on this.

PR #153, opened 2015-10-19
sdk: retry on download failures

pbx0:
See: https://github.com/coreos/mantle/pull/143

@marineam

### DIFF
--- a/sdk/download.go
+++ b/sdk/download.go
@@ -24,8 +24,10 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"time"
 
 	"github.com/coreos/mantle/Godeps/_workspace/src/github.com/coreos/pkg/capnslog"
+	"github.com/coreos/mantle/util"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


marineam:
I think everything in `DownloadFile` short of the mkdir should be a private function that gets wrapped with Retry. As-is retries will make the same full or partial request instead of making use of anything that did get downloaded on the previous attempt. Letting the retries jump forward would actually be sufficient to make things succeed even in the face of our troublesome network card :)

marineam:
Wrapping most of the code also avoids the extra nesting this version has.

pbx0:
@marineam PTAL. I expanded the scope of the retry so it resumes whatever was copied in the first try.

I didn't split it into a private function simply because I needed a nested closure to make util.Retry useful anyway (`func() error` signature). I could just as easily write my own retry loop and split into a private function as well. Any preference?

marineam:
@pb I'd rather avoid the extra nesting but you can wrap the call to the private function in a one line closure to get both things done.

pbx0:
oh right, just wrap the private func:  @marineam  PTAL, no more extra nesting

marineam:
LGTM

PR #155, opened 2015-10-21
kola: print test name when it is actually run

mischief:
previously test names were being printed before being submitted into the
runner's queue, so you would observe a test name being printed much
before it actually began executing, leading to confusion or ordering and
parallelism.


vcaputo:
lgtm

PR #160, opened 2015-10-27
cmd/kola: add spawn subcommand

mischief:


cmd/kola/spawn.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+}
+
+func die(format string, args ...interface{}) {
+	fmt.Fprintf(os.Stderr, format+"\n", args...)
+	os.Exit(1)
+}
+
+func slurp(file string) ([]byte, error) {
```
pbx0: Same functionality as `ioutil.Readfile`?

mischief: fixed, ptal

pbx0:
Code seems fine, but might be better to refactor this into `ore` as a multi-platform helper tool. Otherwise, this might grow into many other kola subcommands. For spawning gce instances in ore,  I found additional commands useful like `list-images` and even `upload` all being available in one command for manual testing and spawning from the sdk.

At the same time, ore ends up sharing (copy-pasta) code with plume but designed with a different user-experience in mind.  So, I don't have a good plan at the moment for unifying all this code but also keeping the user-interfaces clean. Since, I don't plan to refactor this all at the moment, I think its fine if you just want to merge this for now, but it'd be nice to organize all this functionality better in the future.

mischief:
we can refactor it later.

PR #162, opened 2015-10-27
refactor ssh

mischief:
sort of a heavy-handed approach, but now we establish a new ssh connection for each command executed. this makes it possible to deal with reboots during a test.

added a bit of documentation to Machine and Cluster, and deduplicated the common machine sanity check code.

removed NewCommand from Cluster. not really needed.

### DIFF
--- a/platform/gce.go
+++ b/platform/gce.go
@@ -55,11 +55,10 @@ type gceCluster struct {
 }
 
 type gceMachine struct {
-	gc        *gceCluster
-	name      string
-	intIP     string
-	extIP     string
-	sshClient *ssh.Client
+	gc    *gceCluster
+	name  string
+	intIP string
+	extIP string
 }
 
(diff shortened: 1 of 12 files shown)


mischief:
yay, i deleted code.

platform/gce.go
```diff
@@ -502,10 +483,6 @@ func sshCheck(gm *gceMachine) error {
 
 	// Allow a few authentication failures in case setup is slow.
 	sshchecker := func() error {
-		gm.sshClient, err = gm.gc.sshAgent.NewClient(gm.IP())
-		if err != nil {
-			return err
-		}
 		return nil
 	}
```
pbx0: Looks like this `sshCheck` func is now a no-op

pbx0:
Not sure what the original purpose of `cluster.NewCommand` was intended for, but i'm ok with its removal.

LGTM

PR #163, opened 2015-10-27
kola: add docker test

mischief:
not sure we want to merge this as a test that automatically executes, because it took 45 minutes to run for me on AWS on a `c3.2xlarge`.


pbx0:
I think the bar for usefulness on this will be making sure it tests against the same version of docker that is in the image. Whether we go some crazy vendor route or just detect and compile the image version of Docker correctly.

mischief:
@pbx0 we could inspect `docker version` output and get the git commit, and call `makeTestBranch` with that.

PR #170, opened 2015-11-02
cork: run repo init and sync inside chroot after creation

marineam:


cmd/cork/create.go
```diff
@@ -40,6 +45,8 @@ func init() {
 		"sdk-version", "", "SDK version")
 	createCmd.Flags().StringVar(&chrootName,
 		"chroot", "chroot", "SDK chroot directory name")
+	createCmd.Flags().StringVar(&manifestURL,
+		"maniest-url", coreosManifestURL, "Manifest git repo location")
```
mischief: `manifest-url`

marineam: fixed

mischief:
lgtm

PR #172, opened 2015-11-05
cork: create and enter SDK

marineam:
The create and enter commands should be fully functional now. The main issue I can think of that may be an issue for developers is the initial repo sync doesn't support passing an ssh agent into the chroot so git ssh urls cannot be used but the user could re-run repo init after the initial sync to switch. That shouldn't be a problem for automated builds though.


pbx0:
lgtm

vcaputo:
lgtm

PR #173, opened 2015-11-05
platform/gce: overwriting image should wait on delete op

pbx0:


mischief:
shouldn't you check err before op?

pbx0:
Doesn't matter in this case.  If `op != nil` then `err ==nil`.

mischief:
lgtm

PR #176, opened 2015-11-11
kola: rename CloudConfig to UserData

mischief:
it no longer makes sense to call this CloudConfig now that we support ignition.


vcaputo:
lgtm

PR #179, opened 2015-11-11
support google storage downloads from cork download-image

pbx0:


pbx0:
PTAL @mischief -- squashed commit and removed debug line

@marineam

mischief:
lgtm

sdk/download.go
```diff
@@ -50,15 +52,38 @@ func TarballURL(version string) string {
...
+	plog.Infof("Downloading %s to %s", fileURL, file)
 
+	// handle bucket urls by using api to get media link
+	parseURL, err := url.Parse(fileURL)
+	if err != nil {
+		return err
+	}
+	if parseURL.Scheme == "gs" {
+		if client == nil {
+			return fmt.Errorf("DownloadFile requires non-nil client with gs:// links")
```
marineam: The error message makes it sound like a programming error which should be a panic but since the URL is probably user provided I'd switch the if statements and only try to handle gs if client is not nil. Or you can just use the default if it is possible to use the api auth-free for public buckets/objects.

sdk/download.go
```diff
@@ -50,15 +52,38 @@ func TarballURL(version string) string {
...
+		if err != nil {
+			plog.Fatal(err)
+		}
+		path := strings.TrimLeft(parseURL.Path, "/")
+		obj, err := api.Objects.Get(parseURL.Host, path).Do()
+		if err != nil {
+			plog.Fatal(err)
+		}
+		plog.Noticef("%v", fileURL)
+		fileURL = obj.MediaLink
```
marineam: Interesting, iirc last time we tried using media link directly and passing it to the go http client go screwed up escaping and broke the link. I wonder if go or the GS service has been fixed.

sdk/download.go
pbx0: Seems to work fine with public gs:// links with the http default client. I can just remove the logic here since client isn't used until `downloadFile` is called which already sets it to DefaultClient when nil.

sdk/download.go
pbx0: You mentioned there would be dragons here. Not sure what changed, but it seemed to work fine for me which meant less special code for the gs:// case.

sdk/download.go
pbx0: Nevermind, I can't delete the logic because you need client to get the medialink. Still, just using default client seems right here.

pbx0:
PTAL -- client gets set to http.DefaultClient when nil now for gs links. This works fine for getting the medialink on public buckets.

marineam:
LGTM

PR #185, opened 2015-11-13
build: clean up build script

crawford:
I'd like to be able to build from one directory lower by running `../build`.


marineam:
LGTM

PR #189, opened 2015-11-14
version: fix golint complaints

mischief:
the following complaints were fixed:

version/version.go:3:7: exported const Version should have comment or be unexported


pbx0:
lgtm

PR #190, opened 2015-11-14
auth: add comments to make golint happy

pbx0:


auth/debug.go
```diff
@@ -26,7 +26,7 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-// Logging HTTP transport
+// Package auth provides HTTP transport logging
```
mischief: ... provides authenticated http transports? :p

mischief:
can you copy the things fixed into the commit message for posterity?

auth/debug.go
pbx0: Oh yeah, not sure what I was thinking. I just need to delete that package comment and add a doc.go or a general comment to google.go

pbx0:
All things golint complained about were just about comments. Package comments need to start with Package $name ...  Function comments need to start with $functionname

auth/debug.go
mischief: yes, i started on this and moved the comment to auth/google.go since it seemed more appropriate than in auth/debug.go, but in any case, some accurate package comment would be nice.

mischief:
lgtm

PR #191, opened 2015-11-14
kola/test/metadata: add coreos-metadata

crawford:


kola/tests/metadata/contents.go
```diff
@@ -0,0 +1,96 @@
...
+		                   ]
+		               }
+		           }`,
+	})
+
+	register.Register(&register.Test{
+		Name:        "coreos.metadata.azure",
+		Run:         verifyAzure,
+		ClusterSize: 1,
+		Platforms:   []string{"azure"},
```
mischief: we don't even support azure :C

kola/tests/metadata/contents.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/coreos/mantle/platform"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Name:        "coreos.metadata.aws",
+		Run:         verifyAWS,
+		ClusterSize: 1,
+		Platforms:   []string{"aws"},
+		UserData: `{
```
mischief: looks like you could dedup these two userdatas?

pbx0:
lgtm

kola/tests/metadata/contents.go
pbx0: This will essentially just disable the test for now.

PR #195, opened 2015-11-14
cmd: fix golint complaints

mischief:


cmd/kola/kola.go
```diff
@@ -77,36 +77,36 @@ func runRun(cmd *cobra.Command, args []string) {
...
-	for _, item := range list {
+	for _, item := range testlist {
 		fmt.Fprintf(w, "%v\n", item)
 	}
 	w.Flush()
 }
 
-type Item struct {
+type item struct {
 	Name      string
```
pbx0: un-export these whlie we are at it?

PR #197, opened 2015-11-16
fmt, vet and test code prior to building

mischief:


pbx0:
lgtm

mischief:
not done yet. i need to fix up the tests.

.travis.yml
```diff
@@ -1,8 +1,16 @@
...
+  include:
+    - go: 1.4.2
+      install:
+        - go get golang.org/x/tools/cmd/cover
+        - go get golang.org/x/tools/cmd/vet
+    - go: 1.5.1
 
 # Gracefully handle developer forks
 install: |
+    go get golang.org/x/tools/cmd/vet
```
marineam: why is this line here?

.travis.yml
```diff
@@ -10,3 +18,8 @@ install: |
         export TRAVIS_BUILD_DIR="${MANTLE_DIR}"
         cd "${TRAVIS_BUILD_DIR}"
     fi
+
+script:
+  - ./test
```
marineam: why this order?

.travis.yml
mischief: ptal 8-D

.travis.yml
mischief: no particular reason. should it be the other way?

marineam: Not that it matters in this case because the two do not share artifacts but I just usually expect to see something built before tested :-P

mischief:
@marineam ptal

build
```diff
@@ -1,26 +1,16 @@
...
 
 if [[ $# -eq 0 ]]; then
 	set -- cmd/*
 fi
+
 for cmd in "$@"; do
 	cmd=$(basename "${cmd}")
 	echo "Building ${cmd}..."
-	go build -o "bin/${cmd}" "${REPO_PATH}/cmd/${cmd}"
+	go build -o "bin/${cmd}" -ldflags "${GLDFLAGS}" "${REPO_PATH}/cmd/${cmd}"
```
marineam: Why add ldflags?

env
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
+
+ORG_PATH="github.com/coreos"
+PROJ="mantle"
+REPO_PATH="${ORG_PATH}/${PROJ}"
+VERSION=$(git describe --dirty --always)
+GLDFLAGS="-X github.com/coreos/fleet/version.Version=\"${VERSION}\""
```
marineam: fleet? delete please. If we want to start tagging and setting versions from tags that needs to be in a different PR

mischief: WOOPS :-)

mischief:
@marineam ptal.

marineam:
I still think that test script sucks but oh well.

LGTM

mischief:
@marineam so does the whole godep/gopath symlink nonsense, but it's what we have. i just want to go fmt/vet/test in travis.

marineam:
@mischief agreed, I don't use the build script either usually, it doesn't do incremental builds right so `go build` with a normal GOPATH setup is usually faster.

PR #198, opened 2015-11-16
sdk: reverse wait order of bz2 and tar

marineam:
If sudo or tar fails the bunzip process will happily hang there waiting
for something to read its output. On the flip side if bunzip aborts
early tar itself should also fail and complain about bad input. So this
ordering should be reliable.


mischief:
lgtm

PR #199, opened 2015-11-17
network/ntp: hide example ntpd main pkg

pbx0:


mischief:
can't you simply rename ntpd.go to example_ntpd.go?

mischief:
a la https://blog.golang.org/examples

pbx0:
It doesn't terminate nicely like an example would and simply changing the name to example_ntpd.go doesn't make go tools ignore it:
`Directory and file names that begin with "." or "_" are ignored by the go tool, as are directories named "testdata".`
I could change its main to `ExampleNTPD()` and its package to ntpd, but then its dead code and not suitable for use as an Example in the go sense.

pbx0:
reworked the naming a bit

mischief:
lgtm

PR #201, opened 2015-11-18
sdk: create debian_chroot flag file

marineam:
The build scripts check this file to determine if they are inside the
chroot or not, optionally entering the chroot if needed. The optional
entering part uses cros_sdk but the inside chroot check needs to work.


mischief:
lgtm

PR #205, opened 2015-11-26
kola/tests/ignition: fix xfsroot mkfs options

vcaputo:
"-L ROOT" must be supplied in two separate arguments, mkfs.xfs
erroneously accepts this form but will result in the label " ROOT"
in my tests.


mischief:
lgtm

crawford:
heh

PR #208, opened 2015-12-08
Private repo support

bassamtabbara:
This PR adds support for private repositories in the manifest repo, and reduces the prompting done by repo init. 

When an ssh-agent is being used we mount the directory holding the ssh-agent socket in the chroot and copy ~/.ssh/config and ~/.ssh/known_hosts. 

Also ~/.gitconfig is copied into the chroot to avoid repo init prompts and support automated build environments.

Lastly, I added support for specifying a manifest branch to "cork create".

### DIFF
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -24,11 +24,13 @@ const (
 )
 
 var (
-	chrootVersion string
-	chrootName    string
-	manifestURL   string
-	allowReplace  bool
-	createCmd     = &cobra.Command{
+	chrootVersion  string
+	chrootName     string
+	manifestURL    string
+	manifestName   string
+	manifestBranch string
+	allowReplace   bool
+	createCmd      = &cobra.Command{
 		Use:   "create",
 		Short: "Download and unpack the SDK",
 		Run:   runCreate,
@@ -52,6 +54,10 @@ func init() {
 		"chroot", "chroot", "SDK chroot directory name")
 	createCmd.Flags().StringVar(&manifestURL,
 		"manifest-url", coreosManifestURL, "Manifest git repo location")
+	createCmd.Flags().StringVar(&manifestBranch,
+		"manifest-branch", "master", "Manifest git repo branch")
+	createCmd.Flags().StringVar(&manifestName,
+		"manifest-name", "default.xml", "Manifest file name")
 	createCmd.Flags().BoolVar(&allowReplace,
 		"replace", false, "Replace an existing SDK chroot")
 	enterCmd.Flags().StringVar(&chrootName,
@@ -91,7 +97,7 @@ func runCreate(cmd *cobra.Command, args []string) {
 		plog.Fatalf("Create failed: %v", err)
 	}
 
-	if err := sdk.RepoInit(chrootName, manifestURL); err != nil {
+	if err := sdk.RepoInit(chrootName, manifestURL, manifestName, manifestBranch); err != nil {
 		plog.Fatalf("repo init and sync failed: %v", err)
 	}
 }
--- a/sdk/enter.go
+++ b/sdk/enter.go
@@ -256,8 +256,10 @@ func Enter(name string, args ...string) error {
 	return enterCmd.Run()
 }
 
-func RepoInit(name, manifest string) error {
-	if err := SimpleEnter(name, "repo", "init", "-u", manifest); err != nil {
+func RepoInit(name, manifest, manifestName, branch string) error {
+	if err := SimpleEnter(
+		name, "repo", "init", "-u", manifest,
+		"-b", branch, "-m", manifestName); err != nil {
 		return err
 	}
 
--- a/sdk/enter.go
+++ b/sdk/enter.go
@@ -16,9 +16,11 @@ package sdk
 
 import (
 	"fmt"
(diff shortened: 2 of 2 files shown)


sdk/enter.go
```diff
@@ -80,6 +80,62 @@ func simpleChrootHelper(args []string) error {
...
+				return err
+			}
+
+			if err := syscall.Mount(
+				sshSourceDir, sshTargetdir, "none", syscall.MS_BIND, ""); err != nil {
+				return fmt.Errorf("Mounting %q failed: %v", sshSourceDir, err)
+			}
+		}
+
+		if err := os.MkdirAll(filepath.Join(chroot, ".ssh"), 0700); err != nil {
```
marineam: Everything from here down doesn't need to run as root and in fact probably shouldn't so the copied files get the right ownership. Also it should be perfectly fine to copy ssh configs even if there isn't an agent socket, iirc that's what the usual enter_chroot.sh script does. So the rest of this can get move to another function that gets called before `sudo.Run()` in `SimpleEnter`. Also, given the common boiler plate for all the copies a helper `copyToChroot` would help clean that up.

sdk/enter.go
```diff
@@ -80,6 +80,62 @@ func simpleChrootHelper(args []string) error {
...
 	}
 
+	homedir := os.Getenv("HOME")
+	if homedir == "" {
+		return fmt.Errorf("HOME environment variable is not set.")
+	}
+
+	// mount the directory containing the ssh-agent socket in order
+	// to support a private repos during repo sync
+	sshAuthSock := os.Getenv("SSH_AUTH_SOCK")
```
marineam: Using the same value of `$SSH_AUTH_SOCK` both inside and outside the chroot is what `enter_chroot.sh` does but it would probably be cleaner if inside the chroot the path was always bind to `/run/user/1000/ssh-XXXXXX` (create a randomly named temporary dir) and then update the value of `$SSH_AUTH_SOCK`

We should also match `enter_chroot.sh` by adding the equivilent of `mount -t tmpfs -o nosuid,nodev,mode=755 tmpfs /run` before dropping things into /run.

marineam: Although considering redefining `$SSH_AUTH_SOCK` is a behavior change from `enter_chroot.sh` that could come later. As functionality moves into Go code matching the old behavior is sufficient for now.

cmd/cork/create.go
```diff
@@ -52,6 +53,8 @@ func init() {
 		"chroot", "chroot", "SDK chroot directory name")
 	createCmd.Flags().StringVar(&manifestURL,
 		"manifest-url", coreosManifestURL, "Manifest git repo location")
+	createCmd.Flags().StringVar(&manifestBranch,
+		"manifest-branch", "master", "Manifest git repo branch")
```
marineam: While we are at it might as well pass through `--manifest-name=` as well, making this work for CoreOS release branches which need `-b build-123 -m release.xml` to set things up correctly.

bassamtabbara:
@marineam I think I incorporated all your feedback. let met know if there's anything else.

sdk/enter.go
```diff
@@ -80,6 +82,73 @@ func simpleChrootHelper(args []string) error {
...
+	if sshAuthSock != "" {
+		sshSourceDir, sshSocketFile := filepath.Split(sshAuthSock)
+		if _, err := os.Stat(sshSourceDir); err == nil {
+
+			// remove old ssh- directories
```
marineam: No need for this since you are in a clean `/run` directory while this bit of code in `enter_chroot.sh` was cleaning up `/tmp` which isn't tmpfs in the chroot currently.

sdk/enter.go
```diff
@@ -80,6 +82,73 @@ func simpleChrootHelper(args []string) error {
...
+			if err := rm.Run(); err != nil {
+				return err
+			}
+
+			sshTargetdir, err := ioutil.TempDir(rundir, "ssh-")
+			if err != nil {
+				return err
+			}
+
+			err = os.Chown(sshTargetdir, uid, gid)
```
marineam: Since this is a mount target the ownership of the underlying directory doesn't really matter, it will be hidden by the mount point.

marineam:
Just two code comment left. Looks like `gofmt` is grumpy about something so be sure to run that too. Once you are all set go ahead and squash the commits into two "add ssh and user configs" and "add manifest args" commits, no need for the full history of this review in the final log.

Thanks!

sdk/enter.go
bassamtabbara: I've seen multiple /run/user/1000/ssh-xxxxx left behind in the chroot. I wonder if its related to calling SimpleEnter multiple times.

mischief:
ok to test

mischief:
test this please

sdk/enter.go
bassamtabbara: ok I found the issue. the mount target for /run above was wrong.

marineam: Doh!

bassamtabbara:
Ok I fix the remaining issues and squashed the commits. I've tested this on our build system with lots of private repos and it works well.

marineam:
LGTM, thanks!

Issue #171, opened 2015-11-03
kola: check coreos semver


mischief:
kola tests should be able to specify which CoreOS versions they can execute on, and if the remote machine does not have the appropriate version, the tests should be skipped.

i'm not sure what this looks like yet, since the actual test functions nor the platform receive information (*kola.Test) about the currently running test.

mischief:
appears to have been fixed in #209

PR #209, opened 2015-12-09
kola: allow tests to specify min CoreOS semver

pbx0:
This PR adds the ability to specify the minimum semantic version of CoreOS that a given test should run on.

The check is done by spinning up a machine and parsing /etc/os-release before running any tests. This is somewhat costly time-wise. In the future, we may want to simply pass in the CoreOS version by flag. 

To avoid some of the cost, the version check is ignored if the test glob is an exact match. Additionally the second commit adds two fast paths for avoiding spinning up an extra machine.

https://github.com/coreos/mantle/issues/171

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -26,6 +26,7 @@ import (
 	"github.com/coreos/mantle/kola/register"
 	"github.com/coreos/mantle/platform"
 
+	"github.com/coreos/mantle/Godeps/_workspace/src/github.com/coreos/go-semver/semver"
 	"github.com/coreos/mantle/Godeps/_workspace/src/github.com/coreos/pkg/capnslog"
 
 	// Tests imported for registration side effects.
@@ -90,7 +91,7 @@ func testRunner(platform string, done <-chan struct{}, tests chan *register.Test
 	}
 }
 
-func filterTests(tests map[string]*register.Test, pattern, platform string) (map[string]*register.Test, error) {
+func filterTests(tests map[string]*register.Test, pattern, platform string, version *semver.Version) (map[string]*register.Test, error) {
 	r := make(map[string]*register.Test)
 
 	for name, t := range tests {
@@ -104,6 +105,11 @@ func filterTests(tests map[string]*register.Test, pattern, platform string) (map
 
 		// Skip the test if Manual is set and the name doesn't fully match.
 		if t.Manual && t.Name != pattern {
+			continue
+		}
+
+		// Check the test's minVersion when running more then one test
+		if version.LessThan(t.MinVersion) && t.Name != pattern {
 			continue
 		}
 
@@ -134,7 +140,12 @@ func RunTests(pattern, pltfrm string) error {
 	var passed, failed, skipped int
 	var wg sync.WaitGroup
 
-	tests, err := filterTests(register.Tests, pattern, pltfrm)
+	semver, err := getClusterSemver(pltfrm)
+	if err != nil {
+		plog.Fatal(err)
+	}
+
+	tests, err := filterTests(register.Tests, pattern, pltfrm, semver)
 	if err != nil {
 		plog.Fatal(err)
 	}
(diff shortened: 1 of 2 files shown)


kola/register/register.go
```diff
@@ -30,18 +36,28 @@ type Test struct {
...
 
 // Register is usually called in init() functions and is how kola test
-// harnesses knows which tests it can choose from. Panic if existing
+// harnesses knows which tests it can choose from. Panics if existing
 // name is registered
 func Register(t *Test) {
 	_, ok := Tests[t.Name]
 	if ok {
-		panic("test already registered with same name")
+		plog.Fatal("test already registered with same name")
```
mischief: this should probably stay as panic so it is easier to identify the caller

pbx0: back to panic now

kola/harness.go
```diff
@@ -107,6 +109,17 @@ func filterTests(tests map[string]*register.Test, pattern, platform string) (map
 			continue
 		}
 
+		var minVersion *semver.Version
+		if t.MinVersion == "" {
+			minVersion, _ = semver.NewVersion("0.0.0")
+		} else {
+			minVersion, _ = semver.NewVersion(t.MinVersion) // err checked in test registration
+		}
+		plog.Infof("version check %v is less than %v : %v", version, t.MinVersion, version.LessThan(*minVersion))
```
mischief: can this be debug instead of info?

pbx0: doh, I meant to remove that completely

pbx0:
retest this please

pbx0:
retest this please

pbx0:
retest this please

pbx0:
retest this please

mischief:
lgtm. can we use this for anything currently?

pbx0:
I'm not sure if we have an immediate use case for this. This is probably only useful if we have tests that only pass on some of our releases. We've had this situation in the past, but I think fixes have trickled through all our current release channels. 

I will defer merging to someone who plans to use this.

kola/harness.go
```diff
@@ -147,14 +148,39 @@ func RunTests(pattern, pltfrm string) error {
 	var passed, failed int
 	var wg sync.WaitGroup
 
-	semver, err := getClusterSemver(pltfrm)
+	// Avoid incurring cost of starting machine in getClusterSemver when
+	// either:
+	// 1) we already know 0 tests will run
+	// 2) glob is an exact match which means minVersion will be ignored
+	//    either way
+	maxVersion, _ := semver.NewVersion(strconv.FormatInt(math.MaxInt64, 10) + ".0.0")
```
crawford:
```go
maxVersion := semver.Version{Major: math.MaxInt64}
```

kola/harness.go
```diff
@@ -107,6 +108,18 @@ func filterTests(tests map[string]*register.Test, pattern, platform string) (map
 			continue
 		}
 
+		var minVersion *semver.Version
+		if t.MinVersion == "" {
```
crawford: If `t.MinVersion` where a `semver.Version` instead of a `string`, this entire block could be removed.

crawford:
We need this for the Ignition v1-to-v2.0.0 transition.

kola/harness.go
pbx0: Looking back on this, I think I went through this janky step at the time because its slightly more ergonomic for a test writer to specify `MinVersion: "v1.1.1"` then `Min Version: semver.Must(semver.NewVersion("v1.1.1")`. 

Its certainly more simple in the code to use a `semver.Version` though. Do you still feel strongly either way?

crawford: I much prefer `Min Version: semver.Must(semver.NewVersion("v1.1.1")`.

kola/harness.go
pbx0: done

pbx0:
@crawford PTAL -- addressed both comments. 

The syntax for declaring the minimum version in a test will actually be `MinVersion: semver.Version{Major: 1, Minor: 1, Patch:1}` since its not a pointer. Could make it a pointer if you prefer the `semver.Must(semver.NewVersion...` Makes no difference to me.

crawford:
I actually like that better. Make the compiler do some work for once in its life.

crawford:
LGTM

PR #213, opened 2015-12-15
system: add a friendlier mount API

marineam:
The mix of MS_* flags and extra string options makes the plain mount(2)
system call kinda ugly. To clean up code a bit this interface behaves
more like the mount(8) command instead, accepting flags as strings too.


mischief:
LGTM

PR #217, opened 2015-12-22
network/omaha: add no-op server

mischief:


network/omaha/protocol.go
```diff
@@ -48,6 +48,8 @@ type Request struct {
 
 func NewRequest() *Request {
 	return &Request{
+		// silly hack, xml's Encoder doesn't want to set this.
```
marineam: Hm, could you explain further? I haven't ever needed to do this

network/omaha/server.go
```diff
@@ -0,0 +1,130 @@
...
+	reqChanLock sync.Mutex
+	reqChan     chan *Request
+}
+
+func (s *Server) Serve() error {
+	err := s.srv.Serve(s.l)
+	if err != nil {
+		netErr, ok := err.(*net.OpError)
+		if ok {
+			if netErr.Err.Error() == "use of closed network connection" {
```
marineam: Aww man, is string comparison really the only way to do this?

network/omaha/server.go
```diff
@@ -0,0 +1,130 @@
...
+func (s *Server) RequestChan() <-chan *Request {
+	s.reqChanLock.Lock()
+	if s.reqChan == nil {
+		s.reqChan = make(chan *Request)
+	}
+	s.reqChanLock.Unlock()
+	return s.reqChan
+}
+
+func (s *Server) UpdateHandler(rw http.ResponseWriter, r *http.Request) {
```
marineam: I was about to ask why you didn't use the handler I implemented, but then realized I hadn't settled on its API and so was still in a branch on my laptop. Doh. I hadn't considered the use of a channel though, maybe that will clear it up.

Anyway, this is a fine start, can port to the more feature-complete handler later.

network/omaha/protocol.go
mischief: the xml encoder does not set this but the decoder does, leading to a diff in the structs when i compare them in server_test.go.

see https://travis-ci.org/coreos/mantle/jobs/98414566 for the diff output

network/omaha/server.go
mischief: my other idea besides a channel was just stick an overrideable function in the Server struct. called if non-nil with the Request, and the Response used verbatim if non-nil. thoughts?

network/omaha/server.go
```diff
@@ -0,0 +1,130 @@
...
+	s.reqChanLock.Lock()
+	if s.reqChan == nil {
+		close(s.reqChan)
+	}
+	s.reqChanLock.Unlock()
+}
+
+// RequestChan returns a channel that will receive decoded omaha requests.
+//
+// XXX: this can only be called once per Server instance, otherwise the
```
marineam: This doesn't seam accurate, it is safe to call multiple times but there is only channel so it is unlikely to be useful to do so.

network/omaha/server_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/coreos/mantle/Godeps/_workspace/src/github.com/kylelemons/godebug/pretty"
+)
+
+func TestServerRequestResponse(t *testing.T) {
+	// make an omaha server
+	s, err := NewServer(":34567")
```
marineam: Tests shouldn't use hard-coded ports, use ":0" and get the local address from the listener. So either NewServer should do this all the time and update Server.Addr or the api needs to be switched around to offer more control over the server's listener.

network/omaha/protocol.go
marineam: Ah, ok. Well this has nothing to do with the encoder which obviously has no business modifying the structure given to it. So either update the comment to clarify that it is required to ensure consistency with the output of the decoder or add it to the test itself since a full struct comparison is the only place this would ever matter.

network/omaha/server.go
marineam: I started on a similar path, with a special update data structure and updater interface. I didn't use the Request and Response verbatim so the updater interface could deal with apps independently.

Lets stick with this for now, I'll dust off my code after the holidays and see how this fits together.

network/omaha/server.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/coreos/mantle/Godeps/_workspace/src/gopkg.in/tylerb/graceful.v1"
+)
+
+func NewServer(addr string) (*Server, error) {
+	l, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+
+	srv := &graceful.Server{
```
marineam: I'm guessing we want to set `NoSignalHandling` to true, seems bad form for libraries to go clobering signal handlers as this one appears to do.

marineam:
I'm not really sold on the need for `graceful.Server` in this code, seems like just exposing access to listener.Close() would be sufficient.

network/omaha/server.go
mischief: signal.Notify won't clobber; all callers are notified.

"It is allowed to call Notify multiple times with different channels and the same signals: each channel receives copies of incoming signals independently."

https://godoc.org/os/signal#Notify

network/omaha/server.go
mischief: yes, unfortunately this error is not typed nor is the instance of the error exported for comparison.

```
net/net.go:284:     errClosing                = errors.New("use of closed network connection")
```

mischief:
ptal

network/omaha/server.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return nil
+}
+
+func (s *Server) Stop() {
+	s.l.Close()
+
+	s.reqChanLock.Lock()
+	if s.reqChan == nil {
```
marineam: This test looks backwards

network/omaha/server.go
```diff
@@ -0,0 +1,125 @@
...
+		close(s.reqChan)
```
marineam: Add `recChan = nil` to ensure a waiting handler doesn't try to write to the closed channel.

marineam:
Just a little bit of fixing in Stop, otherwise looks good.

mischief:
ptal

marineam:
[LGTM](http://lgtm.in/i/cUzMKHT6m)

PR #219, opened 2015-12-28
Add gid=5 to /dev/pts mounting options

bassamtabbara:
Without this I was seeing the following error when calling update_chroot:

```
* In order to use glibc with USE=-suid, you must make sure that
 * you have devpts mounted at /dev/pts with the gid=5 option.
 * Openrc should do this for you, so you should check /etc/fstab
 * and make sure you do not have any invalid settings there.
 * ERROR: sys-libs/glibc-2.21-r1::coreos failed (pretend phase):
 *   mount & fix your /dev/pts settings
 * 
 * Call stack:
 *     ebuild.sh, line   93:  Called pkg_pretend
 *   environment, line 3012:  Called eblit-run 'pkg_pretend'
 *   environment, line 1108:  Called eblit-glibc-pkg_pretend
 *   environment, line  967:  Called check_devpts
 *   environment, line  790:  Called die
 * The specific snippet of code:
 *               die "mount & fix your /dev/pts settings";
 * 
 * If you need support, post the output of `emerge --info '=sys-libs/glibc-2.21-r1::coreos'`,
 * the complete build log and the output of `emerge -pqv '=sys-libs/glibc-2.21-r1::coreos'`.
 * The complete build log is located at '/var/log/portage/sys-libs:glibc-2.21-r1:20151228-154637.log'.
 * For convenience, a symlink to the build log is located at '/var/tmp/portage/sys-libs/glibc-2.21-r1/temp/build.log'.
 * The ebuild environment file is located at '/var/tmp/portage/sys-libs/glibc-2.21-r1/temp/environment'.
 * Working directory: '/usr/lib64/python2.7/site-packages'
 * S: '/var/tmp/portage/sys-libs/glibc-2.21-r1/work/glibc-2.21'
```

I'm not quite sure what changed but this seems to workaround the issue.

### DIFF
--- a/sdk/enter.go
+++ b/sdk/enter.go
@@ -46,8 +46,8 @@ type enter struct {
 }
 
 // wrapper for bind mounts to report errors consistently
-func bind(src, dst string) error {
-	err := syscall.Mount(src, dst, "none", syscall.MS_BIND, "")
+func bind(src, dst, extra string) error {
+	err := syscall.Mount(src, dst, "none", syscall.MS_BIND, extra)
 	if err != nil {
 		return fmt.Errorf("Binding %q to %q failed: %v", src, dst, err)
 	}
@@ -56,7 +56,7 @@ func bind(src, dst string) error {
 
 // bind and remount read-only for safety
 func bindro(src, dst string) error {
-	if err := bind(src, dst); err != nil {
+	if err := bind(src, dst, ""); err != nil {
 		return err
 	}
 	if err := syscall.Mount(src, dst, "none", syscall.MS_REMOUNT|syscall.MS_BIND|syscall.MS_RDONLY, ""); err != nil {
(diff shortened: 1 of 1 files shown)


sdk/enter.go
```diff
@@ -99,7 +99,7 @@ func (e *enter) MountAPI() error {
 		return err
 	}
 	// /dev/pts must be read-write because emerge chowns tty devices.
-	if err := bind("/dev/pts", filepath.Join(e.Chroot, "dev/pts")); err != nil {
+	if err := bind("/dev/pts", filepath.Join(e.Chroot, "dev/pts"), "gid=5"); err != nil {
```
marineam: Huh, I'm unclear how this helps. As far as I know it isn't possible to change any mount options at the same time as performing a bind. Also since gid is an option for the devpts filesystem, if it is applied at all I would expect it to apply to the system globally which is probably going to be undesirable for the host system. An alternative would be to mount a new instance of the devpts filesystem but I don't know if that has other problems considering chroot isn't a fully isolated container.

We may be best off just updating the glibc ebuild to not raise this error, I just need to double check that the SDK doesn't actually care that much about /dev/pts.

marineam:
I've posted https://github.com/coreos/coreos-overlay/pull/1705 as an alternative way to fix this issue.

marineam:
Merged  coreos/coreos-overlay#1705 and started a release build to make the fixed binary package available.

bassamtabbara:
thank you!

PR #220, opened 2015-12-28
Make cork's --sdk-version optional

bassamtabbara:
cork create --sdk-version is now optional. If not specified we will read COREOS_SDK_VERSION from version.txt in the manifest repo. 

I'm shelling out to "git clone" instead of using the git golang lib, hope thats OK.

### DIFF
--- a/sdk/download.go
+++ b/sdk/download.go
@@ -190,7 +190,7 @@ func DownloadSignedFile(file, url string, client *http.Client) error {
 }
 
 func DownloadSDK(version string) error {
-	tarFile := filepath.Join(RepoCache(), "sdk", TarballName(version))
+	tarFile := filepath.Join(RepoCache(), "sdks", TarballName(version))
 	tarURL := TarballURL(version)
 	return DownloadSignedFile(tarFile, tarURL, nil)
 }
--- a/sdk/version.go
+++ b/sdk/version.go
@@ -17,7 +17,9 @@ package sdk
 import (
 	"bufio"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
@@ -27,9 +29,7 @@ const (
 	coreosId = "{E96281A6-D1AF-4BDE-9A0A-97B76E56DC57}"
 )
 
-func GetVersion(dir string) (ver string, err error) {
-	const key = "COREOS_VERSION="
-
+func getVersion(dir, key string) (ver string, err error) {
 	f, err := os.Open(filepath.Join(dir, "version.txt"))
 	if err != nil {
 		return
@@ -53,8 +53,36 @@ func GetVersion(dir string) (ver string, err error) {
 	return
 }
 
+func GetVersionFromDir(dir string) (string, error) {
+	return getVersion(dir, "COREOS_VERSION=")
+}
+
 func GetLatestVersion() (string, error) {
-	return GetVersion(BuildImageDir("latest"))
+	return getVersion(BuildImageDir("latest"), "COREOS_VERSION=")
+}
+
+func GetSDKVersion() (string, error) {
+	return getVersion(filepath.Join(RepoRoot(), ".repo", "manifests"), "COREOS_SDK_VERSION=")
+}
+
+func GetSDKVersionFromDir(dir string) (string, error) {
+	return getVersion(dir, "COREOS_SDK_VERSION=")
+}
+
+func GetSDKVersionFromRemoteRepo(url, branch string) (string, error) {
+	tmp, err := ioutil.TempDir("", "")
+	if err != nil {
+		return "", err
+	}
+	defer os.RemoveAll(tmp)
+
+	clone := exec.Command("git", "clone", "-q", "--depth=1", "--single-branch", "-b", branch, url, tmp)
+	clone.Stderr = os.Stderr
+	if err := clone.Run(); err != nil {
+		return "", err
+	}
+
+	return GetSDKVersionFromDir(tmp)
 }
 
 func GetDefaultAppId() string {
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -50,7 +50,7 @@ var (
 
 func init() {
 	createCmd.Flags().StringVar(&chrootVersion,
-		"sdk-version", "", "SDK version")
+		"sdk-version", "", "SDK version. Defaults to the SDK version in version.txt")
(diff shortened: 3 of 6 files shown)


sdk/download.go
```diff
@@ -318,3 +319,18 @@ func UpdateSignedFile(file, url string, client *http.Client) error {
...
 }
+
+func DetectSDKVersion(url, branch string) (string, error) {
+	tmp, err := ioutil.TempDir("", "")
+	if err != nil {
+		return "", err
+	}
+	defer os.RemoveAll(tmp)
+
+	clone := exec.Command("git", "clone", "--depth=1", "--single-branch", "-b", branch, url, tmp)
```
marineam: At the very least connect stderr so a failure can be diagnosed.

bassamtabbara: Ok. its noisy though which is why I removed it :-)

bassamtabbara: This is what it looks like:

```
2015-12-28T23:16:40Z cork: Detecting SDK version
Cloning into '/tmp/732493167'...
remote: Counting objects: 9, done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 9 (delta 0), reused 5 (delta 0), pack-reused 0
Unpacking objects: 100% (9/9), done.
Checking connectivity... done.
```

sdk/version.go
```diff
@@ -53,8 +51,24 @@ func GetVersion(dir string) (ver string, err error) {
 	return
 }
 
+func GetVersion() (string, error) {
```
marineam: Remove, this is just duplicating `GetLatestVersion` which seems like the better name anyway.

sdk/version.go
```diff
@@ -53,8 +51,24 @@ func GetVersion(dir string) (ver string, err error) {
...
+	return getVersion(dir, "COREOS_VERSION=")
+}
+
 func GetLatestVersion() (string, error) {
-	return GetVersion(BuildImageDir("latest"))
+	return getVersion(BuildImageDir("latest"), "COREOS_VERSION=")
+}
+
+func GetSDKVersion() (string, error) {
+	return getVersion(BuildImageDir("latest"), "COREOS_SDK_VERSION=")
```
marineam: I'm not sure this would ever be useful, but it would be useful to have a shorthand for getting the SDK version from `.repo/manifests`, perhaps change this to do that instead.

sdk/download.go
marineam: Ah, and git doesn't have a quiet option? Perhaps then use CombinedOutput and dump on error.

bassamtabbara: ok it looks like passing -q to git clone will avoid the extra noise:

```
 cork create --replace
2015-12-28T23:21:09Z cork: Detecting SDK version
2015-12-28T23:21:10Z cork: Downloading SDK version 891.0.0
```

and here it is when something is wrong

```
 cork create --replace --manifest-url=foo
2015-12-28T23:20:08Z cork: Detecting SDK version
fatal: repository 'foo' does not exist
2015-12-28T23:20:08Z cork: DetectSDKVersion failed: exit status 128
```

sdk/download.go
```diff
@@ -318,3 +319,18 @@ func UpdateSignedFile(file, url string, client *http.Client) error {
 	plog.Infof("Verified file: %s", file)
 	return nil
 }
+
+func DetectSDKVersion(url, branch string) (string, error) {
```
marineam: After https://github.com/coreos/mantle/pull/220/files#r48512571 you can call GetSDKVersion to re-use an existing repo clone instead of fetching.

sdk/download.go
marineam: You can also do things like append -q unless the log level is INFO or DEBUG

sdk/download.go
bassamtabbara: are you saying the git clone can be avoided to get the SDK version?

bassamtabbara:
ok I think I incorporated all your feedback and squashed the commits.

sdk/version.go
```diff
@@ -53,8 +51,20 @@ func GetVersion(dir string) (ver string, err error) {
...
+	return getVersion(dir, "COREOS_VERSION=")
+}
+
 func GetLatestVersion() (string, error) {
-	return GetVersion(BuildImageDir("latest"))
+	return getVersion(BuildImageDir("latest"), "COREOS_VERSION=")
+}
+
+func GetSDKVersion() (string, error) {
+	return getVersion(filepath.Join(RepoRoot(), "manifests"), "COREOS_SDK_VERSION=")
```
marineam: Repo root points to the top level of the whole tree, so this should be `Join(RepoRoot(), ".repo", "manifests")`

sdk/download.go
marineam: Still need to add the `GetSDKVersion` to avoid the git clone if the repo tree is already there. Most commonly this would be the case if `create --replace` is used.

bassamtabbara: ``create --replace``` could change the manifest branch and url, so we would need to handle that case too right?

marineam: Yes, but not sure what I want that behaviour to be. Maybe replace could start skipping the repo init/sync if it isn't needed, or at least make it optional but I think picking that behavior is orthogonal to this.

bassamtabbara: maybe I'm missing something here but I thought ```cork create``` (with or without replace) always called ```repo init``` which could retrieve a new manifest. if we get the ```COREOS_SDK_VERSION``` from .repo/manifests we could end up with the wrong version, right?

marineam: It may be behind but the subsequent update_chroot and/or build_packages will upgrade the chroot to (hopefully) the same state either way. The update_chroot command always happens as part of the create process in the old cros_sdk implementation, probably will be adding it to cork at some point too. If it is a concern maybe a `--sdk-version=check` or `git` or similar keyword could be used to always perform the git clone.

Basically my thought is to use locally cached stuff when reasonable by default, that way there are fewer things to tweak or fix when I find myself working remotely with limited or no internet access and still want my system to behave in a sane way.

bassamtabbara: Fair. I'll check the local repo first for the version and if that fails, the remote repo for now. Once the replace and repo init interactions are figured out we can change the SDK detection behavior too. 

FWIW, it would be good if cork create --replace replaced everything in the chroot to match the new branch and manifest. Maybe even use  --force-sync when doing repo sync.

marineam:
LGTM, thanks!

PR #225, opened 2016-01-07
sdk: only add core user if it does not exist

mischief:
this is needed if the core user already exists, which is the case on
CoreOS.


mischief:
dang, @mention-bot beat me to it

marineam:
LGTM

PR #227, opened 2016-01-07
kola: check /usr/bin for kolet path, since coreos ebuild puts it there

mischief:


marineam:
Hm, the intent was to put it under usr/lib/kola/$GOARCH instead

mischief:
right, because the target arch might differ from host. i'll see about fixing the ebuild then.

PR #234, opened 2016-01-15
coreos.verity.corruption

mischief:
had to move the Skip code to avoid import cycles. sticking it in register seems like a poor choice but it seemed like the best for now.


mischief:
ping @gtank @marineam

marineam:
LGTM

PR #237, opened 2016-01-23
sdk: copy google application credentials into the chroot

marineam:
The trouble of gsutil: .boto files are not always self-contained, they
may refer to service account secret keys in other files, include
arbitrary other files, and so on. The existing support for .boto in
enter.sh assumed that copying ~/.boto and ~/.boto.p12 is sufficient. We
could continue this, adding ~/.boto.json to the list to give a sane
filename to Google's newer credential file format or even begin parsing
.boto and copy/translate paths as best we can. But no matter what our
interpretation of the config will be incomplete and doesn't help with
providing credentials to other tools, including mantle.

A longer term more complete solution would be to make more use of the
new JSON format to provide credentials to all tools. This seems
promising but gsutil only partially supports it. The integration with
gcloud involves generating boto files based on the JSON file.

So either way this feels hacky but config generation sounds more
flexible than assuming paths or config parsing/translation.

This change just covers copying the credentials into the SDK chroot but
in the future the auth package should be updated to more seamlessly
integrate with this JSON format.


mischief:
lgtm

PR #239, opened 2016-02-02
version: use var so Version can be overridden by the go linker

mischief:


marineam:
LGTM, though if the tags are going to be semver maybe the default should be "0.0.0" for the sake of consistency.

mischief:
not a bad idea.

marineam:
LGTM

PR #241, opened 2016-02-09
New Multinode Kubernetes Test

pbx0:
This replaces the old multinode test we had running based off some very old and unoffcial configs. The cluster is now setup according to the coreos manual setup guides in part by using the scripts from https://github.com/coreos/coreos-kubernetes/tree/master/multi-node/generic. 

Right now we also have the kubelet running via rkt fly. Once that script is finalized and accepted in https://github.com/coreos/coreos-overlay/pull/1750 I'll update this PR to use the on-disk script.


platform/platform.go
```diff
@@ -128,6 +128,74 @@ func (t *TestCluster) DropFile(localPath string) error {
 	return nil
 }
 
+// Copy a file between two machines in a cluster.
+func TransferFile(src Machine, srcPath string, dst Machine, dstPath string) error {
```
mischief: can this be turned into e.g. `func ReadFile(m Machine, path string) (io.ReadCLoser, error)` and combined by the caller with e.g.:

```go
rc, _ := ReadFile(fromM, "/foo")
defer rc.Close()
err = InstallFile(rc, toM, "/bar")
```

mischief:
i think instances of `strings.Replace` should be using go `text/template` pkg instead.

platform/platform.go
pbx0: This makes sense, but, how do we avoid leaking sessions with this interface?

mischief: i suppose you need a io.ReadCloser wrapper around `ssh cat` that cleans up ssh on `Close()`

mischief: if you do make this change can you make it in a seperate commit?

kola/tests/kubernetes/controllerInstall.go
```diff
@@ -0,0 +1,795 @@
+package kubernetes
+
+const controllerInstallScript = `#!/bin/bash
+set -e
+
+# List of etcd servers (http://ip:port), comma separated
+export ETCD_ENDPOINTS=
+
+# Specify the version (vX.Y.Z) of Kubernetes assets to deploy
+export K8S_VER=v1.1.3
```
aaronlevy: This ideally is parameterized in some way (also we very likely start by publishing images of v1.1.2, v1.1.4, and v1.1.7)

kola/tests/kubernetes/coreosbasic.go
```diff
@@ -0,0 +1,319 @@
...
+	for _, cmd := range cmds {
+		b, err := master.SSH(cmd)
+		if err != nil {
+			return fmt.Errorf("Failed on cmd: %s with error: %s and output %s", cmd, err, b)
+		}
+	}
+	return nil
+}
+
+func generateWorkerTLSAssets(master platform.Machine, workers []platform.Machine) error {
```
aaronlevy: Is there a reason it's necessary to copy around CA assets then ssh/run openssl? This could be done directly in go with crypto/x509

kola/tests/kubernetes/coreosbasic.go
```diff
@@ -0,0 +1,319 @@
...
+	workerScript := strings.Replace(workerInstallScript, "CONTROLLER_ENDPOINT=", "CONTROLLER_ENDPOINT="+CONTROLLER_ENDPOINT, 1)
+	for _, worker := range workers {
+		if err := runInstallScript(worker, workerScript, ETCD_ENDPOINTS); err != nil {
+			return fmt.Errorf("Installing controller: %v", err)
+		}
+	}
+
+	// configure kubectl
+	if err := configureKubectl(master, MASTER_HOST); err != nil {
+		return nil
```
aaronlevy: return err?

kola/tests/kubernetes/coreosbasic.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	// check that all nodes appear in kubectl
+	f := func() error {
+		if err = nodeCheck(master, workers); err != nil {
```
aaronlevy: nit: redundant if err checks.
```
f := func() {
    return nodeCheck(master, workers)
}
```

kola/tests/kubernetes/coreosbasic.go
```diff
@@ -0,0 +1,319 @@
...
+	return nil
+}
+
+// https://coreos.com/kubernetes/docs/latest/configure-kubectl.html
+func configureKubectl(m platform.Machine, MASTER_HOST string) error {
+	var (
+		CA_CERT    = "/home/core/ca.pem"
+		ADMIN_KEY  = "/home/core/admin-key.pem"
+		ADMIN_CERT = "/home/core/admin.pem"
+		kubeURL    = "https://storage.googleapis.com/kubernetes-release/release/v1.1.3/bin/linux/amd64/kubectl"
```
aaronlevy: Should tie the version of kubectl to the rest of assets / test

kola/tests/kubernetes/controllerInstall.go
mischief: if go `text/template` is used, we can pass it thru the template and maybe test multiple versions of k8s.

pbx0: I've parameterized this (locally) to single version variable in main. I'm not sure the best way to pass that into our tests since this will be run on os releases as well. 

Maybe we just maintain a list of the versions we want to test and run a cluster for each version?

kola/tests/kubernetes/coreosbasic.go
pbx0: We could probably use crypto/x509 but it'd be far more verbose and we'd have to copy all the files to the machines anyway.

Since this test follows our install guides somewhat closely, I think its less of a burden to maintain if we just shell out to the same openssl commands.

kola/tests/kubernetes/controllerInstall.go
mischief: similar style to how https://github.com/coreos/mantle/blob/master/kola/tests/flannel/flannel.go templates different flannel backends? can have tests like `"google.kubernetes.coreosbasic.v1.1.2"` `"google.kubernetes.coreosbasic.v1.2"`

kola/tests/kubernetes/coreosbasic.go
pbx0: done

kola/tests/kubernetes/coreosbasic.go
pbx0: done

kola/tests/kubernetes/coreosbasic.go
pbx0: done

pbx0:
@aaronlevy @mischief PTAL

Fixed some small things, used text/template, and there is now a central list of version tags that are all registered. 

I also added a commit that refactors the transfer function. I am lukewarm about the results since I don't think ReadFile is going to be very useful by itself.

mischief:
is this test going to require the kubelet script that @aaronlevy and @pbx0 just wrote?

if so, the test should specify the min version of coreos required is the one that has that script.

pbx0:
It includes it for the time being, but the idea is that it will depend on the on-disk version, of course.

We could merge that PR and specify min version or we could just skip the test if we can't stat the on-disk script.

mischief:
ok, the merged code should certainly include some kind of check.

probably should allow the Skip code to include a message indicating why the test was skipped.

pbx0:
@mischief PTAL.

Since this commit deletes the previous kubernetes test, I decided to not skip when it can't find the on-disk kubelet script. Instead, if the on-disk doesn't exist, it will fallback to using the test's built-in version.  The scripts are currently identical and we can eventually remove this check.

mischief:
if @aaronlevy has no more complaints lgtm.

aaronlevy:
lgtm

PR #246, opened 2016-02-19
cork: cleanups and fix cloning tags

marineam:
This now works:
```
cork create --manifest-name release.xml --manifest-branch refs/tags/v962.0.0
```

Note: although git and cork will handle just the tag name like v962.0.0 the current repo will not and sticks a refs/heads prefix on anything that it feels like. cork could use `git ls-remote` to resolve names given to repo but that a difference in behavior between cork and repo itself may be confusing.


mischief:
lgtm

PR #247, opened 2016-02-20
packet.net support

### LINKED ISSUE #238: platform: packet.net support
https://github.com/packethost/packngo

mischief:
fixes #238

unfortunately packet.net has no way to upload images currently, so without some tricks we can't run the kola tests on them, but we can at least create instances with spawn.


mischief:
correction, we can run kola tests on packet machines, just not on a fresh image.

bgilbert:
Closing in favor of #629.

PR #248, opened 2016-02-20
platform: handle shebang cloudconfigs

mischief:


platform/conf.go
```diff
@@ -111,6 +116,13 @@ func (c *Conf) copyKeysCloudConfig(keys []*agent.Key) {
 	}
 }
 
+// mildly unreliable but i'm not sure what else to do.
```
marineam: Eww, how about we just wrap scripts in a cloud config instead?

PR #250, opened 2016-02-26
cork: improve enter and add update command

marineam:


mischief:
[lgtm](http://cdn-ak.f.st-hatena.com/images/fotolife/m/mizunokura/20151226/20151226150807.png)

PR #254, opened 2016-03-01
sdk: fix enter

marineam:
Commit f1c32301 was broken, I apparently renamed a variable during a
rebase and didn't retest it. Good grief.


marineam:
Ugg... removing that mention bot now...

PR #255, opened 2016-03-02
repo: start basic repo re-implementation in Go

marineam:
This piece is a bit verbose, creating data structures for the complete
manifest format (at least approximately the documented manifest format,
differences with the python implementation may exist...). This package's
task is just to test that repo is working properly so it will not
implement the vast majority of its features. Just enough to prase a
manifest and check that the right things are checked out.

Based on docs/manifest-format.txt in repo v1.12.32


mischief:
lgtm pending travis/jenkins

PR #256, opened 2016-03-04
cork: add support for asserting source trees are consistent

marineam:
Some independent sanity checking to raise my confidence that builds are working.


mischief:
lgtm

PR #258, opened 2016-03-09
kola/tests/kubernetes: bump nginx timeout

pbx0:
While we are here, fix kubectl flags to avoid deprecation warnings.


mischief:
lgtm

PR #259, opened 2016-03-09
platform: toss Wait error in Manhole

mischief:
we don't care about the error after the shell exits.


pbx0:
Often times, I’ll place a `return platform.Manhole(m)` halfway through function that gets called for a couple machines. Then if on the first machine I find I don’t want to continue the test, I’ll simple type `exit 1` instead of `exit` to tear down the test. This PR would ignore that error right?

PR #261, opened 2016-03-12
(WIP) kubernetes conformance tests

pbx0:
This will run the kubernetes conformance tests on a cluster setup by kola using test binaries pre-built in containers.

Conformance tests are basically a subset of the e2e tests that allow you to pass in an existing cluster. Provider specific tests are not run.

I'll consider this PR ready to submit once I deal with expected errors from the conformance test output. Right now i'm just trying to create a whitelist of errors for each version we test. Any errors outside the whitelist will cause the test to report the errors and fail.

### DIFF
--- a/kola/tests/kubernetes/basic.go
+++ b/kola/tests/kubernetes/basic.go
@@ -20,9 +20,57 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/mantle/kola/register"
 	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/util"
 )
+
+// register a separate test for each version tag
+var basicTags = []string{
+	"v1.1.7_coreos.2",
+	"v1.1.8_coreos.0",
+}
+
+func init() {
+	for i := range basicTags {
+		// use closure to store a version tag in a Test
+		t := basicTags[i]
+		f := func(c platform.TestCluster) error {
+			return CoreOSBasic(c, t)
+		}
+
+		register.Register(&register.Test{
+			Name:        "google.kubernetes.basic." + t,
+			Run:         f,
+			ClusterSize: 0,
+			Platforms:   []string{"gce", "aws"},
+		})
+	}
+}
+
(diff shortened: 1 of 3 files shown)


kola/tests/kubernetes/basic.go
```diff
@@ -20,10 +20,58 @@ import (
...
+}
+
+// Run basic smoke tests on cluster. Assumes master is machine index 1,
+// workers make up the rest.
+func CoreOSBasic(c platform.TestCluster, version string) error {
+	if err := setupCluster(c, 2, version); err != nil {
+		return err
+	}
+
+	master := c.Machines()[1]
```
mischief: first machine is 0 in the slice, so should be [0], [1:]

pbx0: This is kinda a leaky abstraction of the setup function. The first machine created is an etcd node. The second machine is master. The rest are worker nodes. 

I could have the setup function return `(err error, master platform.Machine, workers []platform.Machine)`  but that seemed ugly too.

We could easily run etcd on the master node, I just wanted to keep it separate when I first made the test for some reason.

pbx0: Oh dang, this isn't even safe! I was having tons of trouble testing some of this because I would manhole into c.Machines()[1] but when I ran it without manhole it was logging into the wrong machine.

We can't count on the order of the machine list because its actually stored in `basecluster.machmap`. I didn't realize or had forgot that detail. Ugh.

mischief: can you just make a new type in the test and return that from setupCluster?

e.g.:

```go
type k8sCluster struct {
  etcd Machine
  master Machine
  workers []Machine
}
```

and just store the result of NewMachine()/NewMachines() there. that way you're not dependant on ordering of Machines() in the caller.

pbx0: Thats a good idea. I had just resorted to passing back master and worker pointers, but I'll combine them into a struct.

PR #262, opened 2016-03-14
lang: create a new package for language utilities

marineam:
More specific package names than `util` help keep things organized. This
new package is basically for things Go could have been nice enough to do
for us but doesn't, giving us sads.

First obvious candidate for this is our destructor implementation.

I'll have more stuff for this soon. ;-)


pbx0:
lgtm

PR #264, opened 2016-03-15
platform: make cluster.Machines() less surprising

pbx0:
Change cluster.Machines() to be less surprising. Since it returns a
slice, kola test-writers may assume that the order of this is always the
same.  While changing the backing structure from a map to a slice still
doesn't make this strictly true because elements can be deleted, it will
make it true in the common case. Normally, machines are not torn down
until the end of a test.

I think this is worthwhile regardless if the better advice is to always
carry pointers to machines created in a test.

### DIFF
--- a/platform/base.go
+++ b/platform/base.go
@@ -33,7 +33,7 @@ type baseCluster struct {
 	agent *network.SSHAgent
 
 	machlock sync.Mutex
-	machmap  map[string]Machine
+	machs    []Machine
 }
 
 func newBaseCluster() (*baseCluster, error) {
@@ -49,8 +49,8 @@ func newBaseCluster() (*baseCluster, error) {
 	}
 
 	bc := &baseCluster{
-		agent:   agent,
-		machmap: make(map[string]Machine),
+		agent: agent,
+		machs: make([]Machine, 0),
 	}
 
 	return bc, nil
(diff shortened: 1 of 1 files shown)


pbx0:
@mischief @marineam

marineam:
What about leaving the map and just change the machine-to-list loop maps.SortedKeys that I just added? Or use the linked list implemenation collections. The tediousness of deleting items from a slice is kinda silly.

marineam:
implemented in containers that is: https://golang.org/pkg/container/list/

pbx0:
I thought about doing a sort, but, it will probably be more likely to confuse the common case where machines are created in serial sequence and someone tries to slice off the first machine specifically (usually some form of a master). Many tests have one machine and slice the first element. Its quite possible that future updates to that test to add more machines would break that. Deletes during a test are rare and could be surprising either way.

What if we just change the function to pass back a map instead of a slice? Then its always obvious you can't rely on order. I don't think it'd be much work to modify tests into complying here.

marineam:
Passing back a map sounds good then, just be sure to copy it so the
original map remains private.

euank:
Needs rebase @pbx0

marineam:
I'm going to close this as I don't think it is particularly important and would be better implemented by just changing Machines() to return a copy of the map or use https://godoc.org/github.com/coreos/mantle/lang/maps#SortedKeys or similar to stabilize the order.

If anyone needs this feel free to revive with one of those options.

PR #265, opened 2016-03-15
index: use a hash to trigger index updates instead of time

marineam:
This is more robust since it can detect deletion of child nodes and
changes to the html template. For now CRC32c is used but the code is
structured such that it would be trivial to transition if ever needed.

### DIFF
--- a/index/directory.go
+++ b/index/directory.go
@@ -15,13 +15,17 @@
 package index
 
 import (
+	"encoding/base64"
 	"fmt"
+	"hash"
+	"hash/crc32"
 	"net/http"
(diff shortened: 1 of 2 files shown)


marineam:
Upon further reflection I realized that this PR has the built-in assumption that computing an artificial CRC is meaningfully more efficient for the noop case than it would have been to just execute the template and compute the CRC (or MD5) of the actual contents and I don't know that this is actually true or not. So consider this PR on hold until I test that.

marineam:
Replaced by https://github.com/coreos/mantle/pull/271

PR #267, opened 2016-03-19
Godep and other fixes

marineam:


pbx0:
lgtm -- I have been using gb for mantle as an experiment but stuck with it because it was doing incremental builds.

marineam:
I actually never use the build and test scripts normally, just for the sake
of occasional builds of the whole tree. But the scripts may as well be less
bad :)

Issue #164, opened 2015-10-27
mantle: clean-up manual instance spawning


pbx0:
`ore` and `kola` now both share the ability to manually spawn a VM. `plume` shares code with ore and is out of sync with the copy-pasta. The `kola spawn` command probably belongs in `ore` and we should try and factor common code out among all three binaries and clean-up the user-interfaces.

PR #269, opened 2016-03-19
ore+plume: replace cloud/storage with api/storage package.

marineam:
The cloud/storage package is documented as experimental and has been significantly refactored since this code was written. Since this code needs to be reworked anyway may as well switch to the somewhat more stable api/storage package.

There is a far amount of code duplication in these commands but fixing that is left for another day.


marineam:
Note: build tested, I still need to take some time to actually upload stuff to GCE

marineam:
Ok, upload tests check out :-D

mischief:
lgtm

pbx0:
`There is a far amount of code duplication in these commands but fixing that is left for another day.`

Agreed! I've wanted to address https://github.com/coreos/mantle/issues/164 for some time

PR #270, opened 2016-03-19
Update Google API bindings, use new built-in retry support

marineam:
And finally, we get to the reason behind https://github.com/coreos/mantle/pull/269 and https://github.com/coreos/mantle/pull/268

PR #271, opened 2016-03-19
index: use a hash to trigger index updates instead of time

marineam:
This is more robust since it can detect deletion of child nodes and
changes to the html template. For now CRC32c is used, if that is
insufficient we can switch to MD5 in the future. As an added bonus,
the storage server validates the upload using the provided hash.

Depends on https://github.com/coreos/mantle/pull/270
Replaces https://github.com/coreos/mantle/pull/265

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,9 @@
 language: go
 sudo: false
-matrix:
-  include:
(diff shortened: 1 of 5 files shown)


mischief:
lgtm

PR #273, opened 2016-03-21
Add ARM64 board support in Kola.

lupan2005:


mischief:
please go fmt your code.

cmd/kola/options.go
```diff
@@ -50,3 +57,24 @@ func init() {
 	sv(&kola.AWSOptions.InstanceType, "aws-type", "t1.micro", "AWS instance type")
 	sv(&kola.AWSOptions.SecurityGroup, "aws-sg", "kola", "AWS security group name")
 }
+
+// Sync up the command line options if there is dependency
+func syncOptions() {
+	// check if target board is default or not
+	if kola.QEMUOptions.Board != defaultTargetBoard {
+		// Is the target board supported?
+		if kola.QEMUOptions.Board != "arm64-usr" {
```
mischief: instead of checking for arm64-usr, just try to index `kolaDefaultImages`.

also, i dont think you should override user's settings here - just error out of the board isn't in `kolaDefaultImages`.

mischief:
what was wrong with the disk fd in qemu-system-aarch64?

lupan2005:
When use the disk fd with qemu-system-aarch64, CoreOS was not able to boot up successfully due to an error when mounting the disk image.

cmd/kola/options.go
lupan2005: Updated the patches to address the comments.

cmd/kola/options.go
```diff
@@ -50,3 +57,20 @@ func init() {
 	sv(&kola.AWSOptions.InstanceType, "aws-type", "t1.micro", "AWS instance type")
 	sv(&kola.AWSOptions.SecurityGroup, "aws-sg", "kola", "AWS security group name")
 }
+
+// Sync up the command line options if there is dependency
+func syncOptions() error {
+	for k, v := range kolaDefaultImages {
+		if kola.QEMUOptions.Board == k {
+			// check if DiskImage has been changed
+			if kola.QEMUOptions.DiskImage == kolaDefaultImages[defaultTargetBoard] {
```
marineam: I'd rather make the default for the option blank so it isn't ambiguous whether the user specified the path or it really is the default value.

cmd/kola/options.go
```diff
@@ -15,12 +15,18 @@
...
 import (
+	"fmt"
 	"github.com/coreos/mantle/kola"
 	"github.com/coreos/mantle/sdk"
 )
 
 var (
-	kolaPlatform string
+	kolaPlatform       string
+	defaultTargetBoard = "amd64-usr"
```
marineam: Use `sdk.DefaultBoard()` so kola's --board option respects the default just like the build scripts do.

platform/qemu.go
```diff
@@ -183,13 +201,12 @@ func (qc *QEMUCluster) GetDiscoveryURL(size int) (string, error) {
 
 // Copy the base image to a new nameless temporary file.
 // cp is used since it supports sparse and reflink.
-func setupDisk(imageFile string) (*os.File, error) {
+func setupDisk(imageFile string) (string, error) {
 	dstFile, err := ioutil.TempFile("", "mantle-qemu")
 	if err != nil {
-		return nil, err
+		return "Error: fail to create the destination file", err
```
marineam: Do not stick error messages into file path, either `return "", err` or `return "", fmt.Errorf("something exitra: %v", err)`

platform/qemu.go
```diff
@@ -199,10 +216,10 @@ func setupDisk(imageFile string) (*os.File, error) {
 	cp.Stderr = os.Stderr
 
 	if err := cp.Run(); err != nil {
-		return nil, err
+		return "Error: fail to copy the image file", err
 	}
 
-	return os.OpenFile(dstFileName, os.O_RDWR, 0)
+	return dstFileName, err
```
marineam: err isn't meaningful at this point and is going to be a left over nil from ioutil.TempFile. Just do `return dstFileName, nil`

cmd/kola/options.go
```diff
@@ -50,3 +57,20 @@ func init() {
 	sv(&kola.AWSOptions.InstanceType, "aws-type", "t1.micro", "AWS instance type")
 	sv(&kola.AWSOptions.SecurityGroup, "aws-sg", "kola", "AWS security group name")
 }
+
+// Sync up the command line options if there is dependency
+func syncOptions() error {
+	for k, v := range kolaDefaultImages {
```
marineam: missed this before, no need to loop here since this is a map:

```go
if kola.QEMUOptions.DiskImage == "" {
    image, ok := kolaDefaultImages[kola.QEMUOptions.Board]
    if !ok {
        fmt.Errorf("unsupport board "%q", kola.QEMUOptions.Board)
    }
    kola.QEMUOptions.DiskImage = image
}

marineam:
LGTM, thanks!

PR #274, opened 2016-03-24
index: use the new Pages helper instead of iterating ourselves

marineam:
Removes a little bit of the tedious detail from our code.


mischief:
lgtm

PR #276, opened 2016-03-25
Merge coreos build 998

bassamtabbara:


bassamtabbara:
so sorry, but this is the wrong repo.

PR #277, opened 2016-03-28
Kola: add test for empty config on AWS

crawford:


crawford:
This will require a long-lived key to be used when running the AWS tests. Otherwise, the SSH key won't be on the instance and the test will fail.

crawford:
Will catch issues like https://github.com/coreos/bugs/issues/1184 in the future.

mischief:
lgtm pending ci passing

PR #278, opened 2016-03-28
Kola: add EndVersion to tests

crawford:


kola/harness.go
```diff
@@ -109,8 +109,8 @@ func filterTests(tests map[string]*register.Test, pattern, platform string, vers
 			continue
 		}
 
-		// Check the test's minVersion when running more then one test
-		if version.LessThan(t.MinVersion) && t.Name != pattern {
+		// Check the test's min and max versions when running more then one test
+		if t.Name != pattern && t.MinVersion.LessThan(t.MaxVersion) && (!t.MinVersion.LessThan(version) || !version.LessThan(t.MaxVersion)) {
```
pbx0: This comparison `t.MinVersion.LessThan(t.MaxVersion)` could be checked earlier in the `Register` function and cause a panic.

pbx0:
Otherwise, LGTM

kola/harness.go
crawford: Where is that?

pbx0: Sorry, its unintuitive: `/kola/register/register.go`

That package exists to avoid a cyclic dependency

crawford: Ohhhhh, I see what you are saying. The reason I have it here is so that a Min and Max version of 0.0.0 (the default) still get executed. I guess I could trigger the panic in Register, and then check against 0.0.0 in this section.

crawford:
PTAL. Changed to be an exclusive upper bound (should be easier to write tests) and added a panic.

mischief:
lgtm

PR #279, opened 2016-03-30
Kola: update Ignition and version package

crawford:


crawford:
Actually, I want to rethink this slightly. Ignition will be able to parse V and V-1 configs. We might not need to version the config package.

PR #280, opened 2016-04-04
Kola: add tests for Ignition v2 configs

crawford:


kola/tests/ignition/v2/empty.go
```diff
@@ -0,0 +1,33 @@
...
+
+package ignition
+
+import (
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform"
+)
+
+func init() {
+	register.Register(&register.Test{
```
crawford: These will need a MinVersion, but I'm not sure what that version is going to be just yet.

marineam:
Not particularly excited about having go-yaml nested a layer deep like that, particularly since the rest is still using path rewriting instead of the vendor/ dir magic. But it is a small thing that doesn't impact much else... ehhhh

marineam:
LGTM though, I'll just go cry about Go dep management on my own like usual.

crawford:
Haha, yeah.........................................................................

PR #281, opened 2016-04-06
Kola: fix inverted version logic

crawford:


mischief:
lgtm pending build passing and tests running :P

mischief:
looks like tests are executing now, but the harness can't ssh.

PR #283, opened 2016-04-07
*: add cloud-config stub to NewMachine calls

crawford:
This is needed if the SSH keys need to be added to the config.


pbx0:
lgtm

PR #284, opened 2016-04-11
New package for managing cloud storage

marineam:
This is the core chunk of code for replacing "gsutil rsync && plume index" with a single command that will keeps a local cache of bucket state, avoiding issues caused by the eventual consistency of bucket object listings.

Integrating all this into a new plume command will come in a follow-up commit.

### DIFF
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -97,6 +97,19 @@ func (b *Bucket) WriteDryRun(dryrun bool) {
 	b.writeDryRun = dryrun
 }
 
+// TrimPrefix removes the Bucket's path prefix from an Object name.
+func (b *Bucket) TrimPrefix(objName string) string {
+	if !strings.HasPrefix(objName, b.prefix) {
+		panic(fmt.Errorf("%q missing prefix %q", objName, b.prefix))
+	}
+	return objName[len(b.prefix):]
+}
+
+// AddPrefix joins the Bucket's path prefix with a relative Object name.
+func (b *Bucket) AddPrefix(objName string) string {
+	return b.prefix + objName
+}
+
 func (b *Bucket) Object(objName string) *storage.Object {
 	b.mu.RLock()
 	defer b.mu.RUnlock()
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -1,0 +1,336 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storage
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+
+	"github.com/coreos/mantle/Godeps/_workspace/src/golang.org/x/net/context"
+	"github.com/coreos/mantle/Godeps/_workspace/src/google.golang.org/api/googleapi"
+	"github.com/coreos/mantle/Godeps/_workspace/src/google.golang.org/api/storage/v1"
+)
+
+var (
+	UnknownScheme = errors.New("storage: URL missing gs:// scheme")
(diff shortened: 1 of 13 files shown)


storage/bucket.go
```diff
@@ -0,0 +1,347 @@
...
+	b.writeDryRun = dryrun
+}
+
+func (b *Bucket) Object(objName string) *storage.Object {
+	b.mu.RLock()
+	defer b.mu.RUnlock()
+	return b.objects[objName]
+}
+
+// Check of an object name's directory is the specified prefix.
```
crawford: Check if

storage/bucket.go
```diff
@@ -0,0 +1,347 @@
...
+func matchPrefix(path, prefix string) bool {
```
crawford: https://golang.org/pkg/strings/#HasPrefix?

storage/bucket.go
```diff
@@ -0,0 +1,347 @@
...
+	for objName, obj := range b.objects {
+		if matchPrefix(objName, prefix) {
+			objs[objName] = obj
+		}
+	}
+	return objs
+}
+
+func (b *Bucket) Prefixes() []string {
+	prefixmap := make(map[string]bool)
```
crawford: NIT: `map[string]struct{}` is a bit more idiomatic

marineam: No, bool is actually what I want because it represents a boolean value ;-)

storage/bucket.go
```diff
@@ -0,0 +1,347 @@
...
+		if v.Bucket != "" {
+			u.Host = v.Bucket
+		}
+		return u
+	case *url.URL:
+		return v
+	case nil:
+		return b.URL()
+	default:
+		panic("unknown type")
```
crawford: Maybe:

```go
panic(fmt.Sprintf("unknown type %T", obj))
```

storage/bucket.go
```diff
@@ -0,0 +1,347 @@
...
+	defer stopRequest()
+	plog.Noticef("Fetching %s", b.URL())
+
+	return b.apiErr("storage.objects.list", nil, req.Pages(nil, add))
+}
+
+func (b *Bucket) Upload(ctx context.Context, obj *storage.Object, media io.ReadSeeker) error {
+	// Calculate the checksum to enable upload integrity checking.
+	if obj.Crc32c == "" {
+		obj = dupObj(obj) // avoid editing the original
```
crawford: Should this be shadowed instead?

storage/bucket.go
marineam: No, that would match a name that is nested at any level under a directory, this is matching names that are directly under the given prefix, no further subdirs.

That said, I'm going to delete this since I never wound up using the prefix arg to Objects()

storage/bucket.go
crawford: No, you are only representing the presence of a path. The bool type within a map is effectively a nullable/Option, which is three states.

storage/bucket.go
marineam: What do you mean?

storage/bucket.go
marineam: Fixed in 484f488

storage/bucket.go
```diff
@@ -112,13 +125,13 @@ func matchPrefix(path, prefix string) bool {
 	return path[:i+1] == prefix
 }
 
-func (b *Bucket) Objects(prefix string) map[string]*storage.Object {
-	objs := make(map[string]*storage.Object)
+func (b *Bucket) Objects(prefix string) []*storage.Object {
```
crawford: These objects are going to pop out in a random order now. Is that a problem?

storage/bucket.go
marineam: Gone in 61e351f

storage/bucket.go
marineam: Gone in 61e351f

storage/bucket.go
marineam: Ok, sure, I can use struct{} and _, ok = but I think that looks worse and cluttered. I suppose worse and cluttered is idiomatic Go so I can do that :-P

storage/bucket.go
crawford: Well, as it is, the rest of the code beyond this block operates on the modified copy of obj instead of the original.

storage/bucket.go
crawford: Haha, I love the enthusiasm.

marineam: Done in 230f8b9

storage/bucket.go
marineam: exactly, that's what I want. The hash should be set in order for the server to reject corrupted uploads.

crawford: Oh, I see. Nevermind.

crawford:
LGTM

PR #286, opened 2016-04-16
Finish up the new sync, drop old index code.

marineam:
The new index package is now at parity with the old and sync does everything it should.


mischief:
lgtm

PR #287, opened 2016-04-17
Implement Martin Pool's Natural Order sorting

marineam:
This is the same sort algorithm as used by Apache httpd's index module with `VersionSort` enabled. Considering that is exactly what all my indexing go is mimicking in the first place that seems fitting. Original version available here: https://github.com/sourcefrog/natsort

Interesting trivia for related software, GNU's `sort --version-sort` implements the full [Debian package version format](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version) which I find quite silly complicated and specific for such a generic thing. ;-)

### DIFF
--- a/lang/maps/sorted.go
+++ b/lang/maps/sorted.go
@@ -17,9 +17,12 @@ package maps
 import (
 	"reflect"
 	"sort"
+
+	"github.com/coreos/mantle/lang/natsort"
 )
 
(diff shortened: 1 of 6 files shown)


marineam:
Depends on part of https://github.com/coreos/mantle/pull/286 to fix a test.

mischief:
lgtm

PR #288, opened 2016-04-17
storage/index: natural sort links for better version ordering

marineam:
Depends on https://github.com/coreos/mantle/pull/286 and https://github.com/coreos/mantle/pull/287


mischief:
lgtm

PR #289, opened 2016-04-17
test: end the terrible doom of listing packages by hand

marineam:
This test script has a lot of suck which is why I didn't tend to use it, but it is nice to have coverage, fmt, and vet all wrapped up into one. Take advantage of `go list` to make `gofmt` work like other commands after which a simple `./...` is all we ever need.
    
Adds a few other tidbits: go install everything, `go test -i` only installs dependencies, further speeding up incremental builds. Adds a variant of our standard cover script but this one actually accepts target package as an argument like a sane thing should.

Depends on part of https://github.com/coreos/mantle/pull/286 to fix a broken test.

### DIFF
--- a/cover
+++ b/cover
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+#
+# Generate coverage HTML for a package
+#   PKG=./foo ./cover
+#   ./cover ./foo
+#
+set -e
+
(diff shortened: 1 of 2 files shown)


mischief:
lgtm

PR #291, opened 2016-04-20
use proper path to arm64 EFI firmware for qemu

mischief:


mischief:
retest this please

mischief:
maybe this should use `/build/$BOARD/usr/share/edk2-armvirt/QEMU_EFI.fd`, otherwise something needs to copy it to the image dir during build_image.

marineam:
Yeah, reading from /build/... sounds fine.

marineam:
If it doesn't exist already probably should add a sdk.BoardRoot() function

sdk/repo.go
```diff
@@ -95,6 +95,10 @@ func DefaultBoard() string {
 	return string(board)
 }
 
+func BoardRoot() string {
```
marineam: Support passing in a string and use DefaultBoard if string is blank.

cmd/kola/options.go
```diff
@@ -27,6 +29,11 @@ var (
 		"amd64-usr": sdk.BuildRoot() + "/images/amd64-usr/latest/coreos_production_image.bin",
 		"arm64-usr": sdk.BuildRoot() + "/images/arm64-usr/latest/coreos_developer_image.bin",
 	}
+
+	kolaDefaultBIOS = map[string]string{
+		"amd64-usr": "",
```
marineam: Use `bios.bin` and then this flag can be used for both. And then it is possible to do kola on x86_64 uefi too :)

mischief:
ptal

platform/qemu.go
```diff
@@ -33,6 +33,10 @@ type QEMUOptions struct {
 	// DiskImage is the full path to the disk image to boot in QEMU.
 	DiskImage string
 	Board     string
+
+	// BIOSImage is the full path to the BIOS image to use in QEMU.
+	// Currently only used for UEFI in aarch64.
```
marineam: comment no longer correct

marineam:
a comment needs fixing but otherwise LGTM

PR #292, opened 2016-04-21
New plume release command, supporting storage and GCE

marineam:
The first commit here is the pretty one, it's the happy layer on top of all the storage infrastructure code I've been working on the past couple weeks. The second one for GCE, not so much. The existing code doesn't quite do all I need it to but today is not the day to invest in it. Cleanups will come later.

This replaces `core_promote` except for updating the roller channel which we have tended to do separately anyway.

### DIFF
--- a/cmd/plume/plume.go
+++ b/cmd/plume/plume.go
@@ -15,14 +15,18 @@
 package main
 
 import (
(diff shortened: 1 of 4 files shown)


mischief:
lgtm

PR #297, opened 2016-04-26
Fix non-recursive index generation

marineam:
This bug caused http://alpha.release.core-os.net/amd64-usr/ to appear empty.


crawford:
LGTM

PR #298, opened 2016-04-26
Replace plume index

marineam:
The old `plume index` is now `ore index` and the new command uses the release spec to figure out what storage locations to update rather than needing an explicit list.


mischief:
lgtm

PR #303, opened 2016-04-29
Godeps: re-run save, remove nested vendoring under ignition

marineam:
godep does not support mixing the old and new vendor schemes. This means
it is not aware of packages under github.com/coreos/ignition/config/vendor
when using Godeps/_workspace in mantle. On the flip side if mantle is
migrated to using vendor itself godep no longer gracefully handles
github.com/coreos/etcd/Godeps, leaving us with a big pile of duplicated
packages. This could complicate a number of things, off the top of my
head capnslog won't work right any more if that happens.

etcd has reworked its own vendoring in master so this shouldn't be a
problem after 2.4 or whatever the next major release is. Until then it
is easy to just let godep vendor ignition and its dependencies exactly
how it wants it so lets just do that. Sorry Crawford. :-P

go-semver for mantle has been updated to the newer rev used by ignition.


mischief:
lgtm

mischief:
lgtm

mischief:
woops, should have refreshed github.

PR #304, opened 2016-04-29
plume: define more GCE metedata in release specs

marineam:
Name and Description no longer hard-code "coreos" which feels like the
right thing to do but aren't fully templatized because that seems
overkill. Add Family and Licenses attributes, new types of metadata
useful for finding images and tracking usage. Family is not yet passed
to the API because the Google hasn't refreshed their Go API bindings to
include it yet. Will add that once available.


mischief:
lgtm

PR #305, opened 2016-04-29
cmd/kola: always set a userdata in spawn

mischief:
empty configs are now truely empty, and have no ssh keys injected.

if a config isn't provided, make up a cloud-config so an ssh key gets
injected so that ssh works.


crawford:
LGTM

PR #307, opened 2016-04-30
kola/tests/docker: add docker.resources

mischief:
tests a few docker run arguments which influence cgroup settings.

can't test blkio because blkio settings don't work in virtual
environments due to lack of cfq scheduler on virtual controllers.


marineam:
LGTM

PR #310, opened 2016-05-05
kola: emit tap results

mischief:
initial go at tap.

we should really not filter tests and simply emit SKIP's for them so that the number of tests are consistent...

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -16,6 +16,7 @@ package kola
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


mischief:
```
$ cat aws.tap 
TAP version 13
1..1
not ok 1 - docker.resources # [0] failed to run "docker run --rm --memory-swappiness=50 sleep sleep 0.2": output: "" status: "Process exited with: 2. Reason was:  ()" [1] failed to run "docker run --rm --memory-swappiness=50 sleep sleep 0.2": output: "" status: "Process exited with: 2. Reason was:  ()"
```

kola/harness.go
```diff
@@ -78,6 +80,37 @@ type result struct {
...
+	}
+
+	var werr error
+
+	for i, r := range res {
+		t := r.test
+		err := r.result
+		if err != nil && err == register.Skip {
+			// TODO: print why it was skipped
+			_, werr = fmt.Fprintf(w, "not ok %d - %s # SKIP\n", i+1, t.Name)
```
marineam: shoudn't skip be 'ok'?

mischief:
i misread `DIRECTIVES`.. fixed.

marineam:
LGTM

PR #311, opened 2016-05-06
plume: remove gce commands

marineam:
Very similar functionality exists in `ore`, the main difference is this
one had a little extra version detection goo. If anyone wants that we
can add it to ore. The functionality needed for releases is different.


pbx0:
LGTM

PR #315, opened 2016-05-12
kola: support streaming the journal in updatepayload

marineam:


mischief:
the `journalctl -f` will stop once the machine is rebooted, dunno if that's desirable.

marineam:
yeah, can add another for after boot but the before bit is the important part.

mischief:
lgtm

PR #316, opened 2016-05-12
kola: disable empty ignition config test

marineam:


mischief:
lgtm. we need to implement private key loading from a local file or use of a local agent before this can work.

PR #317, opened 2016-05-12
Revert "kola: disable empty ignition config test"

marineam:
Reverts coreos/mantle#316

Returning Skip doesn't do anything, the test fails before this function even runs. Nevermind...


crawford:
LGTM

PR #319, opened 2016-05-13
ore: add new create-image command

marineam:
I honestly thought this was already here, but I mixed up image creation
with instance creation. Options differ slightly from upload, assembling
the image URL from the board and version instead of taking the full path
and deriving that information from it. In the future I'm going to clean
this up and make usage more consistent across plume, ore, and our build
scripts.


mischief:
lgtm

PR #323, opened 2016-05-16
Add authentication test

mjg59:
Ensure that ssh login attempts with invalid credentials are rejected


marineam:
Looks good, one nit though, perhaps "Credentialed" should actually be more specifically "Password" since I could also see utility in hitting PAM directly via challenge/response/interactive mode and so on later on.

marineam:
Travis claims you need to gofmt as well

crawford:
+1 for Password instead of Credentialed.

crawford:
Update commit message, otherwise LGTM.

marineam:
LGTM

PR #324, opened 2016-05-17
sdk: update CoreOS image signing key

marineam:
Revokes the `B58844F1` subkey used to sign 1045 and 1047, adds a new one for future releases.


mischief:
lgtm

PR #325, opened 2016-05-17
Add some additional tests

mjg59:
Verify that we don't have any unexpectedly world-writable dirs or
files, no unexpected SUID or SGID files and no unauthorised
network listeners.

### DIFF
--- a/kola/tests/misc/files.go
+++ b/kola/tests/misc/files.go
@@ -1,0 +1,151 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package misc
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         SUIDFiles,
+		ClusterSize: 1,
+		Name:        "coreos.filesystem.suid",
+		UserData:    `#cloud-config`,
+	})
+	register.Register(&register.Test{
+		Run:         SGIDFiles,
+		ClusterSize: 1,
+		Name:        "coreos.filesystem.sgid",
+		UserData:    `#cloud-config`,
+	})
+	register.Register(&register.Test{
+		Run:         WritableFiles,
+		ClusterSize: 1,
+		Name:        "coreos.filesystem.writablefiles",
+		UserData:    `#cloud-config`,
+	})
+	register.Register(&register.Test{
+		Run:         WritableDirs,
+		ClusterSize: 1,
+		Name:        "coreos.filesystem.writabledirs",
+		UserData:    `#cloud-config`,
+	})
+}
+
+func sugidFiles(m platform.Machine, validfiles []string, mode string) error {
+	badfiles := make([]string, 0, 0)
+
+	command := fmt.Sprintf("sudo find / -path /sys -prune -o -path /proc -prune -o -type f -perm -%v -print", mode)
+
+	output, err := m.SSH(command)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


kola/tests/misc/users.go
```diff
@@ -0,0 +1,70 @@
...
+
+type user struct {
+	username string
+	shell    string
+}
+
+func CheckUserShells(c platform.TestCluster) error {
+	m := c.Machines()[0]
+
+	ValidUsers := []user{
```
mischief: can this be a map[string]string of user->shell?

mjg59: Yes

kola/tests/misc/users.go
```diff
@@ -0,0 +1,70 @@
...
+		{"root", "/bin/bash"},
+		{"sync", "/bin/sync"},
+		{"shutdown", "/sbin/shutdown"},
+		{"halt", "/sbin/halt"},
+		{"core", "/bin/bash"},
+	}
+
+	output, err := m.SSH("grep -v nologin /etc/passwd /usr/share/baselayout/passwd")
```
marineam: what about `getent passwd`?

kola/tests/misc/users.go
marineam: would avoid the loop, and if not found assert it is equal to nologin to be a bit stricter than the grep

marineam: hm, though this isn't going to work on GCE which adds lots of users. *but* after crawford's update to the GCE oem it will start working if we create instances with the metadata flag to ignore project users/ssh keys.

kola/tests/misc/files.go
```diff
@@ -0,0 +1,152 @@
...
+		return fmt.Errorf("Unknown writable files found: %v", output)
+	}
+
+	return nil
+}
+
+func WritableDirs(c platform.TestCluster) error {
+	m := c.Machines()[0]
+
+	output, err := m.SSH("sudo find / -regextype posix-extended -regex \"/(sys|proc)\" -prune -o -type d -perm -0002 -a ! -perm -1000 -print")
```
marineam: How many sticky-bit dirs are there? can they be whitelisted instead of universally ignoring anything with the sticky bit?

kola/tests/misc/files.go
```diff
@@ -0,0 +1,152 @@
...
+		return fmt.Errorf("Unknown SUID or SGID files found: %v", badfiles)
+	}
+
+	return nil
+}
+
+func SUIDFiles(c platform.TestCluster) error {
+	m := c.Machines()[0]
+
+	validfiles := []string{
```
marineam: nti: For the sake of keeping future diffs clean lets start off with this list being sorted.

kola/tests/misc/users.go
mjg59: Ok

kola/tests/misc/files.go
mjg59: There's a bunch under /tmp - we could whitelist them but my presumption was that anything that's gone to the trouble of setting the sticky bit is probably doing this for good reasons

kola/tests/misc/users.go
```diff
@@ -0,0 +1,59 @@
...
+	output, err := m.SSH("getent passwd | grep -v nologin")
+	if err != nil {
+		return fmt.Errorf("Failed to run grep: output %v, status: %v", output, err)
+	}
+
+	users := strings.Split(string(output), "\n")
+
+	for _, user := range users {
+		userdata := strings.Split(user, ":")
+		if userdata[7] != ValidUsers[userdata[1]] {
```
marineam: check the len first so bad data is a nice error message instead of panic.

kola/tests/misc/files.go
mjg59: Sounds good

kola/tests/misc/users.go
marineam: I'd also rather something like:

```
shell := ValidUsers[userdata[1]]
if shell == "" {
    shell = "/sbin/nologin"
}
```

than the grep above

mjg59: Done

kola/tests/misc/users.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         CheckUserShells,
+		ClusterSize: 1,
+		Name:        "coreos.users.shells",
+		UserData:    `#cloud-config`,
```
marineam: Restrict this to qemu and aws platforms for now

marineam:
Just need to disable the users check for GCE for now, after that looks good!

marineam:
needs gofmt too

PR #326, opened 2016-05-17
recover skips/errors to align with testing.TB

mischief:


util/skip.go
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+// Skip is a sentinel value that can be returned by tests that are skipped
+// rather than passing or failing.
+type Skip string
```
marineam: Is this in util just to avoid a loop with platform and kola harness packages? What about a kola/skip pkg?

mischief: yea, basically:

```
$ ./build kola
Building kola
import cycle not allowed
package github.com/coreos/mantle/cmd/kola
        imports github.com/coreos/mantle/kola
        imports github.com/coreos/mantle/kola/register
        imports github.com/coreos/mantle/platform
        imports github.com/coreos/mantle/kola/register
```

mischief: though, it is unfortunate that recovery is done in harness and panicking is done in platform...

platform/platform.go
```diff
@@ -136,6 +136,32 @@ func (t *TestCluster) DropFile(localPath string) error {
...
+}
+func (t *TestCluster) skip(why string) {
+	panic(util.Skip(why))
+}
+
+func (t *TestCluster) Skip(args ...interface{}) {
+	t.skip(fmt.Sprint(args...))
+}
+
+func (t *TestCluster) SkipNow() {
```
marineam: testing may have this one but lets not bother, requiring skips to have a message.

kola/harness.go
```diff
@@ -418,16 +418,30 @@ func RunTest(t *register.Test, pltfrm string) error {
...
-	}
+	defer func() {
+		if err != nil {
+			time.Sleep(2 * time.Second)
+		}
+	}()
+
+	defer func() {
+		if r := recover(); r != nil {
+			switch r := r.(type) {
```
marineam: `r := r.(type)` is valid? But should just be `switch r.(type) {` I assume

mischief: `The TypeSwitchGuard may include a short variable declaration. When that form is used, the variable is declared at the beginning of the implicit block in each clause. In clauses with a case listing exactly one type, the variable has that type; otherwise, the variable has the type of the expression in the TypeSwitchGuard.` - spec

`If the switch declares a variable in the expression, the variable will have the corresponding type in each clause. It's also idiomatic to reuse the name in such cases, in effect declaring a new variable with the same name but a different type in each case.` - effective go

without the short variable declaration, pretty sure r is still interface{}.

marineam: ohhhh, ok.

kola/harness.go
```diff
@@ -418,16 +418,30 @@ func RunTest(t *register.Test, pltfrm string) error {
...
+			// can't have multiple types in a single case.
+			case error:
+				err = r
+			case skip.Skip:
```
marineam: `Skip` implements `error`, should the more specific `Skip` be first?

mischief: interesting point, but according to my test, the concrete type is chosen over the interface, although i'm unsure if this is documented somewhere...

```
$ sudo ./bin/kola --qemu-image ~/src/coreos/src/build/images/amd64-usr/latest/coreos_production_image.bin run coreos.verity.corruption   
2016-05-17T23:44:35Z kola: === RUN coreos.verity.corruption on qemu
2016-05-17T23:44:47Z kola: --- SKIP: coreos.verity.corruption on qemu (11.985s)
2016-05-17T23:44:47Z kola:         verity is not enabled
2016-05-17T23:44:47Z kola: 0 passed 0 failed 1 skipped out of 1 total
```

marineam: Yeah, the spec doesn't appear to specify if type switches are evaluated in-order like expression switches, and their example doesn't try to match interface types either. Though it seems fair to indicate our intended behavior by keeping the order of evaluation we want in the code.

Also you can reduce the nesting by putting nil as the first case of the switch instead of inside an if

marineam: So this whole thing should be:

```go
switch r := recover().(type) {
case nil:
case error:
    err = r
default:
    err = fmt.Errorf("test panicked: %v", r)
}
```

mischief:
ptal

kola/harness.go
```diff
@@ -95,16 +95,14 @@ func (res resultSlice) ToTAP(w io.Writer) error {
...
-		if err != nil {
-			switch err.(type) {
-			case skip.Skip:
-				_, werr = fmt.Fprintf(w, "ok %d - %s # SKIP: %s\n", i+1, t.Name, err)
-			default:
-				_, werr = fmt.Fprintf(w, "not ok %d - %s # %s\n", i+1, t.Name, err)
-			}
-		} else {
+
+		switch r.result.(type) {
```
marineam: `switch err := r.result.(type)`

kola/harness.go
```diff
@@ -95,16 +95,14 @@ func (res resultSlice) ToTAP(w io.Writer) error {
...
+		case nil:
 			_, werr = fmt.Fprintf(w, "ok %d - %s\n", i+1, t.Name)
+		case skip.Skip:
+			_, werr = fmt.Fprintf(w, "ok %d - %s # SKIP: %s\n", i+1, t.Name, err)
+		case error:
```
marineam: the type should be `error` to begin with so keeping this case as `default:` seems better, less likely for refactoring in the future to accidentally leave us with no log message at all.

kola/harness.go
```diff
@@ -261,20 +259,18 @@ func RunTests(pattern, pltfrm string) error {
...
-		} else {
+		switch err.(type) {
+		case nil:
 			plog.Noticef("--- PASS: %s on %s (%.3fs)", t.Name, pltfrm, seconds)
 			passed++
+		case skip.Skip:
+			plog.Errorf("--- SKIP: %s on %s (%.3fs)", t.Name, pltfrm, seconds)
+			plog.Errorf("        %v", err)
+			skipped++
+		case error:
```
marineam: same as above, lets keep `default`

kola/harness.go
```diff
@@ -418,16 +413,26 @@ func RunTest(t *register.Test, pltfrm string) error {
...
+		}
+	}()
 
-	return err
+	defer func() {
+		r := recover()
+		switch r := r.(type) {
+		case error:
+			err = r
+		default:
```
marineam: Needs a nil case

kola/harness.go
```diff
@@ -418,16 +413,26 @@ func RunTest(t *register.Test, pltfrm string) error {
...
-	// run test
-	err = t.Run(tcluster)
-
 	// give some time for the remote journal to be flushed so it can be read
 	// before we run the deferred machine destruction
-	if err != nil {
-		time.Sleep(10 * time.Second)
-	}
+	defer func() {
+		if err != nil {
```
marineam: Lets move this to the bottom of the switch below, seems silly to defer two functions when a single one will do.

marineam:
LGTM

PR #330, opened 2016-05-28
Implement processing update_engine payloads

marineam:
To facilitate testing of update_engine and experiment with changes to the payload format I want an independent implementation. Already found a couple of quirks, one of which is clearly a bug in update_engine.


mischief:
lgtm

PR #331, opened 2016-05-28
AnonymousFile: like a temporary file but even more so

marineam:
I've had this kicking around for a while but was kinda annoying because I had to roll my own linkat syscall wrapper. Well happy day, someone capitalized the wrapper in sys/unix so we can use it now. :)


crawford:
Neat. LGTM.

PR #335, opened 2016-06-22
Implement generating full updates for update_engine

marineam:
Another major piece in making sense of update_engine and testing it, a nearly complete implementation of generating full updates.


mischief:
what's going on in travis..?

marineam:
bah, forgot about travis, will need to insert some skips. It is too old to support O_TMPFILE :(

mischief:
lgtm

PR #337, opened 2016-07-01
flannel: test if docker is using flannel

ajeddeloh:


marineam:
LGTM but I'm not super familiar with this test, any opinions @mischief ?

ajeddeloh:
This will fail a good portion or the time until https://github.com/coreos/bugs/issues/1393 is fixed.

mischief:
@marineam this change is actually provoking a race where docker will run without attaching to the flannel bridge. that should get fixed first. unless we'd like to merge this to increase the pressure to fix it.

euank:
@ajeddeloh Do you want to rebase and get this in?

I'd like the commit to include a message about this change as well.

euank:
I'm closing this since it seems dead.

if it's not, please rebase and create a new PR @ajeddeloh.

PR #338, opened 2016-07-11
kola/tests/rkt: invoke etcd3 in rkt wrapper

mischief:


marineam:
Looks ok, we should do the full cluster discovery test too, though I don't recall if that runs on gce/aws. If not it'll need to be paramaterized to use the public etcd discovery service in addition to the built in etcd used by qemu.

So LGTM but please file a bug so we remember to add the cluster test too.

mischief:
@marineam that will be a bit more work. one must write systemd drop-ins for env vars or overriding ExecStart. the purpose of this test was more 'does something run in rkt' (since we have no actual rkt tests), not 'does etcd3 do stuff', since we already have etcd tests for the host binaries.

marineam:
@mischief are we shipping etcd3 as a host binary?

mischief:
@marineam ah, no. etcd binary on host hasn't been updated in quite some time.....

PR #339, opened 2016-07-11
platform: add Platform method

mischief:


platform/aws.go
```diff
@@ -32,6 +32,12 @@ type awsMachine struct {
 	mach    *ec2.Instance
 }
 
+var _ Machine = (*awsMachine)(nil)
```
marineam: What is the point of this?

mischief: it asserts awsMachine implements Machine at build time.

marineam: Ohh, is this an assertion that it provides the interface? But doesn't NewMachine provide that assertion?

mischief: i suppose it does, this is simply more explicit.

marineam: Considering Go tends to not use explicit assertions I think the implicit one in NewMachine is perfectly sufficient.

marineam:
Update the commit message since the assertion is gone, but LGTM

PR #341, opened 2016-07-12
trash godeps for glide+vendor, ignore bin/, bump go versions

mischief:


marineam:
Hm, could you pull the .gitignore and .travis.yml changes out to a different pr?

Also, in order to make things more incremental how about we switch to vendor using godep and then switch to glide. It would be nice to make the transition without actually modifying the vendored versions, updating them should be independent.

I'm going to do some maintaince on mantle myself so I can take a stab at that if you want a break from godep.

mischief:
i did try that. i had some serious trouble getting godep to cooperate. even the procedure documented in the godep repo didnt work.

marineam:
ok, I'll give it a go and see if I have better luck.

PR #343, opened 2016-07-13
Update cloudinit and other cleanups

marineam:
Some more maintenance before the storm


mischief:
I feel like the trend is local before external imports, but ok. Lgtm.

marineam:
I could have sworn that order was mentioned in the basic Go style guide but in usual minimalist form it seems to not be there. I guess I was just remembering it from other styles:

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
https://www.python.org/dev/peps/pep-0008/#imports

If it turns out Go style is trending differently we can flip later but sticking with this for now for the sake of not mixing both styles.

PR #347, opened 2016-07-20
platform refactoring

mischief:


marineam:
If possible I think I'd like to leave the Cluster/Machine interfaces in `platform`, reducing that package to just be interfaces in the end.

kola/harness.go
```diff
@@ -399,7 +399,7 @@ func RunTest(t *register.Test, pltfrm string) (err error) {
 
 	// Cluster -> TestCluster
 	tcluster := platform.TestCluster{
-		Name:        t.Name,
+		TestName:    t.Name,
```
marineam: This whole PR is pretty damn big. Lets split it up into even smaller commits and smaller PRs so you can get some of this churn out of the way quickly. e.g. Renaming Name to TestName probably should be its own commit and easily submitted independently of the rest. Pruning code from ore prior to the refactor could also stand alone.

PR #349, opened 2016-07-21
cmd/ore: remove create command, subsumed by kola spawn

mischief:


pbx0:
https://github.com/coreos/mantle/issues/164 is related. This doesn't seem unreasonable but why do you think this command is better served in kola over the ore command? Manually spawning instances seems more like a helper tool and less related to kola testing (Though, if your plan was to deprecate ore, I would understand). My main concern is just UX around what the purpose of each tool in mantle provides so people know the right place to look for certain things.

Also, I know the mantle readme.md isn't' really maintained, but we should delete the docs for the ore create command if this goes through and add the kola spawn command.

marineam:
I'm fine with the functionality being in either, perhaps ore is the better place since kola is primarily for the automated test suite. If so the change should then be to pull whatever functionality the kola command provides that this doesn't into this one.

mischief:
i put spawn in kola since it already runs vms anyway - i think ore is a fine place for image management only. as a result of removing vm creation, the binaries will be smaller and compile faster.

we should definitely talk about the future of the various mantle components OOB.

marineam:
well, no reason for create/spawn to be a heavy hitter compile-wise though since creating a vm is just about interacting with APIs. Don't need to mix vm creation with the full kola cluster magic do we?

pbx0:
That seems like a reasonable division to me. But if manual instance spawning is to be relegated to kola spawn then I think a few more things need to be cleaned up first:

-  `ore destroy-instances` functionality should also be deleted and moved into kola spawn somehow. (if kola spawn is used with remove=false, how do you delete it?)

- `ore list-instances` should also be deleted and have kola spawn cover its use case at least to some degree.

Also worth noting that in the past I've used `ore list-images` somewhat extensively with these other commands to quickly see what images I can pass in. Not really sure if that should go along with these other commands or not. And if thats too much for this PR maybe we can just track the general plan here. 

One big feature of `kola spawn` over `ore create-instances` seems to be that it actually supports all the platforms. Ore is gce specific which make sense for certain things since we need gce storage specific helper utilities for dealing with our image hosting and stuff. However, spawning instances should surely support all platforms like kola does. The idea that kola functionality should support all platforms and ore not necessarily needing to does give us another division point to divide functionality between ore and kola. Anyway, I'm not sure what the right abstraction is here, but, I think this is a good place to create a plan for it even if its not all done in one PR.

mischief:
to capture OOB chat:

going forward, we'll be moving kola spawn stuff into ore subcommands:

- ore
    - gce
        - create-instance
        - create-image
    - aws
        - create-instance
        - create-image
    - ...

pbx0:
That sounds like a good cleanup to me. Also, I can see how the spawn ui is simpler to use then the current ore commands, so, if you wanted to bring that design over and make it `ore gce spawn` I would be in favor of it.

Once that done we should probably file an issue for bringing `destroy-instances` and `list-instanes` into line as well.

mischief:
yeah, the cleanup would include those other commands as well. :-)

PR #355, opened 2016-08-03
platform: export machine check function

mischief:


vcaputo:
lgtm, do you have a patch lined up which takes advantage of these newly public functions?  It might make more sense to put them all in one PR so the publicizing and external usage arrives in a single merge....

mischief:
yes, the intention is to split off the cluster implementations from the platform interfaces. to do that, some functionality needs to be exported.

PR #362, opened 2016-08-08
platform: split gce api/cluster out

mischief:


platform/api/gcloud/image.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"google.golang.org/api/compute/v1"
+)
+
+// CreateImage creates an image on GCE and and wait for completion. If
+// overwrite is true, an existing image will be overwritten if it exists.
+func (a *API) CreateImage(name, source string, overwrite bool) error {
```
marineam: In order for this to be useful for plume it is going to need to accept a full compute.Image because there are a number of options that need to be set like family and whatnot. That's what my aborted attempt at the api refactor was doing.
￼

marineam:
Why gcloud instead of GCE?

platform/api/gcloud/image.go
mischief: can we stuff them in Options?

mischief:
@marineam the package name reflects the umbrella term for google cloud platform, like aws covers ec2 and s3. `package gcp` might be better, but i don't really care. i can rename it to gce if that's better.

platform/api/gcloud/image.go
marineam: Don't know, but eh, lets not worry about that until attempting to port plume. Can always change the API later.

marineam:
LGTM

PR #366, opened 2016-08-15
platform: Enable selinux when bringing up or rebooting machines

mjg59:
Ensure that selinux is enabled when running tests - we want to ensure that
thing we expect to work.


vcaputo:
Is this ideal?  It seems like it would be preferable to have the machines just boot with selinux enabled/restrictive mode, unless I'm missing something...

mjg59:
@vcaputo That would be preferable, but there's no straightforward way for us to do this right now. Once we have good Ignition support via fw_cfg for qemu, we can switch to that instead.

vcaputo:
oki doke, lgtm

PR #368, opened 2016-08-15
kola/tests/kubernetes: test docker and rkt runtime

pbx0:
Bumps this test to using our v1.3 coreos-kubernetes scripts found [here](https://github.com/coreos/coreos-kubernetes/tree/master/multi-node/generic) which allows testing of kubernetes with rkt as a runtime.

I generally am OK leaving this test as just testing one version of kubernetes and serving as smoke test for the OS since the scripts in coreos-kubernetes only really support one minor version of kubernetes at a time. 

Tested rkt and docker runtimes on 1122.0.0, tested docker runtime on 1068.9

### DIFF
--- a/kola/tests/kubernetes/basic.go
+++ b/kola/tests/kubernetes/basic.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
 	"github.com/coreos/mantle/platform"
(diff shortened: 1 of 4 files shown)


mischief:
does this work on gce under kola?

can you document where the scripts are coming from?

pbx0:
Yes, I tested this using kola on 1122.0.0 for both runtimes and 1068.9 for just the docker runtime only (since for the rkt test, I've set to minVersion 1122).

Also, added some comments to `setup.go controllerInstall.go and workerInstall.go` reflecting where the scripts come from.

mischief:
rebase and lgtm

pbx0:
rebased

PR #369, opened 2016-08-16
locksmith cluster test

mischief:


kola/tests/locksmith/locksmith.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		output, err := m.SSH("sudo systemctl restart --no-block etcd2 locksmithd.service")
+		if err != nil {
+			return fmt.Errorf("failed to start etcd2.service on %s: output %q: error: %v", m.ID(), output, err)
+		}
+	}
+
+	// make sure etcd is ready
+	etcdCheck := func() error {
```
pbx0: nit: The etcd test exports a function that check's etcd readyness: https://github.com/coreos/mantle/blob/master/kola/tests/etcd/util.go#L34

I don't think it really matters what you use though. I believe I exported it for use in the kubernetes test.

mischief: this is specifically checking that locksmithctl can reach etcd.

unfortunately etcdctl cluster-health is still borked and has a zero exit code..

pbx0: Sounds good. Although, I think etcd fixed the exit code: https://github.com/coreos/etcd/pull/3905 (and you commented on it ;)

mischief: seems that it does work now for 2.3.2 in the os, dunno what went wrong before when i tried it.

pbx0:
lgtm

PR #371, opened 2016-08-17
kola/tests/rkt: make etcd-wrapper test work again

mischief:
the service name changed from etcd3-wrapper to etcd-wrapper, and we must
specify a tag now.


vcaputo:
lgtm

PR #374, opened 2016-08-17
kola/tests/ignition: do fs operations on /dev/disk/by-partlabel/ROOT

mischief:
operating on /dev/disk/by-label/ROOT appears to race with udev, at least
when using mkfs.btrfs, so switch to /dev/disk/by-partlabel/ROOT which
appears to not race.


marineam:
LGTM

PR #378, opened 2016-08-23
kola/tests/locksmith: use send-need-reboot instead of reboot

mischief:
this exercises locksmithd's locking logic, and will retry if locking
fails. locksmithctl reboot will not retry or block if the lock is
already taken.


vcaputo:
lgtm

PR #381, opened 2016-08-30
introduce azure api bindings and ore azure subcommand

mischief:
for now, i've only added upload-blob until upstream azure-sdk-for-go is improved for our needs.

the azure subcommand only builds under go1.7 because of the azure api requirement of TLS renegotiations.

### DIFF
--- a/platform/api/azure/api.go
+++ b/platform/api/azure/api.go
@@ -1,0 +1,47 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package azure
+
+import (
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/management"
+	"github.com/coreos/pkg/capnslog"
+)
+
+var (
+	plog = capnslog.NewPackageLogger("github.com/coreos/mantle", "platform/api/azure")
+)
+
+type API struct {
+	client management.Client
+	opts   *Options
+}
+
+// New creates a new Azure client. If no publish settings file is provided or
+// can't be parsed, an anonymous client is created.
+func New(opts *Options) (*API, error) {
+	client, err := management.ClientFromPublishSettingsFile(opts.PublishSettingsFile, "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create azure client: %v", err)
+	}
+
+	api := &API{
+		client: client,
+		opts:   opts,
+	}
+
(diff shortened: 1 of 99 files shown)


platform/api/azure/api.go
```diff
@@ -0,0 +1,46 @@
...
+	plog = capnslog.NewPackageLogger("github.com/coreos/mantle", "platform/api/azure")
+)
+
+type API struct {
+	client management.Client
+	opts   *Options
+}
+
+// New creates a new Azure client. If no publish settings file is provided or
+// can't be parsed, an anonymous client is created.
```
marineam: If the user expects to be logged in the failure should be fatal.

marineam:
Other than the auth downgrade this looks good.

platform/api/azure/api.go
mischief: unless i defer creation of the client until the api call, a user will always have to pass a publish settings file even if it isn't needed. it's not needed in the case of UploadBlob, which only uses storage account credentials.

mischief: is it more reasonable to always require both publishsettings file and storage credentials even when one set of credentials might not be used?

mischief: alternately, i could split the API into two parts.

crawford: The storage credentials can be derived from the publish settings.

marineam: Always using publish settings and deriving storage credentials from that sounds like a fairly user friendly scheme.

mischief:
ptal

platform/api/azure/storage_mit.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/Microsoft/azure-vhd-utils-for-go/upload"
+	"github.com/Microsoft/azure-vhd-utils-for-go/upload/metadata"
+	"github.com/Microsoft/azure-vhd-utils-for-go/vhdcore/common"
+	"github.com/Microsoft/azure-vhd-utils-for-go/vhdcore/diskstream"
+	"github.com/coreos/pkg/multierror"
+)
+
+const pageBlobPageSize int64 = 2 * 1024 * 1024
+
+// UploadBlob
```
crawford: Expand?

platform/api/azure/storage_mit.go
```diff
@@ -0,0 +1,205 @@
...
+	return nil
+}
+
+// getBlobMetaData returns the custom metadata associated with a page blob which is set by createBlob method.
+// The parameter client is the Azure blob service client, parameter containerName is the name of an existing container
+// in which the page blob resides, parameter blobName is name for the page blob
+// This method attempt to fetch the metadata only if MD5Hash is not set for the page blob, this method panic if the
+// MD5Hash is already set or if the custom metadata is absent.
+//
+func getBlobMetaData(client storage.BlobStorageClient, containerName, blobName string) (*metadata.MetaData, error) {
```
crawford: getBlobMetadata

crawford:
Why is that not vendored?

crawford:
Looks reasonable.

mischief:
@crawford because their version of "UploadBlob" is a gigantic function inside of package main: https://github.com/Microsoft/azure-vhd-utils/blob/master/vhdUploadCmdHandler.go

crawford:
uhhhhhhhhhhh, okay. LGTM assuming the licensing is sorted out.

PR #382, opened 2016-09-06
Remove the systemd-resolved listeners

dm0-:
CoreOS is now shipping systemd-resolved without enabling its DNS stub resolver by default.


mischief:
@dm0- is this the case now for all release channels? otherwise we have to split this up and gate the tests on coreos versions....

dm0-:
Yes.

  * The systemd version in stable has not yet gone to v231, so it never had that listener.
  * Same for beta, except the newly tagged beta now uses v231 with the resolver patched out.
  * Only the current alpha had the active listener, it is disabled for the build tomorrow.

mischief:
lgtm

PR #384, opened 2016-09-08
ore: azure image replication

mischief:


cmd/ore/azure/replicate-image.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdReplicateImage = &cobra.Command{
+		Use:   "replicate-image image offer sku version [regions...]",
```
marineam: With this many positional parameters I would favor making them required options instead. And regions could be a repeatable --region option. Then you can offer more help text on what the heck an "offer" or an "sku" is because I have no idea.

mischief:
ptal

marineam:
LGTM

PR #388, opened 2016-09-15
Check for terminated ssh sessions during reboot

marineam:
x/crypto/ssh recently added `ExitMissingError` to signal sessions that were cleanly terminated w/o any command exit status as typically happens when calling reboot. This method works for old CoreOS versions where systemd-run lacks the --no-block option.


mischief:
lgtm after rebase

PR #389, opened 2016-09-15
platform: optionally dump machine log when a unit fails

marineam:
This makes it practical to actually find out why a unit failed.


mischief:
might be better to gate this behind a different option.. this could get really verbose :-)

marineam:
debug *is* really verbose. e.g. the updatepayload test streams the machine log in debug mode. But that only starts after checkmachine passes so if the machine fails early there isn't any info.

PR #390, opened 2016-09-15
platform/conf: pass through scripts unmodified

marineam:
Lets me get away with some shenanigans. When doing this the script must setup ssh keys itself.


mischief:
just curious, what have you used this for?

after rebase lgtm

marineam:
@mischief converting `kola updatepayload` to work with all versions of CoreOS in order to test different upgrade paths.

PR #391, opened 2016-09-15
updatepayload fixes

marineam:


cmd/kola/updatepayload.go
```diff
@@ -196,7 +196,9 @@ func splitSpaceEnv(envs string) map[string]string {
 	pairs := strings.Fields(envs)
 	for _, p := range pairs {
 		spl := strings.SplitN(p, "=", 2)
-		m[spl[0]] = spl[1]
+		if len(spl) == 2 {
+			m[spl[0]] = spl[1]
+		}
```
mischief: should len(spl)==1 simply set the map key to ""?

marineam: Could, but we don't care about such options in this code.

PR #392, opened 2016-09-16
plume pre-release

mischief:
for now, runs stuff for azure. in the future it could do the aws stuff too.

### DIFF
--- a/cmd/plume/prerelease.go
+++ b/cmd/plume/prerelease.go
@@ -1,0 +1,238 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/Microsoft/azure-vhd-utils-for-go/vhdcore/validator"
+	"github.com/spf13/cobra"
+	"golang.org/x/net/context"
+
+	"github.com/coreos/mantle/auth"
+	"github.com/coreos/mantle/platform/api/azure"
+	"github.com/coreos/mantle/sdk"
+	"github.com/coreos/mantle/storage"
+	"github.com/coreos/mantle/util"
+)
+
+var (
+	preReleaseDryRun bool
+	cmdPreRelease    = &cobra.Command{
+		Use:   "pre-release [options]",
+		Short: "Run pre-release steps for CoreOS",
+		Long:  "Runs pre-release steps for CoreOS, such as image uploading and OS image creation, and replication across regions.",
+		RunE:  runPreRelease,
+	}
+
+	azureOpts = azure.Options{}
+)
+
+func init() {
+	cmdPreRelease.Flags().StringVar(&azureOpts.PublishSettingsFile, "publish-settings", "", "publish settings file")
+	cmdPreRelease.Flags().BoolVarP(&preReleaseDryRun, "dry-run", "n", false,
+		"perform a trial run, do not make changes")
+
+	AddSpecFlags(cmdPreRelease.Flags())
+	root.AddCommand(cmdPreRelease)
+}
+
+func runPreRelease(cmd *cobra.Command, args []string) error {
+	if len(args) > 0 {
+		return errors.New("no args accepted")
+	}
+
+	spec := ChannelSpec()
+	ctx := context.Background()
+	client, err := auth.GoogleClient()
+	if err != nil {
+		return err
+	}
+
+	src, err := storage.NewBucket(client, spec.SourceURL())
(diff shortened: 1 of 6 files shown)


cmd/plume/specs.go
```diff
@@ -93,6 +107,16 @@ var (
 				Publish:     "coreos_production_gce.txt",
 				Limit:       25,
 			},
+			Azure: azureSpec{
+				Image:             "coreos_production_azure_image.vhd.bz2",
+				StorageAccount:    "coreos",
+				Containers:        []string{"publish", "pre-publish"},
+				Label:             "CoreOS alpha",
```
crawford: CoreOS Alpha

cmd/plume/specs.go
```diff
@@ -128,6 +152,16 @@ var (
 				Publish:     "coreos_production_gce.txt",
 				Limit:       25,
 			},
+			Azure: azureSpec{
+				Image:             "coreos_production_azure_image.vhd.bz2",
+				StorageAccount:    "coreos",
+				Containers:        []string{"publish", "pre-publish"},
+				Label:             "CoreOS beta",
```
crawford: CoreOS Beta

cmd/plume/specs.go
```diff
@@ -153,6 +187,16 @@ var (
 				Publish:     "coreos_production_gce.txt",
 				Limit:       25,
 			},
+			Azure: azureSpec{
+				Image:             "coreos_production_azure_image.vhd.bz2",
+				StorageAccount:    "coreos",
+				Containers:        []string{"publish", "pre-publish"},
+				Label:             "CoreOS stable",
```
crawford: CoreOS Stable

cmd/plume/prerelease.go
```diff
@@ -0,0 +1,233 @@
...
+		return err
+	}
+
+	src, err := storage.NewBucket(client, spec.SourceURL())
+	if err != nil {
+		return err
+	}
+
+	// Sanity check!
+	if vertxt := src.Object(src.Prefix() + "version.txt"); vertxt == nil {
```
crawford: What is the point of this check?

cmd/plume/prerelease.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	if err := validator.ValidateVhdSize(unzipped); err != nil {
+		return err
+	}
+
+	// upload blob, do not overwrite
+	plog.Printf("Uploading %q to Azure Storage...", unzipped)
+
+	blobName := fmt.Sprintf("coreos-%s-%s.vhd", specVersion, specChannel)
```
crawford: What is specVersion and specChannel?

cmd/plume/prerelease.go
```diff
@@ -0,0 +1,233 @@
...
+		vhduri, err := url.Parse(spec.Azure.Image)
+		if err != nil {
+			return err
+		}
+
+		vhduri = src.URL().ResolveReference(vhduri)
+
+		plog.Printf("Downloading Azure image %q to %q", vhduri, vhdfile)
+
+		if err := sdk.UpdateFile(vhdfile, vhduri.String(), client); err != nil {
```
marineam: Never touch unverified data in plume. Use `sdk.UpdateSignedFile`

cmd/plume/prerelease.go
```diff
@@ -0,0 +1,233 @@
...
+
+	_, err = io.Copy(outf, bzr)
+	return err
+}
+
+// azurePreRelease runs everything necessary to prepare a CoreOS release for Azure.
+//
+// This includes uploading the vhd image to Azure storage, creating an OS image from it,
+// and replicating that OS image.
+func azurePreRelease(ctx context.Context, client *http.Client, src *storage.Bucket, spec *channelSpec) error {
```
marineam: This function is way to big, break up the major steps into smaller functions please. Basically any time you have `if ... { log skip } else { lots of stuff }` can turn into `func foo() { if ... { log skip; return}; lots of stuff }`

cmd/plume/specs.go
mischief: this is simply mirroring https://github.com/coreos/scripts/blob/master/oem/azure/set-image-metadata.sh#L31, same for the others

crawford: There is a difference in capitalization between labels and image names. Labels should be title case.

mischief: ah, i missed `GROUP="${1^}"`

cmd/plume/prerelease.go
mischief: it ensures version.txt exists. the same check is in https://github.com/coreos/mantle/blob/master/cmd/plume/release.go#L75

cmd/plume/prerelease.go
mischief: from https://github.com/coreos/mantle/blob/master/cmd/plume/specs.go#L160, effectively they are just what you give plume pre-release as arguments.

cmd/plume/prerelease.go
crawford: I get that, but what is the intent? The contents are never used. I assume it's to make sure that the version exists in the bucket, but why not just let it 404 when pulling the image?

cmd/plume/prerelease.go
```diff
@@ -0,0 +1,238 @@
...
+	if spec.Azure.StorageAccount == "" {
+		plog.Notice("Azure image creation disabled.")
+		return nil
+	}
+
+	// construct azure api client
+	plog.Printf("Creating Azure API from %q", azureOpts.PublishSettingsFile)
+	a, err := azure.New(&azureOpts)
```
crawford: Can you use a more descriptive name? I realize it's not idiomatic, but that idiom is idiotic.

marineam:
Please fix travis before merging.

PR #393, opened 2016-09-16
Support upgrading from all CoreOS versions

marineam:
Tested with every CoreOS stable branch released so far. Extends the test to check that the upgraded system itself can perform a second upgrade.


mischief:
@marineam is there some script you have used to do the testing? might be good to stick in in mantle/ or scripts/.

marineam:
@mischief yes, I'll toss the script somewhere soon.

PR #394, opened 2016-09-16
platform: fix enabling SELinux

marineam:
The return values were swapped so the error check was actually testing
for non-nil stdout. However error messages generally go to stderr so
failures were always ignored. Fix error check and just ignore stdout.

In fixing this I need to still tolerate images that do not have SELinux,
such as ancient versions (for upgrade tests) or non-amd64 ports.


mischief:
haha, nice catch.

PR #395, opened 2016-09-21
Remove go1.7 build tags and bump godeps to 1.7

mischief:


marineam:
Looks good as long as both Travis and Jenkins pass

PR #400, opened 2016-09-28
kola/tests/locksmith: properly test locksmith

crawford:
Using coreos-cloudinit to configure and start locksmith properly
exercises the codepath that is encountered in the field. I have no idea
why this test was originally implemented this way, but this new
implementation is more inline with what is actually done.

This now catches https://github.com/coreos/bugs/issues/1562 and https://github.com/coreos/bugs/issues/1588.

### DIFF
--- a/kola/tests/locksmith/locksmith.go
+++ b/kola/tests/locksmith/locksmith.go
@@ -16,7 +16,6 @@ package locksmith
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


crawford:
Jenkins failure is due to it testing against 1180.0.0 (expected). I asked it to retest against 1185.0.0.

crawford:
:/ The status wasn't updated. Here is the new test result: https://jenkins.coreos.systems/job/kola-dev-qemu/505/.

marineam:
@crawford to get the status to update you need to trigger the rebuild from here instead of jenkins.

marineam:
retest this please

PR #403, opened 2016-10-07
sdk: work around gsutil umask bug

marineam:
gsutil changes the global process umask when writing out
~/.gsutil/credstore but gsutil is multi-threaded so this may conflict
with the automatic creation of the ~/.gsutil/tracker-files directory.

Side step the whole mess by replacing ~/.gsutil with a path under /run
that already contains the tracker-files directory.

### DIFF
--- a/sdk/enter.go
+++ b/sdk/enter.go
@@ -44,8 +44,10 @@ client_id = {{.ClientID}}
 client_secret = {{.ClientSecret}}
 {{else}}{{if eq .Type "service_account"}}
 [Credentials]
-gs_service_key_file = {{.Path}}
+gs_service_key_file = {{.JsonPath}}
 {{end}}{{end}}
+[GSUtil]
+state_dir = {{.StateDir}}
 `))
 )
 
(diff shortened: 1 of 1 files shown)


marineam:
retest this please

sdk/enter.go
```diff
@@ -44,8 +44,10 @@ client_id = {{.ClientID}}
 client_secret = {{.ClientSecret}}
 {{else}}{{if eq .Type "service_account"}}
 [Credentials]
-gs_service_key_file = {{.Path}}
+gs_service_key_file = {{.CredsPath}}
```
mischief: is this supposed to be JsonPath?

marineam: Oops, yes. weird it didn't explode...

marineam: fixed

marineam: That claims it is for maps, not structs. I think I didn't catch it just because it is inside a conditional that didn't apply to my testing.

PR #404, opened 2016-10-08
ensure docker containers can make network connections outside of their hosts

joeatwork:
Proposes
- Some minor changes to kola spawn, to make it easier to develop and prove out cluster tests
- A new test, that ensures that docker containers can receive and create connections to remote hosts

Bits worth thinking about:
- The kola spawn changes might run counter to decisions made in https://github.com/coreos/mantle/pull/349 
- The timeouts in the docker test are kinda hacky

### DIFF
--- a/cmd/kola/spawn.go
+++ b/cmd/kola/spawn.go
@@ -36,15 +36,19 @@ var (
 		Short:  "spawn a CoreOS instance",
 	}
 
-	spawnUserData string
-	spawnShell    bool
-	spawnRemove   bool
+	spawnNodeCount int
+	spawnUserData  string
+	spawnShell     bool
+	spawnRemove    bool
+	spawnVerbose   bool
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


marineam:
What do you mean by the timeouts being hacky?

joeatwork:
Re: hacky timeouts: It feels like the right way to do a timeout would be `machine.SSH(ctx, "COMMAND")` or something similar, but making that work felt kind of invasive. Instead, I set two timeouts - one in a spin wait for the server to come up, and another on the server command line itself. I proposed that because messing with the SSH seems like it would be pretty invasive, and might complicate *all* of the tests for the benefit of one.

marineam:
Ah yes, plumbing through context support *is* something we want to do but best done independently of adding a test. Now that we are on 1.7 there should be a lot of opportunities to do that. Our system/exec is another package waiting on context support.

marineam:
Also on the kola-spawn thing, I'd rather merge useful code than block on larger refactoring but that is something to keep in mind if you have time.

PR #405, opened 2016-10-12
cmd/cork: Add setup command

glevand:
In follow up to https://github.com/coreos/coreos-overlay/pull/2212.

@marineam  There are a lot of ways to do this.  Let me know if you have some preference.

### DIFF
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -55,6 +55,11 @@ var (
 	downgradeReplace bool
 	newVersion       string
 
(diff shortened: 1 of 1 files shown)


glevand:
Rebased to latest.

marineam:
Possible future extension: proactively register w/ /proc/sys/fs/binfmt_misc/register. Or at least log and tell the user the change doesn't take effect until the host's systemd-binfmt.service runs.

PR #409, opened 2016-10-20
platform/api/gcloud: allow http and https

pbx0:
This tags are needed in addition to the default firewall rules for the
project which in our case already allow it.


pbx0:
Merge when ready, I have not the rights anymore. @ethernetdan says it works for him

PR #410, opened 2016-10-20
use azure profiles instead of publishsettings

mischief:


auth/azure.go
```diff
@@ -0,0 +1,93 @@
...
+	Environments  []AzureEnvironment  `json:"environments"`
+	Subscriptions []AzureSubscription `json:"subscriptions"`
+}
+
+func (ap *AzureProfile) AsOptions() []azure.Options {
+	var o []azure.Options
+
+	for _, sub := range ap.Subscriptions {
+		newo := azure.Options{
+			SubscriptionID:        sub.ID,
```
crawford: A lot more of these settings are going to change. Just about everything in `AzureEnvironment` will defer depending on the region.

mischief: nothing else is used in the azure api client currently, so i don't think the environments matter right now.

crawford: StorageEndpointSuffix isn't needed when uploading the blobs?

mischief: could be, it seems i'm always using `DefaultBaseURL` currently, which is `"core.windows.net"`

mischief:
@crawford @marineam ptal

mischief:
this should likely be reworked to just read azure profile from `~/.azure/azureProfile.json`, like the cli and like the google token stuff in `auth` package.

auth/azure.go
```diff
@@ -0,0 +1,104 @@
+// Copyright 2014 CoreOS, Inc.
```
marineam: This file on the other hand *can* be 2016 ;-)

cmd/ore/azure/azure.go
```diff
@@ -30,20 +31,53 @@ var (
...
+	if len(opts) == 0 {
+		plog.Fatalf("Azure Profile %q contains no subscriptions", azureProfile)
+	}
+
+	if azureSubscription == "" {
+		opt = &opts[0]
+	} else {
+
+		for _, o := range prof.AsOptions() {
+			if o.SubscriptionName == azureSubscription {
```
mischief: sounds good.

mischief:
retest this please

mischief:
@crawford sounds like an off-by-one, as 2M is the chunk size.

is this reproducible with `ore azure upload-blob`? they use the same code.

crawford:
I'm pretty sure it is, but I am not certain. We had to republish everything that was originally published by this program.

crawford:
This is a good start. There is still some work that needs to be done around the UX, but it's good enough for now.

PR #411, opened 2016-10-21
kola/vendor: vendor old docker binary, for use in regression tests

joeatwork:
This one bears some explaining. I'd like to refer to the associated binary in a kola test, which means I need to refer to it in an emerge so the file shows up during test runs (I'm planning to install it near the kolet, and `DropFile()` it during a regression test)

I'd like to do that with the following commit strategy:

- Merge this PRQ, with the unused file present in kola
- Merge a PRQ in coreos-overlay that refers to the new file, and installs it on the file system
- Finally, merge a kola test that assumes the new file will be available and installed

That's three separate merges. Sound ok?

### DIFF
--- a/cmd/kola/spawn.go
+++ b/cmd/kola/spawn.go
@@ -36,11 +36,12 @@ var (
 		Short:  "spawn a CoreOS instance",
 	}
 
-	spawnNodeCount int
-	spawnUserData  string
-	spawnShell     bool
-	spawnRemove    bool
-	spawnVerbose   bool
+	spawnNodeCount      int
+	spawnProcessConfigs bool
+	spawnUserData       string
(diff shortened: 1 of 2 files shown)


mischief:
i really do not like vendoring a docker binary.

i'd rather we just build docker from source on the to-be-tested machine...

joeatwork:
Sounds good! I originally proposed something like https://github.com/coreos/coreos-overlay/pull/2253 but I think there were reservations from the team. (I also tried to vendor and build an old docker lib as part of mantle, but that got very complex.)

Are there examples of pushing/providing test-only packages during tests? In this case, we'd need go1.4, an old version of docker, etc.

PR #412, opened 2016-10-21
Old docker client test

joeatwork:
UPDATED: Running the new test now depends on (ONLY) https://github.com/coreos/coreos-overlay/pull/2257 - this can also merge before that, since the new test is skipped unless the old client is available.

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
 	"time"
 
 	"github.com/coreos/go-semver/semver"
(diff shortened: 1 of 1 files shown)


kola/tests/docker/docker.go
```diff
@@ -64,6 +65,18 @@ func init() {
 
 		MinVersion: semver.Version{Major: 1192},
 	})
+	register.Register(&register.Test{
```
marineam: We aren't (yet) running kola on arm but do we have a way to restrict this to amd64? If not this can go in but file a bug to fix that since we do need to start testing arm at some point here...

marineam:
The title of this PR makes no sense

kola/tests/docker/docker.go
joeatwork: Right now we don't have any way to limit architectures in Kola (that I know of, I asked around a bit before committing this)

joeatwork:
Title changed (the "later" referred to an elaborate "merge that first, then this, then that" setup for an early version of the changes)

kola/tests/docker/docker.go
marineam: Ok, just file a bug so we don't forget then :)

marineam:
Fix the CI jobs before merging though :)

kola/tests/docker/docker.go
joeatwork: Filed at https://github.com/coreos/mantle/issues/423

PR #418, opened 2016-10-26
kola/tests: update etcd wrapper test

crawford:
This variable name changed with the new implementation of the wrapper
script. We aren't worrying about backward compatibility here because the
existance of this script was never advertised.


marineam:
Oh, fix  		MinVersion:  semver.Version{Major: 1122},

PR #425, opened 2016-11-18
make importing OS tests optional from using harness

pbx0:
I have a project vendoring mantle to maintain a separate test suite from the OS test suite. In this scenario I have no control over whether the OS test suite is registered if I wish to use the kola harness from my main package. I could just edit the kola code but I'd like to strictly maintain it as vendored upstream library so all changes to kola go back to this repo.

Moving the registration list to a separate package explicity imported by the main package allows the projects vendoring this code to choose not to register the OS test suite.

Alternatively this list of imports could all just go in the main package instead of having this indirect registry package. I have no opinion on which I like more.

cc @mischief  @crawford

### DIFF
--- a/cmd/kola/kola.go
+++ b/cmd/kola/kola.go
@@ -26,6 +26,9 @@ import (
 	"github.com/coreos/mantle/cli"
 	"github.com/coreos/mantle/kola"
 	"github.com/coreos/mantle/kola/register"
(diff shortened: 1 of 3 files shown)


marineam:
Jenkins build failure is unrelated to this change. So merge at will.

PR #426, opened 2016-11-18
platform: change Machine.SSH to also return stderr

pbx0:
This is probably the most used function of the platform interfaces used
by test writers. Adding stderr allows test writers to provide far more
helpful error messages without reaching for the verbose SSHClient to
accomplish that. Often SSH commands that fail just return the err code
up the stack which is useless. Also, sometimes the stderr output
provides useful logs even if the command didn't fail which is why I
didn't just replace the error code with stderr.

### DIFF
--- a/platform/base.go
+++ b/platform/base.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"os"
 	"sync"
 
 	"github.com/coreos/pkg/multierror"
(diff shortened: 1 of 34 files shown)


marineam:
The existing behavior of passing through stderr is much better than discarding the output entirely as this change does for existing code. So if you want to deal with the stderr output I'd recommend adding a new function with this new behavior.

platform/base.go
```diff
@@ -76,25 +75,31 @@ func (bc *BaseCluster) PasswordSSHClient(ip string, user string, password string
 	return sshClient, nil
 }
 
-func (bc *BaseCluster) SSH(m Machine, cmd string) ([]byte, error) {
+func (bc *BaseCluster) SSH(m Machine, cmd string) (stdout, stderr []byte, err error) {
```
marineam: This should be a new function to avoid the temptation to discard stderr when test writers don't particularly care about it.

pbx0:
@marineam Sounds good, I will add a second method. SSHOutput? SSHOut? SSH2? SSHWithStderr?

Additionally, would you consider keeping the current SSH signature the same except that, on error, we return stderr in the error instead of the error code (I've never witnessed this to be useful other then knowing `err != nil`). If no error is returned from the cmd, stderr could still be printed to stdout. My hope is that this would enable better error handling, but just not cover the use-case of needing to use or silence stderr output when the command doesn't fail. (I'm thinking of you, openssl commands with spurious output)

marineam:
Hm, maybe os/exec may offer inspiration here? https://golang.org/pkg/os/exec/#ExitError

Tricky thing is output in stderr may or may not correlate with the command experiencing a fatal error. Maybe we stick with one SSH command but instead of connecting stderr directly to the terminal we can Tee it. One direction logs immediately, optionally adding a prefix so it is clear in the log what the output is from. The other direction is a limited buffer that just keeps the last line or two for inclusion in the returned err if the command happens to fail.

marineam:
For reference here is a generic bit of code that dumps an io stream to the log: https://github.com/coreos/mantle/blob/master/util/logio.go#L29

For this case it would probably need to be reformulated as a Writer though.

euank:
Closing in favor of https://github.com/coreos/mantle/pull/725

The ergonomics issue got handled there with a `TestCluster.SSH` helper, and the few remaining callers just deal with it.

PR #428, opened 2016-12-16
platform/api/gcloud: respect basename option

pbx0:
Somewhere in the shuffle this option stopped being respected. This means its difficult to tell from which CI job rogue kola instances are from and know which ones are safe to reap. Useful now that kola tests are being run in more places also won't harm anything not utilizing the basename option.

cc @marineam


pbx0:
Wasn't even close to getting that name right :). Compiles now.

PR #429, opened 2016-12-20
allow selection of image project for gce

pbx0:
The gcloud should distinguish the image project from the instance project. Mainly, its useful to be able to select vm images from a project like coreos-cloud but have the instances still spawned on the coreos-gce-testing project. Otherwise one has to guess which jenkin's created image is good. 

cc @marineam @mischief

### DIFF
--- a/platform/api/gcloud/compute.go
+++ b/platform/api/gcloud/compute.go
@@ -46,10 +46,11 @@ func (a *API) mkinstance(userdata, name string, keys []*agent.Key) *compute.Inst
 		})
 	}
 
-	prefix := "https://www.googleapis.com/compute/v1/projects/" + a.options.Project
+	instancePrefix := "https://www.googleapis.com/compute/v1/projects/" + a.options.Project
+
 	instance := &compute.Instance{
 		Name:        name,
-		MachineType: prefix + "/zones/" + a.options.Zone + "/machineTypes/" + a.options.MachineType,
+		MachineType: instancePrefix + "/zones/" + a.options.Zone + "/machineTypes/" + a.options.MachineType,
 		Metadata: &compute.Metadata{
 			Items: metadataItems,
 		},
@@ -66,7 +67,7 @@ func (a *API) mkinstance(userdata, name string, keys []*agent.Key) *compute.Inst
 				Type:       "PERSISTENT",
 				InitializeParams: &compute.AttachedDiskInitializeParams{
 					DiskName:    name,
-					SourceImage: prefix + "/global/images/" + a.options.Image,
+					SourceImage: a.options.Image,
 					DiskType:    "/zones/" + a.options.Zone + "/diskTypes/" + a.options.DiskType,
 				},
 			},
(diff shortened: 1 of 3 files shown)


marineam:
Alternatively, should we support longer form image paths that include the
project?

pbx0:
@marineam If we support longer-form image paths we could reduce `--gce-image` and `--gce-image-project` to one flag, but it would be a breaking change which I don't think is that big a deal for us.  Also we should change `--gce-project` to the long-form for parity between the flags.

If your OK with a minor breakage of jenkins jobs i'm happy to change to long-form.

marineam:
I meant support both, assume the project id when if the image name doesn't
contain the parent project path. Other CLI tools behave similarly.

That behavior wouldn't break anything would it?

pbx0:
Oh, yes, that'd be fine. I will update to support long and short form names for the images.

pbx0:
Updated so that the image project will be taken from the image name if it is longform and from the Project flag otherwise.

platform/api/gcloud/compute.go
```diff
@@ -66,7 +73,7 @@ func (a *API) mkinstance(userdata, name string, keys []*agent.Key) *compute.Inst
 				Type:       "PERSISTENT",
 				InitializeParams: &compute.AttachedDiskInitializeParams{
 					DiskName:    name,
-					SourceImage: prefix + "/global/images/" + a.options.Image,
+					SourceImage: imagePrefix + "/global/images/" + a.options.Image,
```
marineam: If Image is the full path `https://www.googleapis.com/compute/v1/projects/projname/global/images/imgname` Won't this create URLs like `https://www.googleapis.com/compute/v1/projects/projname/global/images/imgname/global/images/https://www.googleapis.com/compute/v1/projects/projname/global/images/imgname`?

platform/api/gcloud/compute.go
```diff
@@ -46,10 +46,17 @@ func (a *API) mkinstance(userdata, name string, keys []*agent.Key) *compute.Inst
 		})
 	}
 
-	prefix := "https://www.googleapis.com/compute/v1/projects/" + a.options.Project
+	instancePrefix := "https://www.googleapis.com/compute/v1/projects/" + a.options.Project
+
+	// If long image name given for image, respect the image project
+	imagePrefix := instancePrefix
+	if strings.HasPrefix(a.options.Image, "https") {
```
marineam: I don't know about other tools off the top of my head but the old gcutil accepted image names relative to the API endpoint, e.g. `projects/name/global/images/foo`

So how about this to be as flexible as possible:
 - If prefix is the API endpoint `https://www.googleapis.com/compute/v1/` pass the value through verbatim.
 - If prefix is `projects/` append it to the URL endpoint and use it.
 - If image contains no `/` construct the appropriate URL using the default project.
 - Otherwise the name is invalid.

platform/api/gcloud/compute.go
pbx0: Ah, right, this is wrong.

platform/api/gcloud/compute.go
pbx0: SGTM

pbx0:
@marineam PTAL. I moved the logic to a more appropriate place and implemented support for the full endpoint url, the shortened endpoint url starting with projects, and just the plain image name. Anything that doesn't start with those two given prefixes and contains "/" is rejected.

PR #430, opened 2016-12-20
kola/tests: fix a bunch of race conditions in files tests

dm0-:
The tests are checking if files exist with problematic modes, so we
don't care about them if they are deleted before stat is called.


mischief:
out of curiosity what files are affected by this?

dm0-:
The last failure I saw from this was `/etc/passwd.lock` and `/etc/gshadow.lock`, which was a result of the GCE agent adding/modifying accounts.

PR #431, opened 2016-12-21
kola/tests/docker: stop using kolet

dm0-:
This is what I used to test fixing GCE jobs.  I don't know if this is worth changing or not.  At least it saves 40 lines.


dm0-:
I'll close this since it's a part of #440.

PR #435, opened 2016-12-30
Import a stripped down version of the "testing" package as "harness"

marineam:
This package is not the final form I want it to be but it is far enough along to make sense as a stand-alone package while remaining as close to the original "testing" package as possible. That way we have a fairly well known starting point for reviewing all the changes/extensions/removals I intend to make.

Why do this in the first place? For starters, the "testing" package offers a nice model for reporting errors and logging that makes it clear where in the code and what test reported the message, even in the face of parallel tests. The current kola framework with its global logging and return-on-error lacks this ability. Also taking a step back and looking at the "testing" package offers an opportunity to redesign the kola framework to be more reusable. The downside is the "testing" package is explicitly designed to be driven by the "go test" command and nothing else which doesn't work well for producing a tool like the kola command. Forking the "testing" package also offers opportunities to experiment with and improve integration with contexts, logging, and more.


pbx0:
@marineam This LGTM so far. I think i'll be able to give more useful feedback once we start porting the rest of kola over to use this over the old harness.

marineam:
@pbx0 yeah, that's the idea. After this changes can be thought of as things we will do differently than the normal "testing" package.

PR #440, opened 2017-01-11
Fix kola on GCE ... mostly

dm0-:
For fun, here is a list of some of the race conditions I can recall offhand from running kola on GCE (in addition to the non-race problems).

  * All tests assume they can run immediately when SSH connects.  There are no service dependency relations between `sshd` and anything being tested.
  * Similarly, the cloud-config service has no dependencies with tested services, so even when SSH would connect and start services itself, the services' configuration files sometimes weren't written yet.
  * There were the usual `find` races, where files could sometimes be deleted between `readdir` and `stat`.  (This one was fixed separately in #430.)
  * The etcd commands will all happily run and fail before the cluster is healthy.  This includes all services acting as etcd clients.
  * The fleet daemon doesn't block or use etcd immediately, so `systemctl start fleet ; fleetctl status` will fail even if the etcd cluster is healthy.
  * When stopping sshd before locksmith reboots, the connection would still sometimes terminate too soon resulting in error.
  * For locksmith reboots, the tests would sometimes start checking for the newly booted system before it went down.
  * Network interfaces are created asynchronously, so they don't always exist between flannel running and the `ip` commands.

So...  There are still four failures that I can recall seeing that are not addressed here.  The two Kubernetes tests fail randomly, and this is likely due to the cloud-config races (and probably others) described above.  There was also a failure to reach the GCE metadata service, and I've seen flannel's vxlan test fail seemingly because an etcd system lost its connection to the cluster.  The first two need work on the Kubernetes setup, and the other two sound like they could have been problems with GCE itself.  I'll ignore those for the time being, in the interest of getting semi-functional GCE tests now.

### DIFF
--- a/kola/tests/locksmith/locksmith.go
+++ b/kola/tests/locksmith/locksmith.go
@@ -16,15 +16,14 @@ package locksmith
 
 import (
 	"fmt"
-	"time"
 
 	"golang.org/x/net/context"
 
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/kola/tests/etcd"
 	"github.com/coreos/mantle/lang/worker"
 	"github.com/coreos/mantle/platform"
-	"github.com/coreos/mantle/util"
 )
 
 func init() {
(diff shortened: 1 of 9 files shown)


marineam:
an alternative to adding sshd dependencies, perhaps the initial ssh check should poll systemd to report if the system has finished booting or not? I think I meant to do that at some point but didn't get around to it or it didn't interact well enough with cloud-config, don't recall. But that way the ignition config just needs to enable the service and the test starts when the system as a whole is in a consistent state, not just the few services added as a sshd dep.

kola/tests/locksmith/locksmith.go
```diff
@@ -83,13 +84,16 @@ func locksmithCluster(c cluster.TestCluster) error {
 	for _, m := range machs {
 		worker := func(c context.Context) error {
 			// XXX: stop sshd so checkmachine verifies correctly if reboot worked
-			// XXX: run locksmithctl under systemd-run so our current connection doesn't drop suddenly
-			cmd := "sudo systemctl stop sshd.socket; sudo systemd-run --quiet --on-active=2 --no-block locksmithctl send-need-reboot"
+			// XXX: run under systemd-run so our current connection doesn't drop suddenly
+			cmd := "sudo systemd-run --quiet --on-active=5 --no-block sh -exc 'systemctl stop sshd.socket && exec locksmithctl send-need-reboot'"
```
marineam: alternatively you can check for and ignore https://godoc.org/golang.org/x/crypto/ssh#ExitMissingError

dm0-:
I've replaced the service dependency dropins in each test in favor of waiting for systemd to declare it's fully booted.  It seems to work as well after a few test runs.  I haven't made the SSH change yet (and I'm not certain if that was the type of error that was causing the failure).

dm0-:
Oh, and I've also made `coreos-metadata.service` optional in the Ignition configs with `Wants=`, since the service failed to start when I tried it on QEMU, and kola supplies those variables for QEMU platform tests itself.

marineam:
That should be the correct error, the Reboot function used to have a similar systemd-run trick but I changed it to catch the error after the ssh library finally exported it.

dm0-:
Okay, I've changed it to use the same test.  Some Jenkins jobs are testing it now.

marineam:
LGTM

dm0-:
It looks like all three nodes spawned for that locksmith tests failed due to SSH errors in Jenkins.

    [0] no /etc/os-release file [1] no /etc/os-release file [2] ssh unreachable: dial tcp 104.154.208.61:22: getsockopt: connection refused

I'll try to figure out something else.

dm0-:
So it appears that the failure was just because SSH wasn't retrying when systemd reported `stopping` state.  I've added it to the acceptable states, and it passed.

PR #442, opened 2017-01-13
lang/maxint: define MaxInt and related constants

marineam:
Somehow `int` being implementation defined means you should blindly use
it without checking for overflow. I am very very confused by this.


marineam:
Closing, my code changed so I don't need this any more.

PR #443, opened 2017-01-17
New package for processing journals from remote sources

marineam:
This implements a parser for the journal export format used by journalctl and various journal remote tools:

https://www.freedesktop.org/wiki/Software/systemd/export/

The reason for implementing this is it will enable a higher level journal slurper to SSH to a remote host multiple times and each time resume reading the log where it left off because it'll have access to the __CURSOR field.

Since the export format is overly verbose for humans I've implemented support for writing logs in a format similar to journalctl's default output.

### DIFF
--- a/network/journal/entry.go
+++ b/network/journal/entry.go
@@ -1,0 +1,111 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


marineam:
@pbx0 right, and by using the cursor field you can restart the reader every time it fails for some reason, here is an example: https://github.com/marineam/mantle/blob/journal-test/network/journal/_test/main.go#L56-L79

My next step is to write something like that using SSH

pbx0:
Right on, this looks like a great start.

PR #446, opened 2017-01-23
Aws vendor bump

euank:
There's extra noise because I also used a newer godeps version.


euank:
😄 Jenkins green, woo!

PR #448, opened 2017-01-24
Disable test using ignition on qemu

marineam:
Many tests were converted to use ignition for configuration without implementing ignition support in the qemu platform code. Disable for now, I'm in the middle of cleaning up the qemu code and will add ignition support in the process.


marineam:
The jenkins tests are likely using the tagged kola version in the SDK, not directly off of master.

PR #451, opened 2017-01-25
Fix the `--debug` flag for the `ore` command

euank:
Per the commit descriptions.

Before this change:
```
$ ./bin/ore --debug version    
mantle/ore version 0.0.0
```
With this change:
```
$ ./bin/ore --debug version
2017-01-25T23:14:14Z cli: Started logging at level DEBUG
mantle/ore version 0.0.0
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 *.sw?
-bin/
-
+/bin/
+/gopath/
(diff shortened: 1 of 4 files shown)


euank:
@marineam open to better ideas!

Only other idea I've got is going upstream and adding `PersistentPreRunButReallyPersist` or `CascadingPreRun` or something.

marineam:
More of I'm not a fan of the PreRun mechanism to begin with.

PR #457, opened 2017-02-01
kola/docker: add regression test for 1785

euank:
This test fairly reliably fails for 1235-8 and has passed 2/2 times for 1235-6.

It takes a little under 2 minutes to run, so not too bad.

cc @dm0-


mischief:
test names are fairly arbitrary, except for the dots. you can do kola run docker.* to glob tests in 'groups'.

PR #464, opened 2017-02-15
Add support for dumping machine user-data and journal to GCE

marineam:
One slight catch about this scheme that I've found: if you are switching between platforms between runs of kola you'll need to manually remove the _kola_temp directory after QEMU runs since they run as root while GCE and others can be run as your normal user.

### DIFF
--- a/cmd/kola/bootchart.go
+++ b/cmd/kola/bootchart.go
@@ -66,9 +66,9 @@ func runBootchart(cmd *cobra.Command, args []string) {
 	if kolaPlatform == "qemu" {
 		cluster, err = qemu.NewCluster(&kola.QEMUOptions, outputDir)
 	} else if kolaPlatform == "gce" {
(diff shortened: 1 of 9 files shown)


marineam:
Note that although the output dir parameter is now being passed to AWS clusters it isn't implemented there yet, that'll be one more step.

marineam:
Minor update to fix machine.Destroy, the journal attribute is only set half-way through the setup so it may be nil if an error occurs early on in setup.

```diff
--- a/platform/machine/gcloud/machine.go
+++ b/platform/machine/gcloud/machine.go
@@ -61,8 +61,10 @@ func (gm *machine) Destroy() error {
                return err
        }
 
-       if err := gm.journal.Destroy(); err != nil {
-               return err
+       if gm.journal != nil {
+               if err := gm.journal.Destroy(); err != nil {
+                       return err
+               }
        }
 
        gm.gc.DelMach(gm)
```

PR #466, opened 2017-02-15
Some QEMU arm64 fixes

dm0-:
Here are some non-timeout-related fixes from my ARM testing branch.

Half of them are re-enabling the QEMU tests to run those again since Ignition is supported now.  (They could probably go in a different PR.)  Perhaps coreos-metadata should be modified to exit successfully with a "Nothing to do" message on QEMU instead of having the last commit, but this will at least get everything working again.


marineam:
The behavior of coreos-metadata.service certainly needs to be fixed. Perhaps file a bug and add a TODO comment with the link the applicable files?

kola/tests/docker/docker.go
```diff
@@ -254,7 +256,12 @@ func dockerNetwork(c cluster.TestCluster) error {
 // Regression test for https://github.com/coreos/bugs/issues/1569 and
 // https://github.com/coreos/docker/pull/31
 func dockerOldClient(c cluster.TestCluster) error {
-	oldclient := "/usr/lib/kola/amd64/docker-1.9.1"
+	arch := "amd64"
+	if _, ok := c.Cluster.(*qemu.Cluster); ok && kola.QEMUOptions.Board != "" {
+		arch = strings.SplitN(kola.QEMUOptions.Board, "-", 2)[0]
+	}
+
+	oldclient := fmt.Sprintf("/usr/lib/kola/%s/docker-1.9.1", arch)
```
marineam: Is there a docker binary for arm that works for this test? iirc I don't think there is and this test was left like this with the intent to add a arch option to tests similar to the existing platform but that never happened. I guess we never thought of doing the type assertion to detect arch instead.

Since there won't ever be a non-amd64 binary for this test perhaps instead keep the type assertion but use it to do Skip("only applicable to amd64") or similar just for the sake of clarity.

dm0-: Okay, updated.  The oldest CoreOS alpha image for arm64 has Docker 1.10.3, so at least we don't have that binary version.

dm0-:
I've opened coreos/bugs#1815 and added a link to it for each affected Ignition file.

PR #472, opened 2017-02-17
kola/tests/misc: add a test to find broken symbolic links

dm0-:
This is a follow-up to coreos/bugs#1807.  It checks `/` and `/usr`.  It might be worth extending to some of the tmpfs file systems to verify all tmpfiles.d settings, but then more whitelists would need to be added since there are packages intentionally creating broken links.

### DIFF
--- a/kola/tests/misc/files.go
+++ b/kola/tests/misc/files.go
@@ -25,6 +25,12 @@ import (
 
 func init() {
 	register.Register(&register.Test{
+		Run:         DeadLinks,
+		ClusterSize: 1,
+		Name:        "coreos.filesystem.deadlinks",
+		UserData:    `#cloud-config`,
+	})
+	register.Register(&register.Test{
 		Run:         SUIDFiles,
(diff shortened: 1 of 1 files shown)


kola/tests/misc/files.go
```diff
@@ -85,6 +91,22 @@ func sugidFiles(m platform.Machine, validfiles []string, mode string) error {
 	return nil
 }
 
+func DeadLinks(c cluster.TestCluster) error {
+	m := c.Machines()[0]
+
+	/* Accept return code 1 for link loop warnings.  */
+	output, err := m.SSH(`sudo find -L / /usr -ignore_readdir_race -xdev -type l 2>/dev/null || [ "x$?" = x1 ]`)
```
euank: We don't need the ` = x1` bit or the stderr redirect if we use `find ... -xtype l` instead.

dm0-: Okay, updated.

PR #473, opened 2017-02-19
More harness refactoring

marineam:
The two main big things here:
 - Moving code around between files to organize it a little better.
 - Pass options as a struct instead of command line flags (though those can still be used via a FlagSet).

Minor things:
 - Update license headers.
 - Properly stop profiling/close files when test run finishes.
 - Remove some unneeded re-implementations.


harness/suite.go
```diff
@@ -128,107 +128,93 @@ func (s *Suite) Run() error {
...
+		f.Close()
 	}
-	return nil
-}
 
-// before runs before all testing.
-func (m *Suite) before() {
-	if *memProfileRate > 0 {
-		runtime.MemProfileRate = *memProfileRate
+	if *memProfile != "" && *memProfileRate > 0 {
```
pbx0: Is all this for pprof profiling of kola code?

marineam: Yes

PR #474, opened 2017-02-20
kola/tests/misc: ignore dead links in certain paths

dm0-:
This test made kola fail on GCE due to the rkt agent writing dead links in `/var`.  Expand the test coverage to all files, and add an ignore list.

### DIFF
--- a/kola/tests/misc/files.go
+++ b/kola/tests/misc/files.go
@@ -94,9 +94,20 @@ func sugidFiles(m platform.Machine, validfiles []string, mode string) error {
 func DeadLinks(c cluster.TestCluster) error {
 	m := c.Machines()[0]
 
-	output, err := m.SSH(`sudo find / /usr -ignore_readdir_race -xdev -xtype l`)
+	ignore := []string{
+		"/dev",
+		"/proc",
+		"/run/udev/watch",
+		"/sys",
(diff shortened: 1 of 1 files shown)


kola/tests/misc/files.go
```diff
@@ -94,9 +94,19 @@ func sugidFiles(m platform.Machine, validfiles []string, mode string) error {
 func DeadLinks(c cluster.TestCluster) error {
 	m := c.Machines()[0]
 
-	output, err := m.SSH(`sudo find / /usr -ignore_readdir_race -xdev -xtype l`)
+	ignore := []string{
+		"/dev",
+		"/proc",
+		"/run/udev/watch",
+		"/sys",
+		"/var/lib/rkt",
```
marineam: I presume /var/lib/docker should be excluded as well for the same reason as rkt. And why do /dev /proc /sys and /run need to be there if you are using -xdev?

marineam: ohhhh, you dropped -xdev

dm0-: Okay, I've added Docker to the list.

It was using `-xdev` to avoid the bad links in `/proc` and `/run`, but I removed it in favor of excluding the paths.

PR #475, opened 2017-02-22
enhancements

glevand:
Some enhancements.

With the patch `Jenkinsfile: Restrict to sdk node` you will need to add the `sdk` label to your jenkins nodes.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,7 @@ properties([
     pipelineTriggers([pollSCM('H/15 * * * *')])
 ])
 
-node('docker') {
+node('amd64 && docker') {
     stage('SCM') {
         checkout scm
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,9 +3,6 @@
 properties([
     buildDiscarder(logRotator(daysToKeepStr: '20', numToKeepStr: '30')),
 
-    [$class: 'GithubProjectProperty',
-     projectUrlStr: 'https://github.com/coreos/mantle'],
-
     [$class: 'CopyArtifactPermissionProperty',
      projectNames: '*'],
 
--- a/cmd/cork/downloadimage.go
+++ b/cmd/cork/downloadimage.go
@@ -58,12 +58,12 @@ func init() {
 	downloadImageCmd.Flags().BoolVar(&downloadImageVerify,
 		"verify", true, "verify")
 	downloadImageCmd.Flags().Var(&downloadImagePlatformList,
-		"platform", "Choose qemu, gce, or aws. Multiple platforms can be specified by repeating the flag")
+		"platform", "Choose qemu, qemu_uefi, gce, or aws. Multiple platforms can be specified by repeating the flag")
 
 	root.AddCommand(downloadImageCmd)
 }
 
-type platformList []string // satisfies pflag.Value interface
+type platformList [][]string // satisfies pflag.Value interface
 
 func (platforms *platformList) String() string {
 	return fmt.Sprintf("%v", *platforms)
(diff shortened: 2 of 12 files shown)


glevand:
Rebased to latest.

sdk/verify.go
```diff
@@ -278,7 +281,18 @@ func VerifyFile(file string) error {
...
 	defer signature.Close()
 
-	if err := Verify(signed, signature); err != nil {
+	var key string
+	if verifyKeyFile == "" {
+		key = buildbot_coreos_PubKey
+	} else {
+		b, err := ioutil.ReadFile(verifyKeyFile)
+		if err != nil {
+			plog.Fatal(err)
```
marineam: In packages outside of main please return errors instead of calling Fatal.

sdk/download.go
```diff
@@ -72,7 +72,7 @@ func DownloadFile(file, fileURL string, client *http.Client) error {
 		path := strings.TrimLeft(parseURL.Path, "/")
 		obj, err := api.Objects.Get(parseURL.Host, path).Do()
 		if err != nil {
-			plog.Fatal(err)
+			plog.Fatal(fmt.Errorf("%s: %s", err, fileURL))
```
marineam: While we are here lets switch to returning the error instead.

cmd/cork/downloadimage.go
```diff
@@ -75,21 +78,22 @@ func (platforms *platformList) Type() string {
...
 
-	// maps names of platforms to filename prefix
-	platformMap := map[string]string{
-		"qemu": "_image.bin.bz2",
-		"gce":  "_gce.tar.gz",
-		"aws":  "_ami_image.bin.bz2",
+	// Maps names of platforms to a list of file suffixes to download.
+	platformMap := map[string][]string{
+		"qemu":      {"_image.bin.bz2"},
+		"qemu-uefi": {"_qemu_uefi_efi_code.fd", "_qemu_uefi_efi_vars.fd", "_image.bin.bz2"},
```
marineam: lets do `"qemu_uefi"` for the sake of consistency.

Jenkinsfile
```diff
@@ -3,9 +3,6 @@
 properties([
     buildDiscarder(logRotator(daysToKeepStr: '20', numToKeepStr: '30')),
 
-    [$class: 'GithubProjectProperty',
-     projectUrlStr: 'https://github.com/coreos/mantle'],
```
dm0-: This property is only here because it was carried over from the previous Jenkins implementation, and I don't really care if we remove it, but is this causing problems?  This is only used for cosmetic enhancements such as adding the project side-bar link on the job and some changelog links (as far as I know).

Jenkinsfile
```diff
@@ -18,7 +15,7 @@ properties([
     pipelineTriggers([pollSCM('H/15 * * * *')])
 ])
 
-node('docker') {
+node('sdk && docker') {
```
dm0-: This seems wrong.  The SDK has nothing to do with this build, since it's just starting a Docker container.  (In fact, it can't be built in the SDK, since the SDK doesn't include Docker.)  This change should be dropped.

Jenkinsfile
glevand: I didn't notice any problems, just thought I'd clean it up.

Jenkinsfile
glevand: We need some way to keep this build off arm64 nodes.  I already use the label `sdk` in the other  jenkins jobs https://github.com/coreos/jenkins-os/pull/1, so used it here.  Do you have an idea you think better?

dm0-: Okay, we're already using Gentoo-style architecture labels, so `amd64 && docker` would work here for us.  I also think that label would better reflect the problem that is trying to be solved.

I will comment similarly on that other PR.

glevand:
Made changes @marineam requested.  Changed commit log of `Jenkinsfile: Restrict to sdk node`.

glevand:
Changed to `Jenkinsfile: Restrict to amd64 node`.

PR #477, opened 2017-02-22
kola: begin migration to new harness test framework

marineam:
Two main behavior changes here: text output changes some, looking more like `go test` output. The tap output no longer contains test numbers but in theory those are optional.


harness/suite.go
```diff
@@ -280,3 +294,43 @@ func (s *Suite) runTests(w io.Writer) error {
 func (s *Suite) outputPath(path string) string {
 	return filepath.Join(s.opts.OutputDir, path)
 }
+
+// cleanOutputDir creates/empties Options.OutputDir.
+// If the path already exists it must be named similar to `_foo_temp`
+// or contain `.harness_temp` to indicate removal is safe; we don't
+// want users wrecking things by accident with `-outputdir /tmp`
+func (s *Suite) cleanOutputDir() error {
```
pbx0: Not sure if this is the right place, but it might be worth surfacing an error that, based on the user given name, the output dir wasn't safe to remove.

marineam: As is it'll report an EEXIST error from Mkdir in that case. something a bit more specific seems reasonable.

marineam: Made a better error message in 1f7e7a5

PR #478, opened 2017-02-23
kola/docker: add regression test for CVE-2016-8867

euank:
Tested locally against the current stable ami and verified it passed (specifically: `AWS_REGION=us-west-2 ./bin/kola run --platform aws  --aws-ami ami-444dcd24 --aws-type t2.small --basename "euank-kola-test" "docker.user-no*"`)


euank:
Dropped ls's output, on success it did spit out a permission denied error.

PR #480, opened 2017-02-24
kola/tests/ignition: add test for users and groups

crawford:


kola/tests/ignition/v2/passwd.go
```diff
@@ -0,0 +1,160 @@
...
+		Name:        "coreos.ignition.v2.users",
+		Run:         users,
+		ClusterSize: 1,
+		UserData: `{
+		             "ignition": { "version": "2.0.0" },
+		             "passwd": {
+		               "users": [
+		                 {
+		                   "name": "core",
+		                   "passwordHash": "$1$xG18DXSV$8DSq5bJfyy60ZBnYJflbK0"
```
marineam: Considering this test doesn't need a working password how about use an obviously invalid hash so the machine that spins up doesn't have a publicly known credential.

PR #485, opened 2017-02-28
harness: add an example for how to customize test suites

marineam:
Slowly porting kola to the new harness package is a little tricky so I
wanted a clear/concise example of how harness can be used to illustrate
what I'm trying to do.


marineam:
Renamed the package to _example because the test script tries to run --help on all main packages but only the pflag/cobra based commands exit 0, the example which uses the normal flag parsing exists non-zero.

Not super important to compile/run the example in testing so just hide the package.

pbx0:
Renaming to _example sgtm.

PR #490, opened 2017-03-01
pluton: support NodeCheck becoming more flexible

pbx0:
In the future I'd like to this function to become more robust and based
on how the Cluster was setup and what things are expected to be there.
This just lays the groundwork interfacewise so future tests rely on it.
I also hardcode the timeout value to the highest one in use. I don't
think we need to expose it to test writers.


pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

PR #491, opened 2017-03-01
pluton: update readme with examples and more about roadmap

pbx0:


marineam:
Now that this has merged could this move to the main README? Or is there a particular reason to keep it on its own?

pbx0:
@marineam: I don't have strong opinions on the matter.  I figured there would be people looking to use pluton soon that have no idea what it is or what kola is and they might be less confused if they land on the pluton specific readme.

PR #496, opened 2017-03-04
kola/misc: add toolbox 'dnf install' test

euank:
This is a regression test for the issue where toolbox's fedora updated and
caused breakages.
This is also just a test that the toolbox works at all really.

This doesn't test the root-cause, and a test for that should
be added as well.

### DIFF
--- a/kola/tests/misc/toolbox.go
+++ b/kola/tests/misc/toolbox.go
@@ -1,0 +1,47 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


kola/tests/misc/toolbox.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// regression test for https://github.com/coreos/bugs/issues/1676
+func sigKill(c cluster.TestCluster) error {
+	m := c.Machines()[0]
+
+	output, err := m.SSH(`toolbox sh -c 'dnf install -y tcpdump; tcpdump --version >/dev/null && echo PASS' 2>/dev/null`)
+
+	if err != nil {
+		return fmt.Errorf("Error sig-killing in a toolbox: %v", err)
```
marineam: This error message reports a sig-kill regardless of the actual error.

kola/tests/misc/toolbox.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         sigKill,
+		ClusterSize: 1,
+		Platforms:   []string{"aws", "gce"}, // Network access for toolbox
```
dm0-: Use `ExcludePlatforms: []string{"qemu"}` here to allow for future `azure`, `packet`, whatever.

euank:
Rebased, addressed comments

PR #507, opened 2017-03-14
kola/docker: test docker skim

euank:
This PR reworks the docker tests in kola such that any test may be
targetted at a combination of target configurations.

The current configurations are 'host' and 'skim-1.12.6', but it should
be easy to add more.

This effectively doubles the number of docker tests and thus does add
not-insignificant runtime.

### DIFF
--- a/kola/tests/docker/docker_util.go
+++ b/kola/tests/docker/docker_util.go
@@ -1,0 +1,167 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker
+
+import (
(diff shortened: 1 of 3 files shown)


kola/tests/docker/docker_util.go
```diff
@@ -0,0 +1,167 @@
...
+Documentation=http://docs.docker.com
+After=docker.socket
+Requires=docker.socket
+
+[Service]
+Type=simple
+
+ExecStart=/usr/bin/rkt run --dns=host --interactive \
+  --trust-keys-from-https \
+  --stage1-name=users.developer.core-os.net/skim/stage1-skim:0.0.1 \
```
marineam: Is this going to get a more permanent home on Quay? Or do we need to create something? `users.developer` should just be temporary.

euank: (sorry for the incredibly slow response)

The stage1 can't live on quay iiuc because it needs specific aci annotations to behave as a stage1.

It's intentional that it lives in an obviously non-production environment because it's not a recommended thing.

It might be better to live elsewhere though as long as it's still clear it's not a production supported thing.

marineam: In the past I've posted source tarballs under `gs://coreos-net-storage/mirror/snapshots/` (backend for `https://storage.core-os.net/mirror/snapshots/`). So some path under the coreos-net-storage may be appropriate. And given the --trust-keys-from-https that seems better too since the ssl connection between CF and Google is strictly enforced for that bucket while it is not for the users bucket (we would have had to rename the users bucket to have robust ssl)

euank:
Closing as quite stale at this point.

PR #508, opened 2017-03-14
pluton/tests/bootkube: scale test is too flakey for main test namespace

pbx0:
While we are here, tweak error handling a little.

cc @marineam


pbx0:
coreosbot run pluton tests

PR #510, opened 2017-03-14
pluton: move self-hosted destructive tests into main CI

pbx0:
cc @hongchaodeng


pbx0:
coreosbot run pluton tests

PR #511, opened 2017-03-14
pluton/spawn: don't use rkt-fly on bootkube start

pbx0:
This was causing my bootkube PR for 1.6 updates to fail. Its unclear why, but removing rkt-fly stopped bootkube from nearly immediately exiting. 

It was one of the last remaining differences between how I setup bootkube and the scripts from the in-repo hack-dir. I added the rkt-fly so I could separate stdout from stdin in rkt, there is probably another way to accomplish this.


pbx0:
coreosbot run pluton tests

PR #514, opened 2017-03-17
kola/tests/misc: test the coreos-install cloud-config path

dm0-:
With systemd v233, the default.target.wants links are not used.  It only affected cloudinit, and it went unnoticed since this specific activation path is not used by any test.  Therefore, add a test to make sure /var/lib/coreos-install/user_data actually takes effect.

This tests for coreos/bugs#1872.

### DIFF
--- a/kola/tests/misc/install.go
+++ b/kola/tests/misc/install.go
@@ -1,0 +1,56 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


kola/tests/misc/install.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// Simulate coreos-install features
+
+// Verify that the coreos-install cloud-config path is used
+func InstallCloudConfig(c cluster.TestCluster) error {
+	m := c.Machines()[0]
+
+	// Verify the host name was set from the cloud-config file
+	if output, err := m.SSH("hostname"); err != nil || !bytes.Equal(output, []byte("cloud-config-worked")) {
```
marineam: This should be retried a few times since there is no strict dependency between cloudinit and ssh.

dm0-: Is that still an issue since 32d2ff17e2690c2d0dcc27122459f612dd315d4b?  Shouldn't the test execution wait until all units are running?

marineam: Ah, right. NM then

PR #518, opened 2017-03-18
pluton: use new harness directly

pbx0:
Ready for review! New harness simplifies things so much for pluton considering that I no longer import anything from `kola/`.

Since we generally only run on GCE for bootkube testing I'm punting on support for other cloud platforms for now. It should take very little modification to pipe through something like all the AWS options and have things work, but, I'm mainly unsure of how I want to deal with the notion of a platform from a `pluton.Test` perspective.

### DIFF
--- a/cmd/pluton/pluton.go
+++ b/cmd/pluton/pluton.go
@@ -21,8 +21,7 @@ import (
 	"text/tabwriter"
 
 	"github.com/coreos/mantle/cli"
-	"github.com/coreos/mantle/kola"
-	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/pluton/harness"
 	_ "github.com/coreos/mantle/pluton/tests/bootkube"
 
 	"github.com/coreos/pkg/capnslog"
@@ -39,11 +38,10 @@ var (
 	}
 
 	cmdRun = &cobra.Command{
-		Use:    "run [glob pattern]",
-		Short:  "Run run pluton tests by category",
-		Long:   "run all pluton tests (default) or related groups",
-		Run:    runRun,
-		PreRun: preRun,
+		Use:   "run [glob pattern]",
+		Short: "Run run pluton tests by category",
+		Long:  "run all pluton tests (default) or related groups",
+		Run:   runRun,
 	}
 
 	cmdList = &cobra.Command{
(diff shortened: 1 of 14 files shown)


cmd/pluton/pluton.go
```diff
@@ -21,9 +21,8 @@ import (
 	"text/tabwriter"
 
 	"github.com/coreos/mantle/cli"
-	"github.com/coreos/mantle/kola"
-	"github.com/coreos/mantle/kola/register"
-	_ "github.com/coreos/mantle/pluton/tests/bootkube"
+	"github.com/coreos/mantle/pluton/harness"
+	//_ "github.com/coreos/mantle/pluton/tests/bootkube"
```
marineam: delete

marineam:
Looking good so far

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
@marineam PTAL

pbx0:
coreosbot run pluton tests

pluton/harness/run.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	return filteredTests, nil
+}
+
+// RunTest is called inside the closure passed into the harness. Currently only
+// GCE is supported, no reason this can't change
+func runTest(t pluton.Test, h *harness.H) {
+	h.Parallel()
+
+	testDir := filepath.Join(Opts.OutputDir, t.Name)
```
marineam: This can be replaced by `testDir := h.OutputDir()` now.

pbx0:
coreosbot run pluton tests

pluton/cluster.go
```diff
@@ -43,6 +47,7 @@ type Info struct {
 	KubeletTag      string // e.g. v1.5.3_coreos.0
 	Version         string // e.g. v1.5.3+coreos.0
 	UpstreamVersion string // e.g. v1.5.3
+	Options                // pipe through how test was configured
```
marineam: I don't see any sign of this actually being used. So you could just drop this and move the Options attributes to the Test struct and simplify declarations.

pbx0: One use I had of Options being separate from a test struct was that one could do for most tests:
```
var defaults = pluton.Options{
        SelfHostEtcd:    false,
        InitialMasters: 1,
        InitialWorkers:  1,
}

func init() {
        // main test suite run on every PR
        harness.Register(pluton.Test{
                Name:    "bootkube.smoke",
                Run:     smoke,
                Options: defaults,
        })
        harness.Register(pluton.Test{
                Name: "bootkube.selfetcd.smoke",
                Run:  smoke,
                Options: pluton.Options{
                        SelfHostEtcd:   true,
                        InitialMasters: 1,
                        InitialWorkers: 1,
                },
        })
}
```
Its not perfect, but its an easy way for the owner of the test repo to declare what their normal cluster setup looks like. Anything not default has to specify ALL options still of course.

I will drop piping through Options into the Info struct at least though since it has no uses yet and I don't remember any good use-cases I had in mind at the time.

pluton/harness/run.go
pbx0: done.

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pluton/cluster.go
marineam: Personally I think `Options: pluton.Options{` when the defaults aren't applicable seems more awkward than not being able to use defaults.

You can leave it as-is for now to see if that sort of defaults usage proves valuable or not but unless Options grows in complexity I suspect you may be best served just forcing InitialMasters/Workers to `1` in the setup code if they are left as zero since largely that's the value most tests are using anyway.

PR #522, opened 2017-03-21
qemu: Avoid copying disk image

bgilbert:
Use the base image as COW backing store for a qcow2 disk image. This saves 750 MB of I/O per instance, avoids thrashing the page cache, and reduces SSD wear. On my SSD-backed system, starting from a cold cache, it reduces the `kola run` runtime from 840s to 595s.

### DIFF
--- a/platform/machine/qemu/cluster.go
+++ b/platform/machine/qemu/cluster.go
@@ -161,7 +161,7 @@ func (qc *Cluster) NewMachine(cfg string) (platform.Machine, error) {
 		"-uuid", qm.id,
 		"-display", "none",
 		"-add-fd", "fd=4,set=1",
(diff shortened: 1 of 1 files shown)


bgilbert:
On my machine I'm seeing a slight advantage to qcow (614s averaged over three runs) over cp (668s) even when a tmpfs is mounted on the SDK's /tmp. I also tried loopback-mounting an XFS filesystem with reflink support enabled onto the SDK /tmp and putting the source image in it; again, slight advantage to qcow (639s averaged over two runs) over reflinks (666s). Re reflinks, I'd also argue that we shouldn't require a somewhat unusual system configuration to get reasonable performance.

marineam:
re reflink I was mostly curious. I'll do a little testing to compare to your results.

PR #526, opened 2017-03-29
tests/docker/userns: re-enable for aws

euank:
With our current alpha kernel it passes. See
https://github.com/coreos/bugs/issues/1826

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -95,7 +95,6 @@ func init() {
 			  - name: dockremap
 			    create: {}
 		*/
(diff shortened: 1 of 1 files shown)


kola/tests/docker/docker.go
```diff
@@ -95,7 +95,7 @@ func init() {
 			  - name: dockremap
 			    create: {}
 		*/
-		Platforms:  []string{"qemu", "gce"}, // aws: https://github.com/coreos/bugs/issues/1826
+		Platforms:  []string{"qemu", "gce", "aws"},
```
marineam: When all platforms are working you can just omit the platforms line entirely.

euank:
Done

PR #527, opened 2017-03-31
Include a command to enable using signed binary packages

dm0-:


cmd/gangue/gangue.go
```diff
@@ -0,0 +1,139 @@
...
+		Short: "Google Storage download and verification tool",
+	}
+
+	get = &cobra.Command{
+		Use:   "get [url] [path]",
+		Short: "download and verify a file from Google Storage",
+		Run:   run,
+	}
+
+	gpgKeyFile, jsonKeyFile, output, source string
```
marineam: source/output don't need to be global

cmd/gangue/gangue.go
```diff
@@ -0,0 +1,139 @@
...
+
+func run(cmd *cobra.Command, args []string) {
+	var client *http.Client
+
+	if len(args) == 2 {
+		source = args[0]
+		output = args[1]
+	} else if len(args) == 1 {
+		source = args[0]
+		output, _ = os.Getwd()
```
marineam: Just use "." if you want to not worry about error handling.

cmd/gangue/gangue.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func main() {
+	bv := get.PersistentFlags().BoolVar
+	sv := get.PersistentFlags().StringVar
+
+	bv(&serviceAuth, "service-auth", false, "use non-interactive auth when running within GCE")
+	sv(&jsonKeyFile, "json-key", "", "use a service account's JSON key for authentication")
+	bv(&verify, "verify", true, "use GPG verification")
+	sv(&gpgKeyFile, "verify-key", "", "PGP public key file to verify signatures")
```
marineam: Perhaps explicitly state that this is an alternative key, since our default image signing key will be used if blank.

cmd/gangue/gangue.go
```diff
@@ -0,0 +1,139 @@
...
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(1)
+	}
+}
+
+func main() {
+	bv := get.PersistentFlags().BoolVar
+	sv := get.PersistentFlags().StringVar
+
+	bv(&serviceAuth, "service-auth", false, "use non-interactive auth when running within GCE")
```
marineam: we usually put these into `init()`

dm0-:
Okay, I've updated it.  (A lot of those variables were globals when I was trying to figure out how to make Cobra take positional arguments, but it looked like they still haven't merged support for that on the root command.)

cmd/gangue/gangue.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	if parsedURL.Path[len(parsedURL.Path)-1] == '/' {
+		return fmt.Errorf("URL must not be a directory path %v", rawURL)
+	}
+	return nil
+}
+
+func run(cmd *cobra.Command, args []string) {
+	var client *http.Client
+	output, source := "", ""
```
marineam: var syntax seems more fitting here

cmd/gangue/gangue.go
```diff
@@ -0,0 +1,142 @@
...
+		source = args[0]
+		output = "."
+	} else {
+		fmt.Fprintf(os.Stderr, "Expected one or two arguments\n")
+		os.Exit(1)
+	}
+
+	// Perform some basic sanity checks on the options
+	if output == "" {
+		fmt.Fprintf(os.Stderr, "The output file path is required\n")
```
marineam: This seems confusing when it is actually optional, perhaps just convert a blank arg to "."

dm0-:
Okay, updated.

cmd/gangue/gangue.go
```diff
@@ -0,0 +1,141 @@
...
+		err = sdk.UpdateFile(output, source, client)
+	}
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(1)
+	}
+}
+
+func init() {
+	get := &cobra.Command{
```
marineam: init() is normally the top function, right below the var () block, and the get command can be defined in the var block along with root for the sake of consistency.

After that lgtm.

dm0-: Okay, done.

PR #528, opened 2017-04-01
Make gangue downloads cleaner

dm0-:
After building with signed bootstrap packages, some new issues arose because the toolchain packages are downloaded into the same directory as built packages.

- The `sdk.UpdateSignedFile` function leaves detached signatures in place, which may be desired in most cases, but since the scripts upload packages by the directory, they would sign the leftover signatures while uploading them.

- The `sdk.UpdateFile` function doesn't respect `umask` since it creates the file with `ioutil.TempFile`, which uses mode `0600`.  This can cause read failures when `emerge` downloads the binary packages as root, then other parts of the SDK scripts try to read them as an unprivileged user.  The change here just uses `ioutil.TempFile` to generate the file name and lets the download function write the file, but this will affect all commands using `sdk.UpdateFile`, so I'm not sure about it.

### DIFF
--- a/sdk/download.go
+++ b/sdk/download.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
@@ -263,17 +262,11 @@ func UpdateFile(file, url string, client *http.Client) error {
 		return err
 	}
 
-	t, err := ioutil.TempFile(filepath.Dir(file), "sdkUpdateCheck")
-	if err != nil {
-		return err
-	}
-	t.Close()
-	tempFile := t.Name()
-	defer os.Remove(tempFile)
-
+	tempFile := file + ".part"
 	if err := DownloadFile(tempFile, url, client); err != nil {
 		return fmt.Errorf("%s: %s", url, err)
 	}
+	defer os.Remove(tempFile)
 
 	equal, err := cmpFileBytes(file, tempFile)
 	if os.IsExist(err) { // file may not exist, that is ok
--- a/cmd/gangue/gangue.go
+++ b/cmd/gangue/gangue.go
@@ -41,8 +41,8 @@ var (
 		Run:   run,
 	}
 
-	gpgKeyFile, jsonKeyFile string
-	serviceAuth, verify     bool
+	gpgKeyFile, jsonKeyFile      string
+	keepSig, serviceAuth, verify bool
 )
 
 func init() {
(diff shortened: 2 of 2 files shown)


sdk/download.go
```diff
@@ -269,11 +269,12 @@ func UpdateFile(file, url string, client *http.Client) error {
 	}
 	t.Close()
 	tempFile := t.Name()
-	defer os.Remove(tempFile)
+	os.Remove(tempFile) // let DownloadFile create it and use umask
```
euank: Perhaps DownloadFile should be able to create a tempfile and return it instead. This feels hacky.

dm0-: The use of tempfiles is what causes the problem in the first place, so I don't see why more should be added.

If unlinking the file is a problem, maybe it could be solved by a manual calculation of an appropriate mode with umask and `chmod`ing the tmpfile instead.

euank: We don't have to use go's TempFile though; by creating and removing a tempfile here, we're already giving up its logic for ensuring the file doesn't exist. At this point, we might as well be generating a random string instead and using it.

marineam: Yeah, this completely defeats the use of TempFile. Since I don't think there is any security issue here strict use of TempFile isn't required so there are three options:

 - Always download to some pre-defined but alternate name, just so nothing confuses incomplete files with valid ones. e.g. use `os.Create(file + ".part")`
 - Use some random string, effectively what this this change does but without the pretense of using TempFile
 - Use chmod.

I've got a feeling I've probably done this TempFile+Remove thing before but per-euank's point lets avoid that pattern. :)

dm0-: I've just changed it to set the mode to something useful.  It's still kind of ugly since I don't see a way to get the umask without changing it.  If that's an issue, I can just move the mode change to the command instead of in the `sdk` library so nothing else gets affected by this.

dm0-: @marineam Sorry, I just pushed a `chmod` version and commented before your commend showed up.

sdk/download.go
```diff
@@ -271,6 +272,11 @@ func UpdateFile(file, url string, client *http.Client) error {
 	tempFile := t.Name()
 	defer os.Remove(tempFile)
 
+	// base the file's mode on umask instead of just 0600
+	umask := syscall.Umask(0777)
```
marineam: Perhaps this step should be done in `init()` so we aren't changing process-wide state when threads may be involved.

dm0-:
It's updated to drop `ioutil.TempFile`.

PR #529, opened 2017-04-03
ore aws enhancements

bgilbert:
- Merge `ore aws create-snapshot` and `ore aws create-images` into `ore aws upload`
- Reuse previous snapshot import task or snapshot if available
- Add `ore aws copy-image` for replicating an AMI across regions
- Enable ENA and Intel 82599 VF networking on HVM images
- Cleanups


cmd/ore/aws/upload.go
```diff
@@ -64,92 +85,145 @@ func defaultUploadFile() string {
...
+func defaultBucketURL(urlPrefix, imageName, board, file, region string) (*url.URL, error) {
+	if urlPrefix == "" {
+		urlPrefix = fmt.Sprintf("s3://%s", defaultBucketNameForRegion(region))
 	}
 
-	s3URL, err := url.Parse(s3URI)
+	s3URL, err := url.Parse(urlPrefix)
 	if err != nil {
-		return "", err
+		return &url.URL{}, err
```
marineam: Just return `nil`

platform/api/aws/images.go
```diff
@@ -289,3 +365,147 @@ func registerImageParams(snapshotID, name, description string, diskBaseName stri
...
+
+	var imageID string
+	describeRes, err := a.ec2.DescribeImages(&ec2.DescribeImagesInput{
+		Filters: []*ec2.Filter{
+			&ec2.Filter{
+				Name:   aws.String("name"),
+				Values: aws.StringSlice([]string{name}),
+			},
+		},
+		Owners: aws.StringSlice([]string{"self"}),
```
euank: TIL about self, cool!

platform/api/aws/images.go
```diff
@@ -289,3 +365,147 @@ func registerImageParams(snapshotID, name, description string, diskBaseName stri
...
+			amis[result.region] = result.imageID
+		}
+		if err == nil {
+			err = result.err
+		}
+	}
+	return amis, err
+}
+
+func (a *API) copyImageIn(sourceRegion, sourceImageID, name, description string, imageTags, snapshotTags []*ec2.Tag, ch chan<- copyImageResult) {
```
euank: I think it's cleaner to just return `copyImageResult` here (since it will only have one result), and the result can be put on a channel by the caller.

platform/api/aws/images.go
```diff
@@ -100,23 +100,86 @@ func (a *API) CreateSnapshot(description, sourceURL string, format EC2ImageForma
...
+		case "cancelled", "cancelling", "deleted", "deleting":
+			continue
+		case "completed":
+			// Either we lost the race with a snapshot that just
+			// completed or this is an old import task for a
+			// snapshot that's been deleted. Check it.
+			_, err := a.ec2.DescribeSnapshots(&ec2.DescribeSnapshotsInput{
+				SnapshotIds: []*string{task.SnapshotTaskDetail.SnapshotId},
+			})
+			if err != nil {
```
euank: nit; this could care about 'isNotExistError' vs other errors

platform/api/aws/images.go
```diff
@@ -289,3 +365,147 @@ func registerImageParams(snapshotID, name, description string, diskBaseName stri
...
+
+	describeSnapshotRes, err := a.ec2.DescribeSnapshots(&ec2.DescribeSnapshotsInput{
+		SnapshotIds: []*string{image.BlockDeviceMappings[0].Ebs.SnapshotId},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("couldn't describe snapshot: %v", err)
+	}
+	snapshot := describeSnapshotRes.Snapshots[0]
+
+	var count int
```
euank: I think a waitgroup expresses the intent of count more clearly than an int.

The channel could then turn into a buffered channel that's closed on `wg.Done` and ranged over.

cmd/ore/aws/copy-image.go
```diff
@@ -0,0 +1,58 @@
...
+		os.Exit(2)
+	}
+
+	amis, err := API.CopyImage(sourceImageID, args)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Couldn't copy images: %v\n", err)
+		os.Exit(1)
+	}
+
+	json.NewEncoder(os.Stdout).Encode(amis)
```
bgilbert: I've been avoiding returning errors from command handlers so we don't spew a usage message, but I'll add a check.

platform/api/aws/images.go
bgilbert: Yup, that's better.  I did not know about waitgroups.

platform/api/aws/images.go
bgilbert: Good call, and if we're wrapping copyImageIn(), the wrapper should create the copyImageResult.

bgilbert:
Updated!

PR #530, opened 2017-04-05
Add an option to cork to verify the manifest tag

dm0-:
This should verify the manifest tag signature before anything attempts to use the repo's contents.

After a quick look over the source, `repo` doesn't appear to run anything out of the repository in its `init` subcommand, so it should be safe to run `repo init`, `git tag -v`, then `repo sync` since the `repo init` is analogous to `git clone`.

The manifest repo was cloned in another place to read its version numbers, so this PR verifies it there as well before returning the versions.

### DIFF
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -241,6 +241,12 @@ func updateRepo() {
 		plog.Fatalf("repo init failed: %v", err)
 	}
 
+	if sigVerify {
+		if err := sdk.RepoVerifyTag(manifestBranch); err != nil {
+			plog.Fatalf("repo tag verification failed: %v", err)
+		}
+	}
+
 	if err := sdk.RepoSync(chrootName); err != nil {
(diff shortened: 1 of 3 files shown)


cmd/cork/create.go
```diff
@@ -105,6 +106,8 @@ func init() {
 		"manifest-branch", "master", "Manifest git repo branch")
 	creationFlags.StringVar(&manifestName,
 		"manifest-name", "default.xml", "Manifest file name")
+	creationFlags.BoolVar(&tagVerify,
+		"tag-verify", false, "Verify the manifest Git tag with GPG")
```
marineam: `verify-signature` sounds less ambigious with the existing verify flag.

marineam:
To be clear, I'm fine with merging this after adjusting the flag name but I think there is more work to do to complete the feature.

dm0-:
Okay, I have renamed it.

Issue #486, opened 2017-02-28
evaluate Terraform for cloud provider integration into Kola


philips:
Terraform has been working nicely for bootkube testing and elsewhere. Further, a bunch of users have been starting to integrate Terraform into their deployments.

It would be really useful to research if we could replace the code that spins-up/down and ssh's into servers with Terraform.

This would also help with enabling Azure and OpenStack testing for Ignition and Container Linux testing.

If you need any expertise in Terraform you can ping @s-urbaniak or @alexsomesan

cc @marineam @ethernetdan @pbx0

philips:
@marineam could you take a look and evaluate the feasibility of this?

bgilbert:
There's an initial implementation in #532, but we've opted not to pursue it further.  The reasons are, roughly:

- We need fine-grained control over how we interact with cloud providers, e.g. fetching instance console logs.
- kola works at a different level of abstraction than other Terraform-based test environments, so it'd be difficult to share a lot of code with them.

PR #532, opened 2017-04-05
Lunaform: a test framework around Terraform

marineam:


lunaform/names.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// NamePrefix generates a hopefully unique resource name prefix that
+// includes the current process, user, and test names for context.
+// The result is at most 57 characters so a small suffix can be added
+// and still fit into the 63 character host name limit.
+func NamePrefix(testName string) string {
+	procName := getProc() // 9 chars max
+	userName := getUser() // 9 chars max
+	testName = nameFilter(testName, 20)
```
pbx0: Is this `testName` a name designated per kola invocation and representative of say the PR job, or would it represent the individual harness test itself?

marineam: This corresponds to the name given in `Register(Test{Name: "foo"})`

pbx0: Just a thought then, when using prefix matching for names, they end up kinda being structured like a directory tree for me with lots of names and dots instead of slashes so they get a bit long. I could be convinced to use something else, but do we need the test name in this? I mostly use the `basename` feature to help reap lost machines so I really mostly care about what and who launched the test rather then the actual test the machine was assigned to. For example, in pluton's case, we have different PR jobs launching different tests and so I like to distinguish which PR job launched the machines I see floating around to know if they are in-use. If a random developer launches the machines I will see they are the default name (so the automatic username is definitely an improvement here).

Of course, kola could also be smarter about intercepting signals (the most common way we have machines floating loose of a test) which, I imagine, would reduce this use-case greatly.

pbx0:
This looks great so far. I think not allowing tests to destroy machines individually is a good thing as well. If a machine needs to be taken out a test can just issue a shutdown or something like that.

lunaform/names.go
marineam: Ok, how about I change the test name piece to read the jenkins build id from the environment if available and a date-time value otherwise.

pbx0: I think that'd be perfect.

marineam: Updated

lunaform/cluster.go
```diff
@@ -0,0 +1,269 @@
...
+	"strconv"
+
+	"github.com/coreos/mantle/harness"
+	"github.com/coreos/mantle/lunaform/tfdata"
+	"github.com/coreos/mantle/network"
+	"github.com/coreos/mantle/platform"
+	"github.com/coreos/mantle/platform/conf"
+)
+
+const gce = `
```
ethernetdan: @marineam what would need to happen to have the terraform be generated by tectonic-installer?

Related to coreos/tectonic-installer#6

/cc @alexsomesan @s-urbaniak @philips

alexsomesan: I don't have enough context around this. Can someone with more insight take a few moments to explain what this is?

marineam: @ethernetdan I don't know yet but more platforms and external terraform will need to be a later step.

@alexsomesan This is a new version of the test framework behind kola/pluton that uses terraform. The framework brings up a number of machines and executes a `go test` style test function. The current system creates machines via API calls (or running QEMU) but the hope is that using terraform will allow integration with the tectonic installer.

bgilbert:
@marineam Is the intention that `platform/machine/*` will eventually go away?

marineam:
@bgilbert not any time soon since QEMU can't migrate to this. platform/machine may get a number of cleanup/changes based on how this goes though.

marineam:
After my recent stream of improvements to kola this is the glue that it currently takes to run the existing kola tests on lunaform: https://github.com/marineam/mantle/commit/170ebd0157e02016ce976a79c7e7138f96d84dbc

A few things remaining:
 - Add an `Architecture()` method to the platform.Cluster interface so tests and framework code doesn't have to do any magic to figure out machine architecture. A `Platform()` method may be similarly useful.
 - Implement the empty user-data ignition tests differently so they do not depend on unmanaged external state. In the case of GCE the test should be able to add per-instance ssh-key metadata. In the case of AWS ssh keys will need to be created/destroyed as a terraform resource.
 - Implement `$discovery` user-data substitution (trivial)
 - Figure out what to do about tests that want to create instances on the fly (non-trivial)

bgilbert:
We're opting not to pursue this further for the reasons listed in https://github.com/coreos/mantle/issues/486#issuecomment-347414512.

PR #535, opened 2017-04-06
cork: update 'aws' platform in download-image

euank:
Ore and plume are gaining the ability to work with this format, so it
makes sense to be able to download it


euank:
Yeah, I was considering just switching it to vmdk entirely, but wasn't sure if we had backwards compatibility concerns.

Considering I don't think this is used anywhere right now, I think I'll just move it to always be vmdk.

PR #536, opened 2017-04-06
Misc test fixes

marineam:
NTP wasn't actually returning an error on failure, and Omaha was dropping into a manhole on failure.

### DIFF
--- a/kola/tests/misc/ntp.go
+++ b/kola/tests/misc/ntp.go
@@ -50,8 +50,6 @@ func NTP(c cluster.TestCluster) error {
 		return fmt.Errorf("Bad network config:\n%s", out)
 	}
 
-	plog.Info("Waiting for systemd-timesyncd.service")
-
 	checker := func() error {
 		out, err = m.SSH("systemctl status systemd-timesyncd.service")
 		if err != nil {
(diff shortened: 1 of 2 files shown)


kola/tests/misc/ntp.go
```diff
@@ -62,13 +60,12 @@ func NTP(c cluster.TestCluster) error {
...
 		}
 
-		plog.Info("systemd-timesyncd.service is working!")
 		return nil
 	}
 
 	err = util.Retry(60, 1*time.Second, checker)
 	if err != nil {
-		return nil
+		return err
```
dm0-: It seems cleaner to just `return util.Retry(...`, but LGTM if that violates some convention

kola/tests/misc/omaha.go
```diff
@@ -76,7 +75,6 @@ func OmahaPing(c cluster.TestCluster) error {
 
 	select {
 	case <-tc:
-		platform.Manhole(m)
```
dm0-: I recall when I first asked @crawford about using `platform.Manhole`, he told me to look here for an example usage, so I'm not sure if that is intentional or not.  It seems fine to me to remove it, though.

marineam: Manhole is just a special case debugging mechanism, if you aren't working on that specific test it isn't going to be clear why kola opened a shell and is waiting for input.

kola/tests/misc/ntp.go
marineam: I just went with a minor change because in a future PR I'm migrating from returning errors to calling Errorf/Fatalf and friends so this line of code is about to change again. I didn't want actual fixes to be hidden in that noise.

PR #537, opened 2017-04-06
kola/tests/locksmith: ignore a stray racy reboot error

dm0-:
There is still a race condition in the locksmith reboot test that is not as common as the last one, but suddenly got more annoying since I made kola failures fail the whole OS build again.  It appears to be an EOF from the Go SSH library's `Wait` function copying output, but since it fails rarely and randomly, this is just a shot in the dark.

### DIFF
--- a/kola/tests/locksmith/locksmith.go
+++ b/kola/tests/locksmith/locksmith.go
@@ -17,6 +17,7 @@ package locksmith
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"time"
 
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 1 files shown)


kola/tests/locksmith/locksmith.go
```diff
@@ -138,6 +139,8 @@ func locksmithCluster(c cluster.TestCluster) error {
 			output, err := m.SSH(cmd)
 			if _, ok := err.(*ssh.ExitMissingError); ok {
 				err = nil // A terminated session is perfectly normal during reboot.
+			} else if _, ok := err.(*io.EOF); ok {
```
euank: `err == io.EOF`?

dm0-: Oh, right.  That one is a variable, not a type.  Updated.

Issue #106, opened 2015-09-16
plume: build aws images


philips:
Plume should build AWS images free of java and python dependencies. This will unblock doing automated tests of AWS images.

crawford:
I don't follow. @mischief or @pbx0, can you elaborate?

bgilbert:
Fixed in https://github.com/coreos/mantle/pull/540.

PR #540, opened 2017-04-08
plume: Support AWS

bgilbert:
Addresses #2, #106, coreos/bugs#1900.

### DIFF
--- a/cmd/plume/prerelease.go
+++ b/cmd/plume/prerelease.go
@@ -54,7 +54,7 @@ func init() {
 	cmdPreRelease.Flags().BoolVarP(&preReleaseDryRun, "dry-run", "n", false,
 		"perform a trial run, do not make changes")
 	cmdPreRelease.Flags().StringVar(&verifyKeyFile,
-		"verify-key", "", "PGP public key to be used in verifing download signatures.  Defaults to CoreOS Buildbot (0412 7D0B FABE C887 1FFB  2CCE 50E0 8855 93D2 DCB4)")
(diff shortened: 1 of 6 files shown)


cmd/plume/prerelease.go
```diff
@@ -264,3 +280,253 @@ func azurePreRelease(ctx context.Context, client *http.Client, src *storage.Buck
...
+		foundBucketRegion := false
+		for _, region := range cloud.Regions {
+			if region != cloud.BucketRegion {
+				destRegions = append(destRegions, region)
+			} else {
+				foundBucketRegion = true
+			}
+		}
+		if !foundBucketRegion {
+			plog.Noticef("AMI will be created in %v but \"plume release\" will not make it public", cloud.BucketRegion)
```
euank: This could probably be fatal, right. afaik this shouldn't ever happen.

Also, if we allow this, then the ami map below potentially has a non-public image entry.

cmd/plume/prerelease.go
```diff
@@ -267,3 +280,239 @@ func azurePreRelease(ctx context.Context, client *http.Client, src *storage.Buck
...
+			return err
+		}
+		// compatibility
+		if err := upload(fmt.Sprintf("%v.txt", entry.Region),
+			bytes.NewReader([]byte(entry.PvAmi))); err != nil {
+			return err
+		}
+	}
+	// drop trailing pipe characters
+	hvmAll.Truncate(hvmAll.Len() - 1)
```
bgilbert: Nope, I'll do that.

cmd/plume/prerelease.go
bgilbert: You're right, we'd publish the AMI ID anyway.  Will make it fatal.

bgilbert:
Updated.

bgilbert:
Thanks @euank!

PR #550, opened 2017-04-12
pluton: bump timeouts

pbx0:
We are seeing very wide variance in container downloads: bump timeouts in
two places that are affected by this. cc @aaronlevy


pbx0:
coreosbot run pluton tests

PR #552, opened 2017-04-12
pluton: prefetch and cache containers

pbx0:
cc @aaronlevy @marineam 
This sets up a bastion node per test suite that prefetches a given list of containers and extracts them into archives on disk. Once this is done the tests run in parallel and before setting up bootkube, it pulls the container archives and extracts them into the store.

This actually slows down tests runs a bit for a variety of reasons (extra extraction steps, copying containers around on disk, and all nodes having to pull container archives from one machine), but it does allow us to run 7 tests in parallel again. Regardless, this is about stability and isolating as much of the internet downloads to the beginning of a test. Scp'ing around the files isn't the most efficient, but its relatively simple and should be reliable.

One disadvantage is that we have to maintain the list of containers to prefetch and keep them at the right version. At the least, our largest container, hyperkube, infers the tag. Its not fatal if we cache the wrong container, it just means  some internet variance could make it back into the test. In the future, maybe this list is defined along side the tests inside the bootkube repo after the tests migrate there.

You might notice the use of sync.Once is a little odd. The reason is because its important that the spawn package calls the initial `StartBastionOnce` because that package is aware of which containers to prefetch and the pluton harness shouldn't be. sync.Once is the simplest way to ensure that only the first caller causes the bastion to prefetch the images while all subsequent callers are blocked until the first call is complete.

  I also plan to come back around and have more timing profiles logged through the mantle harness, but that depends on first integrating such logging into the spawn package.

### DIFF
--- a/pluton/harness/run.go
+++ b/pluton/harness/run.go
@@ -26,6 +26,8 @@ import (
 	"github.com/coreos/mantle/pluton/spawn"
 )
 
+var bastionMachine platform.Machine
+
 // Call this from main after setting all the global options. Tests are filtered
 // by name based on the glob pattern given.
 func RunSuite(pattern string) {
@@ -45,12 +47,46 @@ func RunSuite(pattern string) {
 	}
 	suite := harness.NewSuite(opts, tests)
 
+	// setup a node for in cluster services not tied to individual test life-cycles
+	var cloud platform.Cluster
+	switch Opts.CloudPlatform {
+	case "gce":
+		var bastionDir = filepath.Join(Opts.OutputDir, "bastion")
+
+		err := os.MkdirAll(bastionDir, 0777)
+		if err != nil {
+			fmt.Printf("setting up bastion cluster: %v\n", err)
+			os.Exit(1)
+		}
+
+		cloud, err = gcloud.NewCluster(&Opts.GCEOptions, bastionDir)
+		if err != nil {
+			fmt.Printf("setting up bastion cluster: %v\n", err)
+			os.Exit(1)
+		}
+
+		bastionMachine, err = cloud.NewMachine("")
+		if err != nil {
+			fmt.Printf("setting up bastion cluster: %v\n", err)
+
+			cloud.Destroy()
+			os.Exit(1)
+		}
+	default:
+		fmt.Printf("invalid cloud platform %v\n", Opts.CloudPlatform)
+		os.Exit(1)
+	}
+
 	if err := suite.Run(); err != nil {
 		fmt.Fprintln(os.Stderr, err)
 		fmt.Println("FAIL")
+
+		cloud.Destroy()
 		os.Exit(1)
 	}
 	fmt.Println("PASS")
+
+	cloud.Destroy()
 	os.Exit(0)
 }
 
@@ -102,7 +138,7 @@ func runTest(t pluton.Test, h *harness.H) {
 		SelfHostEtcd:   t.Options.SelfHostEtcd,
 	}
 
-	c, err := spawn.MakeBootkubeCluster(cloud, config)
+	c, err := spawn.MakeBootkubeCluster(cloud, config, bastionMachine)
 	if err != nil {
 		h.Fatalf("creating cluster: %v", err)
 	}
--- a/pluton/harness/run.go
+++ b/pluton/harness/run.go
@@ -59,7 +59,11 @@ func RunSuite(pattern string) {
 			os.Exit(1)
 		}
 
-		cloud, err = gcloud.NewCluster(&Opts.GCEOptions, bastionDir)
+		// override machine size for bastion
+		bastionOpts := Opts.GCEOptions
+		bastionOpts.MachineType = "n1-standard-4"
+
+		cloud, err = gcloud.NewCluster(&bastionOpts, bastionDir)
 		if err != nil {
 			fmt.Printf("setting up bastion cluster: %v\n", err)
 			os.Exit(1)
@@ -72,6 +76,7 @@ func RunSuite(pattern string) {
 			cloud.Destroy()
 			os.Exit(1)
 		}
+
 	default:
 		fmt.Printf("invalid cloud platform %v\n", Opts.CloudPlatform)
 		os.Exit(1)
--- a/pluton/spawn/containercache/sshcache.go
+++ b/pluton/spawn/containercache/sshcache.go
@@ -1,6 +1,7 @@
 package containercache
 
 import (
+	"bytes"
 	"fmt"
 	"path/filepath"
 	"strings"
@@ -15,6 +16,7 @@ const (
 	cacheDir          = "/home/core/containercache"
 	downloadRetries   = 3
 	downloadRetryWait = 15 * time.Second
+	downloadTimeout   = 2 * time.Minute
 )
 
 var once sync.Once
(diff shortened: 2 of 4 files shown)


pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pbx0:
coreosbot run pluton tests

pluton/harness/run.go
```diff
@@ -45,6 +47,36 @@ func RunSuite(pattern string) {
...
+		if err := os.MkdirAll(Opts.OutputDir, 0777); err != nil {
+			fmt.Printf("setting up bastion cluster: %v\n", err)
+			os.Exit(1)
+		}
+
+		// override machine size for bastion
+		savedVal := Opts.GCEOptions.MachineType
+		Opts.GCEOptions.MachineType = "n1-standard-4"
+
+		cloud, err := gcloud.NewCluster(&Opts.GCEOptions, Opts.OutputDir)
```
marineam: I'd recommend putting it in a 'bastion' or similar sub-dir.

pluton/spawn/containercache/sshcache.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func copyPublicKeys(bastion platform.Machine, machines []platform.Machine) error {
+	for _, m := range machines {
+		err := platform.TransferFile(bastion, "/home/core/.ssh/bastion.key.pub", m, "/home/core/bastion.key.pub")
+		if err != nil {
+			return err
+		}
+
+		out, err := m.SSH("cat /home/core/bastion.key.pub >> /home/core/.ssh/authorized_keys")
```
marineam: This conflicts with anything update-ssh-keys does, instead you can do `update-ssh-keys -a bastion bastion.key.pub`

pluton/spawn/containercache/sshcache.go
```diff
@@ -0,0 +1,278 @@
...
+		err = startBastion(bastion, names)
+	}
+	once.Do(f)
+
+	return err
+}
+
+// Pull all containers cached by the bastion node into the docker and rkt store
+// of the given machines.
+func Pull(bastion platform.Machine, machines []platform.Machine) error {
```
marineam: This seems misnamed since ultimately the copy is performed by a push from bastion to the target machine. Perhaps Load or similar that doesn't imply direction would be more appropriate.

pluton/spawn/containercache/sshcache.go
```diff
@@ -0,0 +1,278 @@
...
+	return nil
+}
+
+func startBastion(bastion platform.Machine, names []ImageName) error {
+	if bastion == nil {
+		return fmt.Errorf("error starting containercache: bastion is nil")
+	}
+
+	// generate key-pair
+	out, err := bastion.SSH(`ssh-keygen -t ed25519 -N "" -f ./.ssh/bastion.key`)
```
marineam: Did agent forwarding not work?

pbx0:
coreosbot run pluton tests

pluton/spawn/containercache/sshcache.go
pbx0: I wasn't sure how to get the public key for this.

pluton/harness/run.go
pbx0: makes sense, will change

pluton/spawn/containercache/sshcache.go
pbx0: sounds good, will change to Load

pluton/spawn/containercache/sshcache.go
marineam: This returns the public keys from the cluster's agent: https://godoc.org/github.com/coreos/mantle/platform#BaseCluster.Keys

There isn't a directly exposed way to inject them into a config so you'll have to do something like this:
```go
var test_cluster, bastion_cluster platform.Cluster
var userdata string

parsed, err := conf.New(userdata)
keys, err := bastion_cluster.Keys()
parsed.CopyKeys(keys)
userdata = parsed.String()
test_cluster.NewMachine(userdata)
```

pbx0: That makes sense. However, I think I prefer to keep things the way they are because this complicates use of the containercache package. Mainly, machines in the test cluster much be given a crafted userdata with the key. And since the logic is lifted out of the containercache pkg, it complicates use of its API since spawn must know to make sure bastion and test machines are setup already to SSH to each other.

pluton/spawn/containercache/sshcache.go
pbx0: done

pbx0:
I've seen all tests pass 3 times (besides now disabled scale test) with bootkube v0.3.13. So looking good stability wise. Next steps are seeing how 1.6 looks and making sure tests are flexible to pick up the new flag changes, but that will be for other PRs.

pbx0:
coreosbot run pluton tests

pluton/harness/run.go
```diff
@@ -45,6 +47,38 @@ func RunSuite(pattern string) {
...
+	case "gce":
+		var bastionDir = filepath.Join(Opts.OutputDir, "bastion")
+
+		if err := os.MkdirAll(bastionDir, 0777); err != nil {
+			fmt.Printf("setting up bastion cluster: %v\n", err)
+			os.Exit(1)
+		}
+
+		// override machine size for bastion
+		savedVal := Opts.GCEOptions.MachineType
```
marineam: Since options is passed to NewCluster there is no need to use the global, just make a copy and edit that instead.

pluton/harness/run.go
```diff
@@ -45,6 +47,38 @@ func RunSuite(pattern string) {
...
+		Opts.GCEOptions.MachineType = "n1-standard-4"
+
+		cloud, err := gcloud.NewCluster(&Opts.GCEOptions, bastionDir)
```
marineam: Since the cluster is discarded and only the machine is kept there isn't anything to call cluster.Destroy. I think the only thing that leaks as a result is the ssh agent socket which isn't a huge deal but if you can it would be good to destroy the cluster too.

pluton/spawn/containercache/sshcache.go
marineam: Ok

pluton/harness/run.go
marineam: Actually, I don't see where the bastion gets destroyed. Am I just missing it?

pluton/spawn/containercache/sshcache.go
```diff
@@ -0,0 +1,278 @@
...
+	Engine string // docker or rkt
+}
+
+// StartBastionOnce will populate a bastion machine with the given containers
+// for rkt and docker. Since its best to run this only once per test suite, the
+// bastion machine should be in a separate platform.Cluster then the one used
+// in the test so it can be available for all tests in a given run. This
+// function is guarded with a sync.Once, so only the first within a test suite
+// it will have it run.
+func StartBastionOnce(bastion platform.Machine, names []ImageName) error {
```
marineam: It may be helpful to drop this run once magic and explicitly manage the life cycle of the bastion in the top level main() function or similar.

pluton/harness/run.go
pbx0: *runs to the cloud console to delete a bunch of nodes* Good catch. Just adding a `defer cloud.Destroy()` should work, since this function gets returned too after the tests are complete.

pluton/spawn/containercache/sshcache.go
pbx0: Once advantage of calling `StartBastionOnce` during the k8s cluster setup code is that it gives us the opportunity to intelligently figure out what containers to prefetch. Right now most things are hardcoded besides bootkube and hyperkube. 

Soon these tests will begin using a bootkube binary instead of a container which means we have the opportunity to move the calling of `StartBastionOnce` after we render the bootkube assets. The bootkube assets contain all the names of containers we need and it wouldn't be difficult to parse them out. This moves us away from a static, hard-coded list of containers to prefetch.

pluton/harness/run.go
marineam: That works.

pluton/harness/run.go
pbx0: done

pluton/harness/run.go
pbx0: Actually defer didn't work because of the Os.Exit's which I believe do not run deferred functions. Just added cloud.Destroy() all possible subsequent exits from the function.

PR #554, opened 2017-04-13
platform/aws: add missing errcheck

euank:
This would lead to a panic rather than a useful error message


euank:
Dupe of #499 which I forgot to merge, whoops!

PR #556, opened 2017-04-13
aws: govcloud tweaks

euank:
couple small bugs I ran into

cc @bgilbert


platform/api/aws/images.go
```diff
@@ -298,6 +298,10 @@ func (a *API) CreateImportRole(bucket string) error {
 	if err != nil {
 		if awserr, ok := err.(awserr.Error); ok && awserr.Code() == "NoSuchEntity" {
 			// Policy does not exist, let's try to create it
+			arnAwsPart := "aws"
+			if a.opts.Region == "us-gov-west-1" {
```
bgilbert: Feels hacky, but in all likelihood this case will never run again.

PR #558, opened 2017-04-14
pluton/spawn: bump static prefetched images for latest bootkube

pbx0:
In the future, as soon as we start using a bootkube binary directly we
can infer nearly all the containers we need to prefetch by starting our
prefetch post bootkube render and inspecting the manifests.

cc @yifan-gu


pbx0:
coreosbot run pluton tests

pbx0:
only tests that fail are the ones that yifan has a fix pending for, merging since its a minor change

PR #560, opened 2017-04-15
Fix omaha test to work around docker 17.04 bug

euank:
ipv6 is all sorts of wonky under 17.04, causing our Jenkins to fail.


euank:
@marineam I don't worry so much about leaking fds on test failure so long as it fails, but modified per your suggestion regardless.

euank:
Regardless of importance, I do appreciate the pointer! I was being sloppy and should have read the code properly first. Thanks!

PR #562, opened 2017-04-17
pluton: use init-master.sh script from bootkube repo directly

pbx0:
WIP since I'm still debugging locally on why this doesn't work.

cc @diegs


aaronlevy:
Did a quick scan. lgtm

pbx0:
coreosbot run pluton tests

PR #563, opened 2017-04-17
Add flag to disable etcd TLS for pluton.

diegs:
This allows the tests to pass after merging
https://github.com/kubernetes-incubator/bootkube/pull/433.

Since work is in progress to change pluton to use the hack scripts, this
is only a temporary fix, and it's easier than porting over all the TLS
certificate placement changes here too.


diegs:
Bootkube change is in, please merge when you have a chance. Thanks!

PR #573, opened 2017-04-19
kola: Convert filesystem and verity tests to subtests

bgilbert:
Unrelatedly, skip the last sleep in util.Retry().


bgilbert:
Updated to add comment.

PR #574, opened 2017-04-19
pluton/spawn: bump bootkube timeout

pbx0:
During the transition to using the hack/scripts directly I reduced the
bootkube timeout from 12 minutes to 5 by accident. Although bootkube is
normally much faster then 12 minutes (with container prefetching) the
timeout should not be this low.


aaronlevy:
lgtm

PR #579, opened 2017-04-21
kola/tests: Test automatic rollback after updating to a bad /usr

bgilbert:


bgilbert:
Updated to add MinVersions to tests.

PR #581, opened 2017-04-22
gangue: support downloading binary packages by a short name

dm0-:
This supports downloading and verifying a binary package by running e.g. `gangue get git`.  The package will be from the same CoreOS version as the host by default, but new options can change that.

### DIFF
--- a/cmd/gangue/portage.go
+++ b/cmd/gangue/portage.go
@@ -1,0 +1,119 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"regexp"
+	"strings"
+)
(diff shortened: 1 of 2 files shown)


cmd/gangue/portage.go
```diff
@@ -0,0 +1,119 @@
...
+// Search through Packages, returning the full URL to a matching binary package
+func findPackageURL(client *http.Client, name string) (string, error) {
+	urlPrefix := fmt.Sprintf("gs://builds.developer.core-os.net/boards/%s/%s/pkgs/", coreosBoard, coreosVersion)
+
+	file, err := ioutil.TempFile("", "Packages")
+	if err != nil {
+		return "", err
+	}
+	file.Close() // Close the file while its content is written
+	tempFile := file.Name()
```
marineam: Although this works I'd suggest calling Name() before Close()

cmd/gangue/portage.go
```diff
@@ -0,0 +1,119 @@
...
+	defer os.Remove(tempFile)
+
+	if err = downloadURL(client, urlPrefix+"Packages", tempFile); err != nil {
```
marineam: seems like at some point we'll need to add support to our download helpers to download to an io.Writer instead of file so buffers can be used in this sort of situation, but that can happen some other day.

cmd/gangue/portage.go
```diff
@@ -0,0 +1,119 @@
...
+	"io/ioutil"
+	"net/http"
+	"os"
+	"regexp"
+	"strings"
+)
+
+// Read default parameters from the host for building PORTAGE_BINHOST
+func getReleaseInfo() (board, version string) {
+	contents, err := ioutil.ReadFile("/usr/share/coreos/release")
```
marineam: If this command is expected to be used in the SDK I'd recommend using VersionsFromManifest and DefaultBoard to be consistent with all other tools.

dm0-:
At this point in CL's life cycle, we're not going to expand on developer workflows.  Closing this.

PR #583, opened 2017-04-24
AWS fixes

bgilbert:


platform/api/aws/api.go
```diff
@@ -95,3 +97,14 @@ func (a *API) PreflightCheck() error {
 
 	return err
 }
+
+// Get the AWS partition for this region
+func (a *API) getPartition() (string, error) {
```
euank: We can nix this partition code.

See https://github.com/euank/mantle/commit/9ae22da24c7b0aa10ae2cd01c527ffa43c857a37

Bump the aws-sdk-go and things can suddenly look prettier :)

bgilbert:
Updated.

cmd/ore/aws/aws.go
```diff
@@ -64,10 +64,12 @@ func preflightCheck(cmd *cobra.Command, args []string) error {
 		Options:         &platform.Options{},
 	})
 	if err != nil {
-		return fmt.Errorf("could not create AWS client: %v", err)
+		fmt.Fprintf(os.Stderr, "could not create AWS client: %v\n", err)
+		os.Exit(1)
```
euank: I do have a preference for returning errors where possible just so `defer`s run.

That's an argument for having an extra wrapper function (similar to what rkt does) that handles errors and doesn't print usage, or to panicing.

I'm okay with this since I don't think there are defers pending from the callers of these functions, but if we ever want to refactor, that wrapper function is worth thinking about imo.

PR #586, opened 2017-04-25
kola: AWS improvements

bgilbert:
- Double the launch timeout so we don't give up on PV instances prematurely
- Notice launch failures before the timeout
- Fix instance leaks on CreateInstances() failure
- Set the Name tag when creating instances


bgilbert:
Just pushed a fix to the last commit, in case you wanted to look again.

euank:
Still LGTM, good catch on the consistency bit

PR #590, opened 2017-04-28
platform/api/aws: Fix kola failures

bgilbert:


platform/machine/aws/cluster.go
```diff
@@ -84,7 +84,7 @@ func (ac *cluster) NewMachine(userdata string) (platform.Machine, error) {
 
 	conf.CopyKeys(keys)
 
-	instances, err := ac.api.CreateInstances(ac.Name(), ac.Name(), conf.String(), 1, true)
+	instances, err := ac.api.CreateInstances(ac.Name(), ac.Name(), conf.String(), 1)
```
euank: If this is the only place CreateInstances has to change, doesn't that mean 'count' is always 1?

bgilbert: Yes, I realized that after the fact.  Changing it to `CreateInstance` would simplify the driver, or alternatively `platform` could learn to request multiple machines at once.  Currently `platform` handles multiple launches by spawning a goroutine for each one, so I think the only thing we're losing with the current system is API throttling headroom -- which means your implied cleanup could make sense.  Thoughts?

euank: It seems possible a batch in a request could complete quicker than a batch of different requests depending on how placement works?

Since that's my (minor) preference, I guess it makes sense to keep it as-is and move to using the argument in the future. I'll properly review the rest now. ...

PR #592, opened 2017-04-28
Refuse to mess with _kola_temp while another kola process is using it

bgilbert:
and likewise for pluton.  I'm tired of accidentally breaking a test run by starting another one in the same directory.

### DIFF
--- a/cmd/kola/bootchart.go
+++ b/cmd/kola/bootchart.go
@@ -48,12 +48,14 @@ func runBootchart(cmd *cobra.Command, args []string) {
 		os.Exit(2)
 	}
 
+	var lockfile *os.File
 	var err error
-	outputDir, err = kola.CleanOutputDir(outputDir)
+	outputDir, lockfile, err = kola.CleanOutputDir(outputDir)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


cmd/kola/bootchart.go
```diff
@@ -48,12 +48,14 @@ func runBootchart(cmd *cobra.Command, args []string) {
...
 
+	var lockfile *os.File
 	var err error
-	outputDir, err = kola.CleanOutputDir(outputDir)
+	outputDir, lockfile, err = kola.CleanOutputDir(outputDir)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Setup failed: %v\n", err)
 		os.Exit(1)
 	}
+	defer lockfile.Close()
```
euank: It could be slightly nicer to use CLOEXEC just in case and make sure we dodge any os.Exit calls down the line.

If you want a nice package to do that, https://github.com/rkt/rkt/blob/v1.25.0/pkg/lock/file.go#L160-L190 has code for it already :)

bgilbert:
Closing in favor of #625.

PR #594, opened 2017-04-30
vendor: update golang.org/x/sys/unix

dm0-:
I ran the following, not sure if it's "correct".

    go get -u golang.org/x/sys/...
    godep update golang.org/x/sys/unix

The actual change isn't important; it just fixes 32-bit ARM for messing around by defining `O_TMPFILE`.


marineam:
Yup, that's correct. Not needed in this case but usually godep update would also need the /... suffix.

PR #595, opened 2017-05-02
platform/aws: create security groups if they don't exist

euank:
cc @bgilbert @marineam

### DIFF
--- a/platform/api/aws/ec2.go
+++ b/platform/api/aws/ec2.go
@@ -111,19 +111,23 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64, wait
 		ud = &tud
 	}
 
+	sgId, err := a.getSecurityGroupID(a.opts.SecurityGroup)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving security group: %v", err)
+	}
 	inst := ec2.RunInstancesInput{
-		ImageId:        &a.opts.AMI,
-		MinCount:       &cnt,
-		MaxCount:       &cnt,
-		KeyName:        &keyname,
-		InstanceType:   &a.opts.InstanceType,
-		SecurityGroups: []*string{&a.opts.SecurityGroup},
-		UserData:       ud,
+		ImageId:          &a.opts.AMI,
+		MinCount:         &cnt,
+		MaxCount:         &cnt,
+		KeyName:          &keyname,
+		InstanceType:     &a.opts.InstanceType,
+		SecurityGroupIds: []*string{&sgId},
+		UserData:         ud,
 	}
 
 	reservations, err := a.ec2.RunInstances(&inst)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error running instances: %v", err)
 	}
 
 	ids := make([]string, len(reservations.Instances))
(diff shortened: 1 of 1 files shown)


platform/api/aws/ec2.go
```diff
@@ -200,3 +204,89 @@ func (a *API) CreateTags(resources []string, tags map[string]string) error {
...
+	}
+	if len(sgIds.SecurityGroups) == 0 {
+		return "", fmt.Errorf("zero security groups matched name %v", name)
+	}
+	return *sgIds.SecurityGroups[0].GroupId, nil
+}
+
+// createSG creates a security group with tcp/22 access allowed from the
+// internet.
+func (a *API) createSG(name string) (string, error) {
```
bgilbert: `createSecurityGroup`?

platform/api/aws/ec2.go
```diff
@@ -200,3 +204,89 @@ func (a *API) CreateTags(resources []string, tags map[string]string) error {
...
+	sg, err := a.ec2.CreateSecurityGroup(&ec2.CreateSecurityGroupInput{
+		GroupName:   aws.String(name),
+		Description: aws.String("kola security group for testing"),
```
bgilbert: `mantle`

platform/api/aws/ec2.go
```diff
@@ -200,3 +204,89 @@ func (a *API) CreateTags(resources []string, tags map[string]string) error {
...
+		_, err := a.ec2.AuthorizeSecurityGroupIngress(&input)
+
+		if err != nil {
+			// We created the SG but can't add all the needed rules, let's try to
+			// bail gracefully
+			_, delErr := a.ec2.DeleteSecurityGroup(&ec2.DeleteSecurityGroupInput{
+				GroupId: sg.GroupId,
+			})
+			if delErr != nil {
+				return "", fmt.Errorf("created sg %v (%v) but couldn't authorize it. Manual deltion may be required: %v", *sg.GroupId, name, err)
```
bgilbert: `deletion`

platform/api/aws/ec2.go
```diff
@@ -200,3 +204,89 @@ func (a *API) CreateTags(resources []string, tags map[string]string) error {
...
+			}
+			return "", fmt.Errorf("created sg %v (%v), but couldn't authorize it and thus deleted it: %v", *sg.GroupId, name, err)
+		}
+	}
+	return *sg.GroupId, err
+}
+
+func isSecurityGroupNotExist(err error) bool {
```
bgilbert: This seems so specific that I'd be tempted to inline it, but I could see it going either way.

euank:
I think for ec2-classic you have to specify the vpc for a few of these calls. Maybe a created-default-vpc will still work correctly, I'm a little fuzzy on all the details.

platform/api/aws/ec2.go
```diff
@@ -200,3 +204,89 @@ func (a *API) CreateTags(resources []string, tags map[string]string) error {
...
+	})
+	if isSecurityGroupNotExist(err) {
+		return a.createSecurityGroup(name)
+	}
+
+	if err != nil {
+		return "", fmt.Errorf("unable to get security group named %v: %v", name, err)
+	}
+	if len(sgIds.SecurityGroups) == 0 {
+		return "", fmt.Errorf("zero security groups matched name %v", name)
```
marineam: Is it possible for this to happen and err to be nil?

euank: I don't think so, this is being overtly defensive

marineam: Ok, if it was normal I would have expected it to call create but being an unexpected case an error is fine.

PR #599, opened 2017-05-08
Pluton: add special hack for hyperkube-git.

diegs:
This is kind of ugly but seems to be the smallest way to let us test
non-release hyperkubes in the different repo.


diegs:
PTAL, updated to parse out the kubelet hyperkube image repo as well.

pbx0:
LGTM

PR #600, opened 2017-05-08
kola/tests/docker: use subtests for simple tests

euank:
This is related to the 'docker-skim' work where I wish to just call
'dockerMetaTests' on skim rather than throw the full suite at it since
duplicating the full suite has been a bit messy.

This also should speed things up a little.

Threw it at AWS once by hand and things looked fine.

cc @bgilbert

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -33,12 +33,6 @@ import (
 
 func init() {
 	register.Register(&register.Test{
-		Run:         dockerResources,
-		ClusterSize: 1,
-		Name:        "docker.resources",
-		UserData:    `#cloud-config`,
-	})
-	register.Register(&register.Test{
 		Run:         dockerNetwork,
 		ClusterSize: 2,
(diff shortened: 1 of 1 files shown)


kola/tests/docker/docker.go
```diff
@@ -85,18 +79,23 @@ func init() {
...
-		ClusterSize: 1,
-		Name:        "docker.networks-reliably",
-		UserData:    `#cloud-config`,
-	})
-	register.Register(&register.Test{
-		Run:         dockerUserNoCaps,
+		Run:         dockerMetaTests,
 		ClusterSize: 1,
-		Name:        "docker.user-no-caps",
+		Name:        `docker.meta`,
```
bgilbert: I find the `meta` naming confusing.  {`generic`,`misc`,`base`}?

</🚲🏚>

euank: Switched it back to 'base', I had that at first

PR #601, opened 2017-05-08
pluton: get kubectl from image

pbx0:
By getting kubectl from the hyperkube image we can simplify a lot of the crusty version handling we had to start hacking around. Should have done this this way from the start. Maybe kubectl used to not be in hyperkube?

cc @diegs


diegs:
LGTM!

PR #602, opened 2017-05-08
 	kola/tests/docker: add 'docker info' parsing test 

euank:
This validates filesystem selection, security options, etc.

This is to improve confidence in coreos/coreos-overlay#2562

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -16,8 +16,11 @@ package docker
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"os"
+	"reflect"
+	"sort"
 	"strings"
 	"time"
 
@@ -113,6 +116,9 @@ func genDockerContainer(m platform.Machine, name string, binnames []string) erro
 }
 
 func dockerBaseTests(c cluster.TestCluster) {
+	c.Run("docker-info", func(c cluster.TestCluster) {
+		testDockerInfo("overlay", c)
+	})
 	c.Run("resources", dockerResources)
 	c.Run("networks-reliably", dockerNetworksReliably)
 	c.Run("user-no-caps", dockerUserNoCaps)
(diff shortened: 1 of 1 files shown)


kola/tests/docker/docker.go
```diff
@@ -382,3 +388,91 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+	// Because we prefer overlay2/overlay for different docker versions, figure
+	// out the correct driver to be testing for based on our docker version.
+	var expectedOverlayDriver string
+	if strings.HasPrefix(info.ServerVersion, "1.12.") {
+		expectedOverlayDriver = "overlay"
+	} else if strings.HasPrefix(info.ServerVersion, "17.04") {
+		expectedOverlayDriver = "overlay"
+	} else if strings.HasPrefix(info.ServerVersion, "17.") {
+		expectedOverlayDriver = "overlay2"
+	} else {
```
bgilbert: Rather than breaking the test next January, I'd suggest defaulting to `overlay2`.  That will produce failures on old Docker versions, which is okay because we'll still notice it.

kola/tests/docker/docker.go
```diff
@@ -382,3 +388,91 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+	}
+
+	expectedFsDriver := expectedFsDriverMap[expectedFs]
+	if info.Driver != expectedFsDriver {
+		c.Errorf("unexpected driver: %v != %v", expectedFsDriver, info.Driver)
+	}
+
+	// Validations shared by all versions currently
+	if !reflect.DeepEqual(info.SecurityOptions, []string{"seccomp", "selinux"}) {
+		c.Errorf("unexpected security options: %+v: wanted selinux, seccomp", info.SecurityOptions)
```
bgilbert: DRY, probably by deleting the wanted list.

euank:
Addressed comments

PR #605, opened 2017-05-09
cmd/plume: Add "dev" channel for testing release process

bgilbert:


cmd/plume/specs.go
```diff
@@ -310,7 +373,7 @@ func AddSpecFlags(flags *pflag.FlagSet) {
 	flags.StringVarP(&specBoard, "board", "B",
 		board, "target board")
 	flags.StringVarP(&specChannel, "channel", "C",
-		"alpha", "channels: "+channels)
+		"dev", "channels: "+channels)
```
euank: perhaps this should just default to "" and be required instead.

cmd/plume/specs.go
```diff
@@ -152,6 +156,56 @@ var (
...
+			},
+			AWS: awsSpec{
+				BaseName:        "ContainerLinuxDev",
+				BaseDescription: "CoreOS Container Linux development image",
+				Prefix:          "coreos_production_ami_",
+				Image:           "coreos_production_ami_vmdk_image.vmdk.bz2",
+				Partitions: []awsPartitionSpec{
+					awsPartitionSpec{
+						Name:         "AWS West",
+						Profile:      "default",
```
euank: I think requiring the name to be `coroes-dev` is slightly safer, but I don't feel strongly.

cmd/plume/specs.go
bgilbert: It's convenient to not have to specify `dev` when developing.

cmd/plume/specs.go
bgilbert: Updated.

PR #606, opened 2017-05-10
cmd/plume: Enable VIRTIO_SCSI_MULTIQUEUE in GCE images

bgilbert:
on new releases only. Our kernels are new enough to support it, and it's supposed to provide a nice performance boost.


bgilbert:
This was used to release 1409.0.0.

PR #608, opened 2017-05-16
cmd/plume: Integrate GCE code with platform/api/gcloud; add GCE image pruning

bgilbert:


bgilbert:
Thanks for the review.

PR #609, opened 2017-05-17
Fix up some kola test output

dm0-:
This also includes fixing some command output being displayed as byte arrays.  (And dropping the absolute path wasn't required; it's just so it's possible to test without wrapper scripts.)

### DIFF
--- a/kola/tests/misc/files.go
+++ b/kola/tests/misc/files.go
@@ -140,7 +140,7 @@ func WritableFiles(c cluster.TestCluster) {
 	}
 
 	if string(output) != "" {
-		c.Fatalf("Unknown writable files found: %v", output)
+		c.Fatalf("Unknown writable files found: %s", output)
 	}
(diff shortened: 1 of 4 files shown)


kola/tests/misc/files.go
```diff
@@ -173,6 +173,7 @@ func StickyDirs(c cluster.TestCluster) {
 		"/dev/mqueue",
 		"/dev/shm",
 		"/media",
+		"/run/torcx/unpack",
```
bgilbert: If this doesn't need to be mode 1777, I'd rather the test fails until we can get the directory fixed.

lucab: I think it shouldn't, or at least it wasn't coded that way on purpose.

dm0-:
I've dropped the torcx path and just squashed the Docker path change into the output cleanup commit.

PR #612, opened 2017-05-17
Fetch upstream k8s versions programmatically.

diegs:
Rather than hardcoding versions when fetching kubectl from the web we
can curl them from a known URL scheme.


diegs:
cc @pbx0

pbx0:
Would this PR also solve your needs? https://github.com/coreos/mantle/pull/601

diegs:
Yes, I thought I saw something fly by :)

diegs:
Abandoning since #601 does this more nicely.

PR #617, opened 2017-05-25
aws: Default to us-west-2 everywhere, and t2.small

bgilbert:
It's too easy to leak instances when we have three different default regions.


bgilbert:
Updated kola default AMI to newer, less broken alpha.

PR #619, opened 2017-05-25
cmd/plume: Avoid pruning a specific GCE alpha image

bgilbert:


bgilbert:
plume sorts by the image creation time, so we're safe for awhile even without this change.  It's not clear how long we'll need to protect that image, though, so I took the cautious approach.

PR #624, opened 2017-06-02
kola: Save instance console logs and check them for errors

bgilbert:


bgilbert:
platform doesn't retry, but when it knows about failures it can do extra reporting (such as saving the console log).

PR #632, opened 2017-06-15
Support automating pre-release

dm0-:
The dev specs were just copied; maybe they should be define separately and reused in both specs.  Not sure if their values should diverge between automated and user dev builds.

### DIFF
--- a/cmd/plume/plume.go
+++ b/cmd/plume/plume.go
@@ -15,9 +15,13 @@
 package main
 
 import (
+	"io/ioutil"
+	"net/http"
+
 	"github.com/coreos/pkg/capnslog"
 	"github.com/spf13/cobra"
 
+	"github.com/coreos/mantle/auth"
 	"github.com/coreos/mantle/cli"
 )
 
@@ -27,7 +31,24 @@ var (
 		Use:   "plume [command]",
 		Short: "The CoreOS release utility",
 	}
+
+	gceJSONKeyFile string
 )
+
+func init() {
+	root.PersistentFlags().StringVar(&gceJSONKeyFile, "gce-json-key", "", "use a JSON key for authentication")
+}
+
+func getGoogleClient() (*http.Client, error) {
+	if gceJSONKeyFile != "" {
+		if b, err := ioutil.ReadFile(gceJSONKeyFile); err == nil {
+			return auth.GoogleClientFromJSONKey(b)
+		} else {
+			return nil, err
+		}
+	}
+	return auth.GoogleClient()
+}
 
 func main() {
 	cli.Execute(root)
--- a/cmd/plume/prerelease.go
+++ b/cmd/plume/prerelease.go
@@ -102,7 +102,7 @@ func runPreRelease(cmd *cobra.Command, args []string) error {
 
 	spec := ChannelSpec()
 	ctx := context.Background()
-	client, err := auth.GoogleClient()
+	client, err := getGoogleClient()
 	if err != nil {
 		plog.Fatal(err)
 	}
--- a/cmd/plume/specs.go
+++ b/cmd/plume/specs.go
@@ -157,7 +157,7 @@ var (
 		},
 	}
 	specs = map[string]channelSpec{
-		"dev": channelSpec{
+		"user": channelSpec{
 			BaseURL: "gs://users.developer.core-os.net/" + os.Getenv("USER") + "/boards",
 			Boards:  []string{"amd64-usr", "arm64-usr"},
 			Destinations: []storageSpec{storageSpec{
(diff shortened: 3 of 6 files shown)


cmd/plume/prerelease.go
```diff
@@ -91,6 +94,19 @@ func init() {
 }
 
 func runPreRelease(cmd *cobra.Command, args []string) error {
+	getClient := func() (client *http.Client, err error) {
+		if gceJSONKeyFile != "" {
+			b, err := ioutil.ReadFile(gceJSONKeyFile)
+			if err != nil {
+				plog.Fatal(err)
+			}
+			client, err = auth.GoogleClientFromJSONKey(b)
```
bgilbert: This will be needed for `plume release` also.  Maybe the similar code in platform/api/gcloud/api.go New() should be made a utility function in auth/google.go.

cmd/plume/specs.go
```diff
@@ -157,7 +157,7 @@ var (
 		},
 	}
 	specs = map[string]channelSpec{
-		"dev": channelSpec{
+		"user": channelSpec{
```
bgilbert: README.md should also be updated.

cmd/plume/specs.go
```diff
@@ -207,6 +207,56 @@ var (
...
+				Image:       "coreos_production_gce.tar.gz",
+				Publish:     "coreos_production_gce.txt",
+				Limit:       2,
+			},
+			AWS: awsSpec{
+				BaseName:        "ContainerLinuxDev",
+				BaseDescription: "CoreOS Container Linux development image",
+				Prefix:          "coreos_production_ami_",
+				Image:           "coreos_production_ami_vmdk_image.vmdk.bz2",
+				Partitions: []awsPartitionSpec{
```
bgilbert: This complexity (two partitions, three regions) exists to test `plume pre-release`.  For nightly Jenkins builds I think we should use `us-west-2` only.  The `LaunchPermissions` in the second `awsPartitionSpec` refers to `coreosdev` and should probably stay.

cmd/plume/specs.go
```diff
@@ -207,6 +207,56 @@ var (
 				},
 			},
 		},
+		"developer": channelSpec{
+			BaseURL: "gs://builds.developer.core-os.net/boards",
+			Boards:  []string{"amd64-usr", "arm64-usr"},
+			Destinations: []storageSpec{storageSpec{
```
bgilbert: We don't actually need the artifacts re-uploaded anywhere, so I'd be inclined to drop this.

cmd/plume/specs.go
```diff
@@ -207,6 +207,56 @@ var (
...
+				BaseURL:       "gs://builds.developer.core-os.net/releases",
+				NamedPath:     "current",
+				VersionPath:   true,
+				DirectoryHTML: true,
+				IndexHTML:     true,
+			}},
+			GCE: gceSpec{
+				Project:     "coreos-gce-testing",
+				Family:      "coreos-dev",
```
bgilbert: Should be `coreos-developer`, and probably the other one should be changed to `coreos-user`.

cmd/plume/specs.go
```diff
@@ -207,6 +207,56 @@ var (
...
+				Description: "CoreOS Container Linux development image",
+				Image:       "coreos_production_gce.tar.gz",
+				Publish:     "coreos_production_gce.txt",
+				Limit:       2,
```
bgilbert: It may be useful to keep more history here?  Jenkins currently does.

cmd/plume/specs.go
```diff
@@ -207,6 +207,56 @@ var (
...
+			},
+			AWS: awsSpec{
+				BaseName:        "ContainerLinuxDev",
```
bgilbert: Likewise, probably `ContainerLinux{Developer,User}`.

cmd/plume/specs.go
dm0-: I've dropped a partition and region, and added back the `LaunchPermissions`.

cmd/plume/specs.go
dm0-: I've set it to `25` like the others.

dm0-:
I've made the suggested changes, except I just moved the optional JSON key auth into `GoogleClient()`.  That grants automation support everywhere.

cmd/plume/prerelease.go
dm0-: I wasn't sure how to cleanly unify this in `auth` between the various command options, but I've added a command that uses a global option in `plume` so all subcommands support JSON keys.

cmd/plume/plume.go
```diff
@@ -27,8 +31,23 @@ var (
...
+	gceJSONKeyFile string
 )
 
+func init() {
+	root.PersistentFlags().StringVar(&gceJSONKeyFile, "gce-json-key", "", "use a JSON key for authentication")
+}
+
+func getGoogleClient() (*http.Client, error) {
+	if gceJSONKeyFile != "" {
+		if b, err := ioutil.ReadFile(gceJSONKeyFile); err == nil {
```
bgilbert: We should error out if a key file was specified and can't be read.

dm0-: Updated.

PR #636, opened 2017-06-17
Switch to Glide

bgilbert:
cc: @dghubble


dghubble:
Can you introduce a Makefile with a `vendor` target that has the glide and/or glide-vc commands used to produce the vendor? So we can be sure its reproducible. https://github.com/coreos/ignition/blob/master/Makefile#L2

dghubble:
I suppose since this repo is using scripts for these things that's ok. I also needed `hg`, but `revendor` produced no diff :+1:

dghubble:
LGTM

glide.yaml
```diff
@@ -0,0 +1,332 @@
...
+  - vhdcore/vhdfile
+  - vhdcore/writer
+- package: github.com/ajeddeloh/go-json
+  version: 73d058cf8437a1989030afe571eeab9f90eebbbd
+- package: github.com/ajeddeloh/yaml
+  version: 1072abfea31191db507785e2e0c1b8d1440d35a5
+- package: github.com/alecthomas/units
+  version: 2efee857e7cfd4f3d0138cc3cbb1b4966962b93a
+- package: github.com/aws/aws-sdk-go
+  version: 40bc7761a9f06daa574d20f2ad5454db02a05953
```
euank: I'd rather use version tags for things, like the aws-sdk, where there are suitable tags available.

bgilbert: See the second commit.  I haven't checked commit IDs in cases where Glide declined to convert to tags, but in the specific case of `aws-sdk-go`, `40bc776` is `v1.8.15-5-g40bc776`.

I'd prefer to leave substantive dependency updates for another day.

bgilbert:
Converted to Makefile target.

PR #639, opened 2017-06-22
platform/machine/qemu: Wait for OEM block device before mounting it

bgilbert:
Hopefully fixes intermittent `mount(8)` failures in Jenkins.


bgilbert:
Switched to Retry.

PR #644, opened 2017-06-28
kola/aws: allow specifying a channel for ami id

euank:
This allows a caller to use the channel name instead of the AMI id.

It's meant as a simple quality-of-life improvement.

### DIFF
--- a/platform/api/aws/ami.go
+++ b/platform/api/aws/ami.go
@@ -1,0 +1,91 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aws
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"sync"
+)
+
+// relaseAMIs matches the structure of the AMIs listed in our
+// coreos_production_ami_all.json release file
+type releaseAMIs struct {
+	AMIS []struct {
+		Name string `json:"name"`
+		PV   string `json:"pv"`
+		HVM  string `json:"hvm"`
+	} `json:"amis"`
+}
+
+var amiCache struct {
+	alphaOnce sync.Once
+	alphaAMIs *releaseAMIs
+
+	betaOnce sync.Once
+	betaAMIs *releaseAMIs
+
+	stableOnce sync.Once
+	stableAMIs *releaseAMIs
+}
+
+// resolveAMI is used to minimize network requests while allowing resolution of
+// release channels to specific AMI ids.
+// If any issue occurs attempting to resolve a given AMI, e.g. a network error,
+// this method panics.
+func resolveAMI(ami string, region string) string {
+	resolveChannel := func(channel string) *releaseAMIs {
+		resp, err := http.DefaultClient.Get(fmt.Sprintf("https://%s.release.core-os.net/amd64-usr/current/coreos_production_ami_all.json", channel))
+		if err != nil {
+			panic(fmt.Errorf("unable to fetch %v AMI json: %v", channel, err))
+		}
+
+		var amis releaseAMIs
+		err = json.NewDecoder(resp.Body).Decode(&amis)
+		if err != nil {
+			panic(fmt.Errorf("unable to parse release bucket %v AMI: %v", channel, err))
+		}
+		return &amis
+	}
+
+	var channelAmis *releaseAMIs
+	switch ami {
+	case "alpha":
+		amiCache.alphaOnce.Do(func() {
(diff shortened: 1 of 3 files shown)


cmd/kola/kola.go
```diff
@@ -128,6 +128,7 @@ func writeProps() error {
 	type AWS struct {
 		Region       string `json:"region"`
 		AMI          string `json:"ami"`
+		ResolvedAMI  string `json:"ami-id",omitempty`
```
bgilbert: I suspect consumers won't care how the original request was phrased, and won't want to check two different fields for the AMI ID.  I'd suggest putting the ID in `AMI` in all cases.

platform/api/aws/api.go
```diff
@@ -44,9 +46,15 @@ type Options struct {
 	// SecretKey is the optional secret key to use. It will override all other sources
 	SecretKey string
 
+	// AMI is the AWS AMI to launch EC2 instances with.
+	// If it is one of the special strings alpha|beta|stable, it will be resolved to an actual ID.
 	AMI           string
 	InstanceType  string
 	SecurityGroup string
+
+	resolvedAMILock sync.Mutex
```
bgilbert: `Options` is for things that are passed in, and this doesn't really fit.  The mutex should go in `API`, and indeed so should `ResolvedAMI`, since it can only be safely accessed via a non-exported mutex.

platform/api/aws/ami.go
```diff
@@ -0,0 +1,77 @@
...
+	// Use cache if we can
+	if a.opts.ResolvedAMI != "" {
+		return a.opts.ResolvedAMI, nil
+	}
+
+	// If successfully resolved, cache the AMI id for next call.
+	// Decide based on the named returns
+	defer func() {
+		if resolveAMIErr == nil {
+			a.opts.ResolvedAMI = resolvedAMI
```
bgilbert: The cache doesn't save any work unless an HTTP request is made, so for clarity this can be moved out of a defer to the point where the channel name is resolved to an AMI.

platform/api/aws/ami.go
```diff
@@ -0,0 +1,77 @@
...
+	case "alpha", "beta", "stable":
+	default:
+		return a.opts.AMI, nil
+	}
+
+	// Lookup the current latest AMI id via our release bucket.
+	// Note, querying the AWS API is an alternate option, but is slower and
+	// requires filtering carefully to avoid shared-but-not-public images. It
+	// also won't work as trivially for china/gov.
+	resp, err := http.DefaultClient.Get(fmt.Sprintf("https://%s.release.core-os.net/amd64-usr/current/coreos_production_ami_all.json", a.opts.AMI))
```
bgilbert: Should this be using the GS code?  There doesn't seem to be much advantage in this case.

platform/api/aws/ami.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return "", fmt.Errorf("unable to describe %v AMI: %v", a.opts.AMI, err)
+	}
+
```
bgilbert: It would be nice to sign the AMIs JSON and verify the signature here, but I don't think that's a blocker for this PR.

platform/api/aws/ami.go
euank: This seems simpler, so I'd rather leave it be unless there's a reason not to.

euank:
@bgilbert addressed comments

platform/api/aws/ami.go
```diff
@@ -28,22 +28,9 @@ type releaseAMIs struct {
...
-	// If successfully resolved, cache the AMI id for next call.
-	// Decide based on the named returns
-	defer func() {
-		if resolveAMIErr == nil {
-			a.opts.ResolvedAMI = resolvedAMI
-		}
-	}()
+func (a *API) AMI() (string, error) {
+	a.amiResolutionLock.Lock()
+	defer a.amiResolutionLock.Unlock()
```
bgilbert: I hadn't anticipated that you'd update `a.opts.AMI` in place.  That is indeed the typical approach in mantle, but it now races against other API objects sharing the Options struct (e.g. in `kola run --parallel`).  Arguably this can be made simpler and more consistent by dropping the `AMI` method and lock (or, if we want to be correct, making the lock static or moving it back to Options) and moving resolution into `New()`.

euank: Moving resolution to `New` is a problem because some users of `platform...aws` don't actually need to deal with AMI resolution, e.g. `plume`, and so might not be able to describe images.

I think moving it back to options is the sanest option.

I don't want to drop it entirely because I want to dodge ever doing multiple http requests when we only really need one.

bgilbert: We should only get into the resolution code path if the AMI field is populated, so plume shouldn't be affected.

I can live with moving the lock back to Options.

euank: .. oh, yeah, it's also not doing a `describe-images` call anymore so moot point. That was the original reason for not putting it in `New`. that comment can be half ignored. Options still seems reasonable to me though.

euank:
Back to the lock being in options

euank:
Fixed the accidental mutex-copy CI caught via a `Copy` helper. Any ideas on how to do this more cleanly would be welcome.

euank:
After online discussion, I think we've come to the conclusion that there aren't great answers. The "stupid simple" answer is a static package-level cache. I've moved it to that. *sigh*

euank:
I give it a 20% chance we see a PR in the future saying "Fix panic" and switching back to proper error handling 🐑

PR #645, opened 2017-06-28
tests/kubernetes: fix 'kubernetes.docker' test

euank:
This test was inadvertently broken in #633 because it relied on
cloud-config's default behavior of populating `/etc/environment`.

This is the simplest fix for that breakage.

### DIFF
--- a/kola/tests/kubernetes/setup.go
+++ b/kola/tests/kubernetes/setup.go
@@ -82,14 +82,10 @@ func setupCluster(c cluster.TestCluster, nodes int, version, runtime string) (*k
 	}
 
 	// configure nodes via generic install scripts
-	if err := runInstallScript(master, controllerInstallScript, options); err != nil {
-		return nil, fmt.Errorf("Installing controller: %v", err)
-	}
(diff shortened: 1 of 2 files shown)


kola/tests/kubernetes/setup.go
```diff
@@ -82,14 +82,10 @@ func setupCluster(c cluster.TestCluster, nodes int, version, runtime string) (*k
 	}
 
 	// configure nodes via generic install scripts
-	if err := runInstallScript(master, controllerInstallScript, options); err != nil {
```
bgilbert: Maybe the remaining error returns in `setupCluster` should also be converted to `Fatal`s, so the function signature doesn't imply that errors will be recoverable?

euank: Done

dm0-:
Should this be enabled on any other platforms while it's being changed?  Maybe just exclude `qemu` if all it needs is Internet access?

euank:
@dm0- I'm still working on updating it to fix the rkt variant. I'll look at platforms along with that.

Issue #647, opened 2017-06-28
cork: not all sudo binaries have --prompt


jedsmith:
CentOS 7.3 under Vagrant. Trying to get a CoreOS SDK going, and `cork` dies on a clean build:

        [w] [vagrant@<snip> sdk]$ cork create --replace --sdk-version=master --verbose
        2017-06-28T21:45:24Z cli: Started logging at level INFO
        2017-06-28T21:45:24Z cork: Downloading SDK version master
        2017-06-28T21:45:24Z sdk: Downloading https://storage.googleapis.com/builds.developer.core-os.net/sdk/amd64/master/coreos-sdk-amd64-master.tar.bz2 to /home/vagrant/sdk/.cache/sdks/coreos-sdk-amd64-master.tar.bz2
        coreos-sdk-amd64-master.tar.bz2: [=============================] 547 MB/547 MB
        2017-06-28T21:45:33Z sdk: Downloaded 547230910 bytes
        2017-06-28T21:45:33Z sdk: Downloading https://storage.googleapis.com/builds.developer.core-os.net/sdk/amd64/master/coreos-sdk-amd64-master.tar.bz2.sig to /home/vagrant/sdk/.cache/sdks/coreos-sdk-amd64-master.tar.bz2.sig
        coreos-sdk-amd64-master.tar.bz2.sig: [=========================] 543 B/543 B
        2017-06-28T21:45:34Z sdk: Downloaded 543 bytes
        2017-06-28T21:45:36Z sdk: Verified file: /home/vagrant/sdk/.cache/sdks/coreos-sdk-amd64-master.tar.bz2
        2017-06-28T21:45:36Z sdk: Path does not exist: /home/vagrant/sdk/chroot
        2017-06-28T21:45:36Z sdk: Unpacking SDK into /home/vagrant/sdk/chroot
        2017-06-28T21:45:36Z sdk: Using /home/vagrant/sdk/.cache/sdks/coreos-sdk-amd64-master.tar.bz2
     >> sudo: invalid option -- '-'
        usage: sudo [-D level] -h | -K | -k | -V
        usage: sudo -v [-AknS] [-D level] [-g groupname|#gid] [-p prompt] [-u user name|#uid]
        usage: sudo -l[l] [-AknS] [-D level] [-g groupname|#gid] [-p prompt] [-U user name] [-u
                    user name|#uid] [-g groupname|#gid] [command]
[...]

PR #648, opened 2017-06-28
sdk: Use short options for sudo

bgilbert:
Long options were added in sudo 1.8.8. CentOS 7.3 has 1.8.6.

Fixes https://github.com/coreos/mantle/issues/647.

cc: @jedsmith


jedsmith:
Checked out against `bgilbert:p` and re-ran my automation, looks better now. Thanks! 👍

Issue #103, opened 2015-09-14
kola: vmware support


mischief:
we should investigate using kola to run tests on vmare.

https://github.com/vmware/govmomi looks like it might be useful.

PR #656, opened 2017-07-19
*: add esx support

arithx:
Fixes https://github.com/coreos/mantle/issues/103


platform/api/esx/api.go
```diff
@@ -0,0 +1,832 @@
...
+		return nil, fmt.Errorf("couldn't find base VM: %v", err)
+	}
+
+	folders, err := datacenter.Folders(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("getting datacenter folders: %v", err)
+	}
+	folder := folders.VmFolder
+
+	cloneSpec, err := a.server.BuildCloneSpec(baseVM, folder, network, resourcePool, datastore)
```
bgilbert: Missing error check.

platform/machine/esx/cluster.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+	}
+
+	return ac.BaseCluster.Destroy()
+}
+
+func (ac *cluster) vmname() string {
+	b := make([]byte, 5)
+	rand.Read(b)
+	return fmt.Sprintf("%s-%x", ac.Name()[0:13], b)
```
bgilbert: If ESX doesn't have length limits on the VM name, you should probably drop the `[0:13]`.

platform/machine/esx/machine.go
```diff
@@ -0,0 +1,121 @@
...
+		return err
+	}
+
+	if am.journal != nil {
+		if err := am.journal.Destroy(); err != nil {
+			return err
+		}
+	}
+
+	// faster when run after termination
```
bgilbert: This comment is AWS-specific.

platform/machine/esx/cluster.go
```diff
@@ -0,0 +1,211 @@
...
+
+		if err := api.AddKey(bc.Name(), keys[0].String()); err != nil {
+			return nil, err
+		}
+	}
+
+	return ac, nil
+}
+
+func (ac *cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error) {
```
bgilbert: By convention the `cluster` variable should be `ec` for ESX, and likewise `em` for `machine`s.

platform/api/esx/api.go
```diff
@@ -0,0 +1,869 @@
...
+	deadline, _ := context.WithDeadline(ctx, time.Now().Add(1000*time.Second))
+	ip, err := vm.WaitForNetIP(deadline, false)
+	if err != nil {
+		return nil, fmt.Errorf("waiting for net ip: %v", err)
+	}
+
+	var pub string
+	var priv string
+
+	/*var privateCIDRs []*net.IPNet
```
bgilbert: No commented-out code, please.

platform/api/esx/api.go
```diff
@@ -0,0 +1,869 @@
...
+
+	archive := &TapeArchive{flagOVAPath}
+	envelope, err := archive.ReadEnvelope("*.ovf")
+	if err != nil {
+		return archive, nil, fmt.Errorf("reading envelope: %v", err)
+	}
+
+	ovfHandler := object.NewOvfManager(server.client.Client)
+	cisp := types.OvfCreateImportSpecParams{
+		DiskProvisioning:   "",
```
bgilbert: Redundant; just omit the empty fields.

platform/api/esx/api.go
```diff
@@ -0,0 +1,869 @@
...
+func (server *ESXServer) GetServerDefaults() (*find.Finder, *object.Datacenter, *object.ResourcePool, *object.Datastore, object.NetworkReference, error) {
+	finder := find.NewFinder(server.client.Client, true)
+	datacenter, err := finder.DefaultDatacenter(server.ctx)
+	if err != nil {
+		return finder, nil, nil, nil, nil, err
+	}
+	finder.SetDatacenter(datacenter)
+	resourcePool, err := finder.DefaultResourcePool(server.ctx)
+	if err != nil {
+		return finder, datacenter, nil, nil, nil, err
```
bgilbert: By convention, in Go, functions usually either succeed or fail completely.  Just return all `nil`s on error, or even better, return a struct with fields.

platform/api/esx/api.go
```diff
@@ -0,0 +1,869 @@
...
+
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(f)
+	return buf.String(), nil
+}
+
+func (a *API) CreateDevice(name string, conf *conf.Conf) (*ESXMachine, error) {
+	ignitionData := base64.StdEncoding.EncodeToString(conf.Bytes())
+
+	finder, datacenter, resourcePool, datastore, network, err := a.server.GetServerDefaults()
```
bgilbert: Missing error check.

platform/conf/conf.go
```diff
@@ -224,6 +224,40 @@ func (c *Conf) Bytes() []byte {
...
+		Contents: contents,
+		Enable: enable,
+	})
+}
+
+func (c *Conf) addSystemdUnitV2(name, contents string, enable bool) {
+	c.ignitionV2.Systemd.Units = append(c.ignitionV2.Systemd.Units, v2types.Unit{
+		Name: name,
+		Contents: contents,
+		Enable: enable,
```
bgilbert: `Enable` is deprecated in 2.1; use `Enabled`.

arithx:
Did fixes for all of your comments, will test against the vSphere machine in the morning to make sure I didn't break anything else and will push up barring everything is working.

platform/conf/conf.go
arithx: Some tests are still using 2.0 configs, should I update those to 2.1 in a seperate commit (or PR)? Without updating those changing this to Enabled causes ignition to fail with an unrecognized key.

bgilbert: Oh, yes, you're right.  Go ahead and leave it as `Enable` for now.  Once https://github.com/coreos/ignition/pull/418 merges, I'll be changing `platform/conf` to use separate data structures for 2.0 and 2.1 configs, and we can fix it then.

platform/api/esx/api.go
```diff
@@ -0,0 +1,759 @@
...
+	}
+
+	uri := fmt.Sprintf("%s/serial.out", name)
+
+	p := soap.DefaultDownload
+
+	f, _, err := datastore.Download(a.ctx, uri, &p)
+	if err != nil {
+		return "", fmt.Errorf("couldn't download console logs: %v", err)
+	}
```
dgonyeo: `defer f.Close()`

platform/api/esx/api.go
```diff
@@ -0,0 +1,759 @@
...
+
+	buf := new(bytes.Buffer)
```
dgonyeo: nit: I'd typically get the contents of `f` via `ioutil.ReadAll`

platform/api/esx/api.go
```diff
@@ -0,0 +1,759 @@
...
+	cloneSpec, err := a.buildCloneSpec(baseVM, folder, network, resourcePool, datastore, ignitionData)
+
+	task, err := baseVM.Clone(a.ctx, folder, name, *cloneSpec)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't clone base VM: %v", err)
+	}
+
+	err = task.Wait(a.ctx)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't clone base VM: %v", err)
```
dgonyeo: having identical error messages (here and the previous `fmt.Errorf`) becomes annoying when one of them gets hit and you don't know which

platform/api/esx/api.go
```diff
@@ -0,0 +1,759 @@
...
+	envelope, err := archive.ReadEnvelope("*.ovf")
+	if err != nil {
+		return nil, nil, fmt.Errorf("reading envelope: %v", err)
+	}
+
+	ovfHandler := object.NewOvfManager(a.client.Client)
+	cisp := types.OvfCreateImportSpecParams{
+		EntityName:         name,
+		OvfManagerCommonParams: types.OvfManagerCommonParams{
+			Locale:           "US"},
```
dgonyeo: It doesn't look like `gofmt` has been run here? Also nit, maybe move this to the line above it.

platform/api/esx/api.go
```diff
@@ -0,0 +1,759 @@
...
+	if err != nil {
+		return nil, nil, nil, nil, nil, err
+	}
+
+	defaultNetwork, err := finder.DefaultNetwork(a.ctx)
+	if err != nil {
+		return nil, nil, nil, nil, nil, err
+	}
+
+	return finder, datacenter, resourcePool, datastore, defaultNetwork, nil
```
dgonyeo: Perhaps a `serverDefaults` struct would be a saner way to hold on to all of these values?

platform/api/esx/api.go
```diff
@@ -0,0 +1,759 @@
...
+	networks := map[string]string{}
+
+	if e.Network != nil {
+		for _, net := range e.Network.Networks {
+			networks[net.Name] = net.Name
+		}
+	}
+
+	for src, dst := range networks {
+		if net, err := finder.Network(ctx, dst); err == nil {
```
dgonyeo: It looks like this is the only place `ctx` is used, maybe just put `context.TODO()` inline here?

platform/api/esx/archive.go
```diff
@@ -0,0 +1,100 @@
...
+		if err != nil {
+			return nil, 0, err
+		}
+
+		if matched {
+			return &ArchiveEntry{r, f}, h.Size, nil
+		}
+	}
+
+	_ = f.Close()
```
dgonyeo: Why not just `defer f.Close()` when it's created? By putting it here you'll fail to close `f` in most possible outcomes here.

platform/api/esx/archive.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return nil, 0, os.ErrNotExist
```
dgonyeo: Why return `os.ErrNotExist`? It looks like the problem was that we didn't find the file we wanted in a tar archive, not that the tar archive doesn't exist (which is what I would initially think with this error)

platform/machine/esx/machine.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+
+	path := filepath.Join(em.dir, "console.txt")
+	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0644)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	f.WriteString(em.console)
```
dgonyeo: The error of this should be checked

platform/api/esx/api.go
```diff
@@ -0,0 +1,759 @@
...
+	if card == nil {
+		return nil, fmt.Errorf("No network device found.")
+	}
+
+	netDev, err := getNetworkDevice(network)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get new network backing device: %v", err)
+	}
+
+	card.Backing = netDev.(types.BaseVirtualEthernetCard).GetVirtualEthernetCard().Backing
```
arithx: I haven't seen it fail and this line of code is based on the govc code for building the cloning spec:

https://github.com/vmware/govmomi/blob/master/govc/vm/clone.go#L289

cmd/kola/kola.go
```diff
@@ -140,6 +140,12 @@ func writeProps() error {
 		InstallerImageBaseURL string `json:"installer"`
 		ImageURL              string `json:"image"`
 	}
+	type ESX struct {
+		Server     string `json:"esx_server"`
+		User       string `json:"esx_user"`
+		Password   string `json:"esx_password"`
+		BaseVMName string `json:"esx_base_vm_name"`
+	}
```
bgilbert: They serve completely different purposes, but happen to have the same contents at the moment.  The current approach is the right one.

platform/api/esx/api.go
```diff
@@ -0,0 +1,759 @@
...
+	// overwritten whenever the machine boots
+	opts = append(opts, &types.OptionValue{
+		Key:   "guestinfo.coreos.config.data",
+		Value: ignitionData,
+	})
+	opts = append(opts, &types.OptionValue{
+		Key:   "guestinfo.coreos.config.data.encoding",
+		Value: "base64",
+	})
+	ovfEnv := fmt.Sprintf(`<Environment oe:id="" xmlns="http://schemas.dmtf.org/ovf/environment/1" xmlns:oe="https://schemas.dmtf.org/ovf/environment/1" xmlns:xml="http://www.w3.org/XML/1998/namespace" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"><PropertySection><Property oe:key="guestinfo.coreos.config.data" oe:value="%s"/><Property oe:key="guestinfo.coreos.config.data.encoding" oe:value="base64"/></PropertySection></Environment>`, ignitionData)
```
arithx: ignitionData was base64 encoded at this point. However this function should have been removed, I missed that this was still here. Was a remnant of when I was trying to figure out alternative ways to set the guestinfo properties without them being overwritten by the hypervisor.

platform/api/esx/archive.go
arithx: would you rather I return something like `fmt.Errrorf("couldn't find file in archive")`?

dgonyeo: Yeah, that seems better to me

platform/api/esx/api.go
```diff
@@ -0,0 +1,742 @@
...
+	if err != nil {
+		return fmt.Errorf("getting datacenter folders: %v", err)
+	}
+	folder := folders.VmFolder
+
+	entity, err := a.uploadToResourcePool(archive, defaults.resourcePool, cisr, folder)
+	if err != nil {
+		return fmt.Errorf("uploading disks to ResourcePool: %v", err)
+	}
+	_ = object.NewVirtualMachine(a.client.Client, *entity)
```
bgilbert: Is that an error you're ignoring?  Deserves a comment either way.

platform/api/esx/api.go
```diff
@@ -0,0 +1,742 @@
...
+
+	return nil
+}
+
+func (a *API) TerminateDevice(vmName string) error {
```
bgilbert: Nit: for consistency, the argument should probably be `name`.

platform/api/esx/api.go
```diff
@@ -0,0 +1,742 @@
...
+
+	task, err = vm.Destroy(a.ctx)
+	if err != nil {
+		return fmt.Errorf("destroying vm: %v", vm)
+	}
+
+	return task.Wait(a.ctx)
+}
+
+func (a *API) buildCISR(name string, flagOVAPath string, finder *find.Finder, defaultNetwork object.NetworkReference, resourcePool *object.ResourcePool, datastore *object.Datastore) (*Archive, *types.OvfCreateImportSpecResult, error) {
```
bgilbert: Nit: Consider spelling out the function name for clarity.

platform/api/esx/api.go
```diff
@@ -0,0 +1,742 @@
...
+	var nets []types.OvfNetworkMapping
+	nets = append(nets, types.OvfNetworkMapping{
+		Name:    "test",
```
bgilbert: Is this the intended name?  If you need something specific to this codebase, consider `mantle`.

platform/api/esx/api.go
```diff
@@ -0,0 +1,742 @@
...
+		resourcePool: resourcePool,
+		datastore:    datastore,
+		network:      defaultNetwork,
+	}, nil
+}
+
+func (a *API) uploadToResourcePool(archive *Archive, resourcePool *object.ResourcePool, cisr *types.OvfCreateImportSpecResult, folder *object.Folder) (*types.ManagedObjectReference, error) {
+	lease, err := resourcePool.ImportVApp(a.ctx, cisr.ImportSpec, folder, nil)
+	if err != nil {
+		return nil, err
```
bgilbert: When to wrap errors is a judgment call, but please `err` on the side of wrapping if there's any chance the context could be ambiguous.

platform/api/esx/api.go
```diff
@@ -0,0 +1,742 @@
...
+func networkMap(finder *find.Finder, e *ovf.Envelope) (p []types.OvfNetworkMapping) {
+	networks := map[string]string{}
+
+	if e.Network != nil {
+		for _, net := range e.Network.Networks {
+			networks[net.Name] = net.Name
+		}
+	}
+
+	for src, dst := range networks {
```
bgilbert: `src` and `dst` are always equal, so this seems unnecessarily confusing.

platform/api/esx/api.go
```diff
@@ -0,0 +1,742 @@
...
+	var property []types.VAppPropertySpec
+
+	var mvm mo.VirtualMachine
+	err := vm.Properties(a.ctx, vm.Reference(), []string{"config", "config.vAppConfig", "config.vAppConfig.property"}, &mvm)
+	if err != nil {
+		return fmt.Errorf("couldn't get config.vappconfig: %v", err)
+	}
+
+	for _, item := range mvm.Config.VAppConfig.(*types.VmConfigInfo).Property {
+		if item.Id == "guestinfo.coreos.config.data" {
```
bgilbert: Does coreos-cloudinit also support this property?  If so, we should call it `userdata` rather than `ignitionData`.  If not, the top-level API call should error out if passed a `conf` which is not `IsIgnition`.

platform/api/esx/archive.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/vmware/govmomi/ovf"
+)
+
+type Archive struct {
```
bgilbert: I believe none of this is used outside the package, so the structs and methods should all be private unless necessary to fulfill an interface.

platform/api/esx/archive.go
```diff
@@ -0,0 +1,99 @@
...
+			return nil, 0, err
+		}
+
+		matched, err := path.Match(name, path.Base(h.Name))
+		if err != nil {
+			return nil, 0, err
+		}
+
+		if matched {
+			return &ArchiveEntry{r, f}, h.Size, nil
```
bgilbert: You're closing `f` on return from this function, so I'm surprised this works.  It should only be closed on error.

platform/api/esx/archive.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse ovf: %s", err.Error())
+	}
+
+	return e, nil
+}
+
+func (t *Archive) Open(name string) (io.ReadCloser, int64, error) {
+	f, err := os.Open(t.path)
+	defer f.Close()
```
bgilbert: Nil dereference on error.

platform/api/esx/archive.go
```diff
@@ -0,0 +1,99 @@
...
+
+	e, err := ovf.Unmarshal(r)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse ovf: %s", err.Error())
+	}
+
+	return e, nil
+}
+
+func (t *Archive) Open(name string) (io.ReadCloser, int64, error) {
```
bgilbert: `name` is actually a glob pattern, so should probably be called something else for clarity.

platform/machine/esx/cluster.go
```diff
@@ -0,0 +1,125 @@
...
+	plog = capnslog.NewPackageLogger("github.com/coreos/mantle", "platform/machine/esx")
+)
+
+type cluster struct {
+	*platform.BaseCluster
+	api *esx.API
+}
+
+// NewCluster creates an instance of a Cluster suitable for spawning
+// instances on Amazon Web Services' Elastic Compute platform.
```
bgilbert: Nope.

platform/machine/esx/cluster.go
```diff
@@ -0,0 +1,125 @@
...
+//
+// NewCluster will consume the environment variables $ESX_HOST, $ESX_USER,
+// and $ESX_PASSWORD to spawn instances and authenticate.
```
bgilbert: Is this actually true?

platform/machine/esx/cluster.go
```diff
@@ -0,0 +1,125 @@
...
+		BaseCluster: bc,
+		api:         api,
+	}
+
+	return ec, nil
+}
+
+func (ec *cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error) {
+	conf, err := ec.RenderUserData(userdata, map[string]string{
+		"$private_ipv4": "${COREOS_ESX_IPV4_PRIVATE_0}",
```
bgilbert: I think this should also set `$public_ipv4`?

platform/machine/esx/cluster.go
```diff
@@ -0,0 +1,125 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	conf.AddSystemdUnit("coreos-metadata.service", "[Unit]\nDescription=VMware metadata agent\n\n[Service]\nType=oneshot\nEnvironment=OUTPUT=/run/metadata/coreos\nExecStart=/usr/bin/mkdir --parent /run/metadata\nExecStart=/usr/bin/bash -c 'echo \"COREOS_ESX_IPV4_PRIVATE_0=$(ip addr show ens192 | grep -Po \"inet \\K[\\d.]+\")\" > ${OUTPUT}'\n", false)
```
bgilbert: Please use a multiline raw string literal for readability.

platform/machine/esx/cluster.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, err
+	}
+
+	ec.AddMach(mach)
+
+	return mach, nil
+}
+
+func (ec *cluster) Destroy() error {
+	return ec.BaseCluster.Destroy()
```
bgilbert: If all you need to do is call the base class method, you can drop this method.

platform/machine/esx/cluster.go
```diff
@@ -0,0 +1,125 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	conf.AddSystemdUnit("coreos-metadata.service", "[Unit]\nDescription=VMware metadata agent\n\n[Service]\nType=oneshot\nEnvironment=OUTPUT=/run/metadata/coreos\nExecStart=/usr/bin/mkdir --parent /run/metadata\nExecStart=/usr/bin/bash -c 'echo \"COREOS_ESX_IPV4_PRIVATE_0=$(ip addr show ens192 | grep -Po \"inet \\K[\\d.]+\")\" > ${OUTPUT}'\n", false)
+
+	id := uuid.NewV4()
+
+	instance, err := ec.api.CreateDevice(id.String(), conf)
```
bgilbert: The device name should be `ec.Name()` plus a random suffix.  That suffix can use some random bytes as other drivers do; a full UUID is unnecessary.

platform/machine/esx/machine.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (em *machine) saveConsole() error {
+	var err error
+	em.console, err = em.cluster.api.GetConsoleOutput(em.ID())
+	if err != nil {
+		return err
+	}
+
+	path := filepath.Join(em.dir, "console.txt")
+	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0644)
```
bgilbert: We're inconsistent about this (i.e. I wrote buggy code) but the mode should really be 0666 so the umask can have its say.

platform/api/esx/api.go
arithx: Adding a comment but this is a VirtualMachine object that is being returned.

platform/api/esx/api.go
arithx: Yes, both ignition & cloudinit support this property.

cmd/ore/esx/esx.go
```diff
@@ -0,0 +1,63 @@
...
+	API     *esx.API
+	options esx.Options
+)
+
+func init() {
+	ESX.PersistentFlags().StringVar(&options.Server, "server", "", "ESX Server IP")
+	ESX.PersistentFlags().StringVar(&options.User, "user", "", "ESX User")
+	ESX.PersistentFlags().StringVar(&options.Password, "password", "", "ESX Password")
+	ESX.PersistentFlags().StringVar(&options.Profile, "profile", "", "Profile")
+	ESX.PersistentFlags().StringVar(&options.BaseVMName, "name", "", "name of base VM")
```
arithx: I moved it into the base command instead of in subcommands as both of the existing sub commands (create-base & delete-vm) require name to be given.

PR #658, opened 2017-08-02
kola/tests/rkt: add systemd+rkt integration tests

euank:
This test is meant to catch a regression in v1492.0.0 whereby `rkt run`
would fail under a systemd unit.

Since Container Linux is the point where a pair of (rkt, systemd) are
run together, and the default rkt `stage1-coreos` interacts with
systemd, it makes sense to test this integration here, not just in rkt's
repo.

The tests introduced here attempt to cover the user-observable details
of the integration between rkt/systemd.

### DIFF
--- a/kola/tests/rkt/rkt.go
+++ b/kola/tests/rkt/rkt.go
@@ -15,12 +15,17 @@
 package rkt
 
 import (
+	"bytes"
 	"fmt"
+	"strings"
 	"time"
 
+	"github.com/coreos/mantle/kola"
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/platform/conf"
+	"github.com/coreos/mantle/platform/machine/qemu"
 	"github.com/coreos/mantle/util"
 )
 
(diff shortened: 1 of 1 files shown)


euank:
I think this is good for review now, cc @bgilbert @lucab

kola/tests/rkt/rkt.go
```diff
@@ -63,3 +75,131 @@ func rktEtcd(t cluster.TestCluster) {
 		t.Fatalf("etcd in rkt failed health check: %v", err)
 	}
 }
+
+// we use subtests to improve testing performance here. Creating the aci is
+// more expensive than actually running most of these tests.
+func rktBase(t cluster.TestCluster) {
+	m := t.Machines()[0]
+
+	// TODO this should not be necessary, but is at the time of writing
```
euank: @lucab Know anything about this offhand?

kola/tests/rkt/rkt.go
```diff
@@ -63,3 +75,131 @@ func rktEtcd(t cluster.TestCluster) {
...
+		}
+
+		machinectlOutput, err := m.SSH(fmt.Sprintf("machinectl show rkt-$(cat %s)", uuidFile))
+		if err != nil {
+			t.Fatalf("error running machinectl: %v, %v", err, string(output))
+		}
+
+		for _, line := range []string{"State=running", "Class=container", "Service=rkt"} {
+			if !bytes.Contains(machinectlOutput, []byte(line)) {
+				t.Errorf("expected machinectl to include %q: was %v", string(line), string(machinectlOutput))
```
bgilbert: We've dumped the entire output, so perhaps this would be better as a `Fatalf`.

kola/tests/rkt/rkt.go
lucab: Not directly, but I guess the selinux policy we ship is not compatible with the systemd inside stage1 (v222).

PR #659, opened 2017-08-03
platform/conf: Use Ignition 2.0.0 parser for 2.0.0 configs

bgilbert:
A 2.1.0 struct can contain a 2.0.0 config, but when serialized it contains empty JSON objects that aren't defined in the 2.0.0 spec, causing Ignition to produce warnings. Since Ignition now exports a function for determining the config version, use it to select the correct parser for the config version we have.

### DIFF
--- a/platform/conf/conf.go
+++ b/platform/conf/conf.go
@@ -22,11 +22,14 @@ import (
 
 	ct "github.com/coreos/container-linux-config-transpiler/config"
 	cci "github.com/coreos/coreos-cloudinit/config"
+	"github.com/coreos/go-semver/semver"
+	ign "github.com/coreos/ignition/config"
 	v1 "github.com/coreos/ignition/config/v1"
 	v1types "github.com/coreos/ignition/config/v1/types"
-	v2 "github.com/coreos/ignition/config/v2_1"
-	v2types "github.com/coreos/ignition/config/v2_1/types"
-	"github.com/coreos/ignition/config/validate/report"
+	v2 "github.com/coreos/ignition/config/v2_0"
+	v2types "github.com/coreos/ignition/config/v2_0/types"
+	v21 "github.com/coreos/ignition/config/v2_1"
+	v21types "github.com/coreos/ignition/config/v2_1/types"
 	"github.com/coreos/pkg/capnslog"
 	"golang.org/x/crypto/ssh/agent"
 )
@@ -55,6 +58,7 @@ type UserData struct {
 type Conf struct {
 	ignitionV1  *v1types.Config
 	ignitionV2  *v2types.Config
+	ignitionV21 *v21types.Config
 	cloudconfig *cci.CloudConfig
 	script      string
 }
@@ -98,13 +102,13 @@ func Unknown(data string) *UserData {
 		data: data,
 	}
 
-	_, _, err := v2.Parse([]byte(data))
+	_, _, err := v21.Parse([]byte(data))
 	switch err {
-	case v2.ErrEmpty:
+	case v21.ErrEmpty:
 		u.kind = kindEmpty
-	case v2.ErrCloudConfig:
+	case v21.ErrCloudConfig:
 		u.kind = kindCloudConfig
-	case v2.ErrScript:
+	case v21.ErrScript:
 		u.kind = kindScript
 	default:
 		// we don't autodetect Container Linux configs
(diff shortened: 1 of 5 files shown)


platform/conf/conf.go
```diff
@@ -266,11 +282,18 @@ func (c *Conf) copyKeysIgnitionV1(keys []*agent.Key) {
 }
 
 func (c *Conf) copyKeysIgnitionV2(keys []*agent.Key) {
-	var keyObjs []v2types.SSHAuthorizedKey
+	c.ignitionV2.Passwd.Users = append(c.ignitionV2.Passwd.Users, v2types.User{
```
euank: I think we technically want to merge with any existing core user if one exists, otherwise I believe this clobbers existing config.

platform/conf/conf.go
```diff
@@ -143,30 +147,34 @@ func (u *UserData) Render() (*Conf, error) {
...
-		ignc, report, err := v2.Parse([]byte(u.data))
-		if err == nil {
+		switch ver {
+		default:
+			ignc, report, err := v21.Parse([]byte(u.data))
+			if err != nil {
+				plog.Errorf("invalid userdata: %v", report)
+				return nil, err
+			}
+			c.ignitionV21 = &ignc
```
euank: Is it possible for this to happen? Wouldn't Parse always return `ErrrUnknownVersion` or the like?

platform/conf/conf.go
bgilbert: This change just restores the code removed by c99fa78bcac8c202b01f5e20ff7a01f2012f9760.  I believe you're correct, but it's not currently an issue with the existing tests.

platform/conf/conf.go
bgilbert: Possible for which to happen?  We end up in this switch arm for all Ignition 2.1 configs, and for any config that `Version` can't parse.

platform/conf/conf.go
```diff
@@ -184,12 +192,7 @@ func (u *UserData) Render() (*Conf, error) {
 			plog.Warningf("rendering Container Linux config: %s", report)
 		}
 
-		// ct still returns 2.0 configs. Convert to 2.1.
-		buf, err := json.Marshal(ignc)
-		if err != nil {
-			return nil, fmt.Errorf("serializing Container Linux config: %v", err)
-		}
-		return Ignition(string(buf)).Render()
+		c.ignitionV2 = &ignc
```
arithx: Comments indicate that previously we were converting the data into 2.1 configs, is it intentional that we're leaving them as 2.0 now?

platform/conf/conf.go
euank: Ah, thanks for explaining; I was missing that 2.1s fall into this default too.

platform/conf/conf.go
arithx: This code should take care of it if we want to be ahead of it:

```
func (c *Conf) copyKeysIgnitionV1(keys []*agent.Key) {
    found := false
    for _, user := range c.ignitionV1.Passwd.Users {
        if user.Name == "core" {
                found = true
                if user.SSHAuthorizedKeys != nil {
                    user.SSHAuthorizedKeys = append(user.SSHAuthorizedKeys, keysToStrings(keys)...)
                } else {
                    user.SSHAuthorizedKeys = keysToStrings(keys)
                }
        }
    }
    if !found {
        c.ignitionV1.Passwd.Users = append(c.ignitionV1.Passwd.Users, v2types.User{
            Name:              "core",
            SSHAuthorizedKeys: keysToStrings(keys),
        })
    }
}

func (c *Conf) copyKeysIgnitionV2(keys []*agent.Key) {
    found := false
    for _, user := range c.ignitionV2.Passwd.Users {
        if user.Name == "core" {
                found = true
                if user.SSHAuthorizedKeys != nil {
                    user.SSHAuthorizedKeys = append(user.SSHAuthorizedKeys, keysToStrings(keys)...)
                } else {
                    user.SSHAuthorizedKeys = keysToStrings(keys)
                }
        }
    }
    if !found {
        c.ignitionV2.Passwd.Users = append(c.ignitionV2.Passwd.Users, v2types.User{
    		Name:              "core",
    		SSHAuthorizedKeys: keysToStrings(keys),
    	})
    }
}

func (c *Conf) copyKeysIgnitionV21(keys []*agent.Key) {
    var keyObjs []v21types.SSHAuthorizedKey
    for _, key := range keys {
        keyObjs = append(keyObjs, v21types.SSHAuthorizedKey(key.String()))
    }
    found := false
    for _, user := range c.ignitionV21.Passwd.Users {
        if user.Name == "core" {
                found = true
                if user.SSHAuthorizedKeys != nil {
                    user.SSHAuthorizedKeys = append(user.SSHAuthorizedKeys, keyObjs...)
                } else {
                    user.SSHAuthorizedKeys = keyObjs
                }
        }
    }
    if !found {
        c.ignitionV2.Passwd.Users = append(c.ignitionV21.Passwd.Users, v21types.PasswdUser{
            Name:              "core",
            SSHAuthorizedKeys: keyObjs,
        })
    }
}
```

platform/conf/conf.go
bgilbert: The comment could have been clearer.  We were converting to 2.1 data structures, but the contents were left in 2.0 format.  With this change we're leaving everything as 2.0.

platform/conf/conf.go
euank: I'm fine punting this to the future since, as mentioned, this doesn't break anything in practice and isn't a new issue introduced in this pr.

bgilbert: Will punt for now.

PR #661, opened 2017-08-03
Enable kola on arm64 machines

glevand:
Allows kola to run on arm64 hosts.  Also includes a patch that increases the debug output of dnsmasq.

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -372,9 +372,17 @@ func runTest(h *harness.H, t *register.Test, pltfrm string) {
 func architecture(pltfrm string) string {
 	nativeArch := "amd64"
 	if pltfrm == "qemu" && QEMUOptions.Board != "" {
-		nativeArch = strings.SplitN(QEMUOptions.Board, "-", 2)[0]
+		nativeArch = boardToArch(QEMUOptions.Board)
+	}
+	if pltfrm == "packet" && PacketOptions.Board != "" {
+		nativeArch = boardToArch(PacketOptions.Board)
 	}
 	return nativeArch
+}
+
+// returns the arch part of an sdk board name
+func boardToArch(board string) string {
+	return strings.SplitN(board, "-", 2)[0]
 }
 
 // scpKolet searches for a kolet binary and copies it to the machine.
--- a/platform/local/dnsmasq.go
+++ b/platform/local/dnsmasq.go
@@ -45,14 +45,26 @@ type Dnsmasq struct {
 	dnsmasq  *exec.ExecCmd
 }
 
-var configTemplate = template.Must(template.New("dnsmasq").Parse(`
+const (
+	numInterfaces = 16
+	numSegments   = 3
+
+	debugConfig = `
+log-queries
+log-dhcp
+`
+
+	quietConfig = `
+quiet-dhcp
+quiet-dhcp6
+quiet-ra
+`
+
+	commonConfig = `
 keep-in-foreground
 leasefile-ro
 log-facility=-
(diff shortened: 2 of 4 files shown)


platform/local/dnsmasq.go
```diff
@@ -79,13 +91,11 @@ dhcp-host={{.HardwareAddr}}{{template "ips" .DHCPv4}}{{template "ips" .DHCPv6}}
...
 {{define "ips"}}{{range .}}{{printf ",%s" .IP}}{{end}}{{end}}
-`))
-
-const (
-	numInterfaces = 16
-	numSegments   = 3
+`
 )
 
+var plog = capnslog.NewPackageLogger("github.com/coreos/mantle", "local")
```
bgilbert: The second argument should be `platform/local`.

kola/harness.go
```diff
@@ -369,7 +369,7 @@ func runTest(h *harness.H, t *register.Test, pltfrm string) {
 
 // architecture returns the machine architecture of the given platform.
 func architecture(pltfrm string) string {
-	nativeArch := "amd64"
+	nativeArch := system.PortageArch()
```
bgilbert: If kola is running on arm64 but `pltfrm` is e.g. `aws`, `architecture` should still return `amd64`.  It looks like the right place to fix the default selection is in `sdk.DefaultBoard()`.  That will fix `QEMUOptions.Board` via `defaultTargetBoard` in `cmd/kola/options.go`.

platform/machine/qemu/cluster.go
```diff
@@ -134,12 +134,14 @@ func (qc *Cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error)
...
 			"-machine", "accel=kvm",
 			"-cpu", "host",
+			"-m", "1024",
 		}
 	case "arm64-usr":
 		qmCmd = []string{
 			"qemu-system-aarch64",
 			"-machine", "virt",
 			"-cpu", "cortex-a57",
+			"-m", "2048",
```
bgilbert: Just to double-check, this is necessary with images built with `./build_image prod`?  On amd64, dev images seem to need more than 1024 MB (https://github.com/coreos/bugs/issues/1594) but prod images don't.

kola/harness.go
glevand: OK, please check what I have in the updated patch.

platform/machine/qemu/cluster.go
glevand: Yes, the prod image will run out of memory.  Strange behavior and unexpected errors start to occur.

glevand:
Rebase to latest.

glevand:
In the current patch set the only change I've made is for sdk.DefaultBoard() to use system.PortageArch() for its default value, which will be the host machine's arch (from runtime.GOARCH).  QEMUOptions.Board should now be set to either a user specified `--board` value, or a default of the host machine.

kola.architecture() should now return amd64 for gce, aws, and esx, and either the `--board` value or the host machine's arch for qemu.

With my new patch `kola/harness: Use PacketOptions.Board` packet will have the same value as qemu, either the `--board` value or the host machine's arch.  Does it make more sense to have a default of amd64 for packet?  It doesn't seem it would be easy to arrange that though.

bgilbert:
> Does it make more sense to have a default of amd64 for packet? It doesn't seem it would be easy to arrange that though.

I think it's fine to default Packet to the host platform.  We should arguably switch the default `--packet-plan` to `baremetal_2a` on ARM, but we can leave that for another day.

PR #663, opened 2017-08-10
kola/tests/misc: add a test to create a new user

dgonyeo:
This commit adds a test to create a new user, and verifies that the
user's password and ssh key were set


bgilbert:
This is really testing Ignition rather than the OS generally.  Perhaps it belongs in the ignition hierarchy or a new CLC one.

dgonyeo:
After looking at this again I'm satisfied with just the blackbox testing covering this use case, since you're right in that this is really just testing Ignition.

PR #664, opened 2017-08-10
kola/tests/misc: add test for root fs on a raid device

dgonyeo:
This commit adds a test wherein the root filesystem is on a software
raid device, and tests whether the machine can survive a reboot and come
up successfully.

### DIFF
--- a/kola/tests/misc/raid.go
+++ b/kola/tests/misc/raid.go
@@ -1,0 +1,129 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package misc
+
+import (
+	"encoding/json"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform"
+	"github.com/coreos/mantle/platform/conf"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         RootOnRaid,
+		ClusterSize: 1,
+		Name:        "coreos.disk.raid.root",
+		// FIXME: This can only work on qemu, since it's overwriting
+		// /usr/share/oem/grub.cfg. The setting being appended (set
+		// linux_append="rd.auto") should probably be the default, and this
+		// should be removed after the OS level fix is made.
+		// https://github.com/coreos/bugs/issues/2099
+		Platforms: []string{"qemu"},
+		UserData: conf.ContainerLinuxConfig(`storage:
+  raid:
+    - name: "ROOT"
+      level: "raid1"
+      devices:
+        - "/dev/disk/by-partlabel/ROOT"
+        - "/dev/disk/by-partlabel/USR-B"
+  filesystems:
+    - name: "ROOT"
+      mount:
+        device: "/dev/md/ROOT"
+        format: "ext4"
+        create:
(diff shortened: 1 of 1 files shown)


kola/tests/misc/raid.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/coreos/mantle/platform/conf"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         RootOnRaid,
+		ClusterSize: 1,
+		Name:        "coreos.misc.root-on-raid",
+		// This will only work on qemu, since it's overwriting
+		// /usr/share/oem/grub.cfg
```
bgilbert: ...pending us fixing this properly.  The comment should have a FIXME and perhaps a bug link and an explanation that we should fix the OS to not require the command-line option.

kola/tests/misc/raid.go
```diff
@@ -0,0 +1,80 @@
...
+      mount:
+        device: "/dev/md127"
+        format: "ext4"
+        create:
+          options:
+            - "-L"
+            - "ROOT"
+    - name: "OEM"
+      mount:
+        device: "/dev/vda6"
```
bgilbert: Please avoid qemu-specific identifiers.

kola/tests/misc/raid.go
```diff
@@ -0,0 +1,80 @@
...
+  raid:
+    - name: "ROOT"
+      level: "raid1"
+      devices:
+        - "/dev/disk/by-partlabel/ROOT"
+        - "/dev/disk/by-partlabel/USR-B"
+  filesystems:
+    - name: "ROOT"
+      mount:
+        device: "/dev/md127"
```
bgilbert: Is there a by-name symlink?  I'm not sure if we can assume `127`.

kola/tests/misc/raid.go
```diff
@@ -0,0 +1,80 @@
...
+package misc
+
+import (
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform/conf"
+)
+
+func init() {
+	register.Register(&register.Test{
```
bgilbert: It would be useful to have a second test that puts a data volume on RAID, doesn't change the kernel command line, and runs on all platforms.

kola/tests/misc/raid.go
```diff
@@ -0,0 +1,80 @@
...
+		Run:         RootOnRaid,
+		ClusterSize: 1,
+		Name:        "coreos.misc.root-on-raid",
```
bgilbert: Test names don't follow the package hierarchy and we don't use `misc`.  `coreos.disk.raid.root`?

kola/tests/misc/raid.go
dgonyeo: I'll do this in a followup PR

kola/tests/misc/raid.go
dgonyeo: I'm not seeing anything in `/dev/disk` at the very least, so I'm not sure what else to do here other than maybe suggest Ignition gains the ability to parse `/etc/mdstat`

bgilbert: `63-md-raid-arrays.rules` talks about `/dev/disk/by-id/md-name-<name>` and `/dev/md/<name>`; do those not exist?

dgonyeo: Oh my bad, I completely missed those. Updates the test to use `/dev/md/ROOT`

kola/tests/misc/raid.go
```diff
@@ -0,0 +1,107 @@
...
+		c.Fatalf("couldn't unmarshal lsblk output: %v", err)
+	}
+
+	for _, b := range l.Blockdevices {
+		if b.Mountpoint != nil && *b.Mountpoint == "/" {
+			if b.Type != "raid1" {
+				c.Fatalf("device %q is mounted at / with type %q (was expecting raid1)", b.Name, b.Type)
+			}
+		}
+	}
```
dgonyeo: After updating it to fail if it didn't find the root mountpoint, I realized that the json wasn't quite organized how I thought. Updated to fix that, and to properly fail if the root mountpoint isn't found.

dgonyeo:
Good to merge, I lack access to do so myself.

PR #665, opened 2017-08-10
kola/tests/misc: test if selinux survives reboots

dgonyeo:
This commit extends the selinux test to follow the instructions in our docs to make selinux survive reboots, and then reboots the machine and checks if selinux is still enabled.


dgonyeo:
Good to merge, but I lack merge access

dgonyeo:
I had missed that reboot implementations don't merely reboot. I was trying to add a test to make sure the docs here were right:
https://coreos.com/os/docs/latest/selinux.html#enable-selinux-enforcement

Guess this isn't going to work out

euank:
Good catch

bgilbert:
It'd also be reasonable to gate SELinux enablement on a RuntimeConfig flag.

PR #670, opened 2017-08-11
cork/create: fix interactive shell check

ajeddeloh:
Here strings will expand variables when used. This escapes them so they
are expanded when bash is run rather than when cork creates the bashrc


ajeddeloh:
I agree; didn't know you could disable it like that.

PR #671, opened 2017-08-11
kola/tests: add test for restarting containerd

ajeddeloh:
Add test to kill containerd and ensure it restarts. Refactor docker info
logic to be common between docker info test and this test.

Depends on: https://github.com/coreos/coreos-overlay/pull/2703

### DIFF
--- a/kola/tests/docker/docker.go
+++ b/kola/tests/docker/docker.go
@@ -27,12 +27,33 @@ import (
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/net/context"
 
+	"github.com/coreos/go-semver/semver"
+
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
 	"github.com/coreos/mantle/lang/worker"
 	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/platform/conf"
+	"github.com/coreos/mantle/util"
 )
+
+type simplifiedDockerInfo struct {
+	ServerVersion string
+	Driver        string
+	CgroupDriver  string
+	Runtimes      map[string]struct {
+		Path string `json:"path"`
+	}
+	ContainerdCommit struct {
+		ID       string
+		Expected string
+	}
+	RuncCommit struct {
+		ID       string
+		Expected string
+	}
+	SecurityOptions []string
+}
 
 func init() {
 	register.Register(&register.Test{
@@ -169,6 +190,18 @@ systemd:
 
       [Install]
       WantedBy=multi-user.target`),
+	})
+	register.Register(&register.Test{
+		// Ensure containerd gets back up when it dies
+		Name:        "docker.containerd-restart",
+		Run:         dockerContainerdRestart,
+		ClusterSize: 1,
+		MinVersion:  semver.Version{Major: 1520},
+		UserData: conf.ContainerLinuxConfig(`
+systemd:
+  units:
+   - name: docker.service
+     enable: true`),
 	})
 }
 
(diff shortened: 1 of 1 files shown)


kola/tests/docker/docker.go
```diff
@@ -459,17 +470,75 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+	}
+	if string(pid) == "" {
+		c.Fatalf("Could not find containerd pid")
+	}
+
+	if _, err = m.SSH("sudo kill " + string(pid)); err != nil {
+		c.Fatal(err)
+	}
+
+	if testContainerdUp(c) {
```
euank: This seems like it would be racy since systemd could have already restarted it, right?

Perhaps checking if the pid changes would be a more robust way to make sure this test isn't giving a false positive?

kola/tests/docker/docker.go
```diff
@@ -459,17 +470,75 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+		c.Fatalf("containerd did not die when killed")
+	}
+
+	// verify it came back
+	if !testContainerdUp(c) {
```
euank: This test should perhaps also check that `systemctl is-active containerd` succeeds since it's testing that containerd was restarted by systemd.

kola/tests/docker/docker.go
```diff
@@ -459,17 +470,75 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+// kills it, the tests that it came back up.
+func dockerContainerdRestart(c cluster.TestCluster) {
+	m := c.Machines()[0]
+	// verify containerd came up
+	if !testContainerdUp(c) {
+		c.Fatalf("Containerd did not start")
+	}
+
+	// kill it and verify it went down
+	pid, err := m.SSH("pgrep containerd")
```
euank: `systemctl show containerd -p MainPID --value` is arguably more robust.

I'm personally always wary of `pgrep`.

ajeddeloh:
Added sugguestions, using systemctl info again instead of is-active (kill two birds with one stone). Also added a delay between killing and checking if it's up to avoid races.

kola/tests/docker/docker.go
```diff
@@ -170,6 +170,17 @@ systemd:
...
+		// Ensure containerd gets back up when it dies
+		Name:        "docker.containerd-restart",
+		Run:         dockerContainerdRestart,
+		ClusterSize: 1,
+		UserData: conf.ContainerLinuxConfig(`
+systemd:
+  units:
+   - name: docker.service
+     enable: true`),
+	})
```
bgilbert: You'll want to set a `MinVersion` to avoid uninteresting failures.

kola/tests/docker/docker.go
```diff
@@ -459,17 +470,86 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+	if err != nil {
+		c.Fatal(err)
+	}
+	if string(newPid) == string(pid) {
+		c.Fatalf("Old and new pid's are the same. containerd did not die")
+	}
+}
+
+func testContainerdUp(c cluster.TestCluster) bool {
+	type containerdInfo struct {
```
bgilbert: Let's move the struct to file scope, rather than redefining parts of it inside different tests.

kola/tests/docker/docker.go
```diff
@@ -459,17 +470,86 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+		c.Fatalf("Could not find containerd pid")
+	}
+
+	// kill it
+	if _, err = m.SSH("sudo kill " + string(pid[8:])); err != nil {
+		c.Fatal(err)
+	}
+
+	// give it a chance to restart
+	if _, err = m.SSH("sleep 2"); err != nil {
```
bgilbert: This will cause test flakes on slow nodes.  It'd be better to poll the unit status until it's stable -- either running or stopped.

kola/tests/docker/docker.go
```diff
@@ -459,17 +470,86 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+// kills it, the tests that it came back up.
+func dockerContainerdRestart(c cluster.TestCluster) {
 	m := c.Machines()[0]
 
+	// verify containerd came up
+	if !testContainerdUp(c) {
+		c.Fatalf("Containerd did not start")
+	}
+
+	pid, err := m.SSH("systemctl show containerd -p MainPID")
```
bgilbert: Why not `--value`?

kola/tests/docker/docker.go
ajeddeloh: Since the change isn't in any channel yet, what should it be set to?

bgilbert: The next-next alpha, which will be 1520.  Of course you'll need to disable that field when testing.

ajeddeloh: Fixed

kola/tests/docker/docker.go
```diff
@@ -459,39 +492,98 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+
+	info, err := unmarshalDockerInfo(m)
+	if err != nil {
+		c.Fatal(err)
+	}
+
+	return info.ContainerdCommit.ID == info.ContainerdCommit.Expected
+}
+
+func unmarshalDockerInfo(m platform.Machine) (simplifiedDockerInfo, error) {
```
euank: this is a little bikesheddy, but I think this method name is misleading. `unmarshal` to me makes it sound like it's idempotent / not doing network traffic.

Maybe `retrieveDockerInfo` or `queryDockerInfo`?

kola/tests/docker/docker.go
```diff
@@ -459,39 +492,98 @@ func dockerUserNoCaps(c cluster.TestCluster) {
...
+	}
+
+	// verify it came back and docker knows about it
+	if !testContainerdUp(c) {
+		c.Fatalf("Docker could not find containerd")
+	}
+
+}
+
+func testContainerdUp(c cluster.TestCluster) bool {
```
euank: More bikeshedding; this doesn't have to return a bool now that there's no attempt to observe containerd not being up.

It can just `c.Fatalf` itself if the commits don't match. Arguably `assert` is clearer than `test` when the helper is doing that, but I know I already started the pattern of using `test` in this file.

ajeddeloh:
Fixed @euank's nits. I don't have write access

PR #672, opened 2017-08-11
kola/tests/misc: add basic test for sssd

dgonyeo:
Adds a test for sssd wherein openldap is run in a container and sssd is
configured to get users from ldap. The id binary is then run to test if
the users exist on the system.

cc @brianredbeard if he has time to check that I'm testing sssd in a practical way

### DIFF
--- a/kola/tests/misc/sssd.go
+++ b/kola/tests/misc/sssd.go
@@ -1,0 +1,168 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package misc
+
+import (
+	"time"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform/conf"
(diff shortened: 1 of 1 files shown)


brianredbeard:
Looks good, though that openldap container seems like a random choice.

If you're going to add both nss and pam you should check both services.
This could mean checking both with getent and another option.

Have you ensured that while the tests are running that the uid pminsky will
not be a valid login with the default password?

kola/tests/misc/sssd.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform/conf"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         SssdId,
+		ClusterSize: 1,
+		Name:        "coreos.sssd.id",
```
bgilbert: `coreos.auth.sssd.id`?

kola/tests/misc/sssd.go
```diff
@@ -0,0 +1,168 @@
...
+	//   up user information
+	// - run `id pminsky` to see if the `pminsky` user exists. If it does, then
+	//   sssd successfully looked up a user from ldap.
+	//
+	// If the test runs for more than 5 minutes, fail.
+	for {
+		var output []byte
+		var err error
+
+		time.Sleep(time.Second * 30)
```
bgilbert: `util.Retry()` would be good for this.

dgonyeo:
The openldap container was chosen because it was the most popular one on the docker hub by an order of magnitude and had pretty extensive docs for using it.

Would `getent passwd pminsky` test nss and `sudo su pminsky -c true` test pam successfully?

dgonyeo:
For some reason I'm having problems running `su pminsky` in the VM. I'll keep investigating that.

dgonyeo:
While this test would be lovely to have, I was never able to get it to a state where I was confident it was testing the right things. If someone else agrees this test would be good, feel free to open a new PR with these changes (and verify it's testing something useful!).

PR #673, opened 2017-08-12
aws: temporarily use image format raw

euank:
VMDK seems to be having issues


euank:
Closing, this doesn't really make sense to merge and after a little playing around it's unclear whether raw uploads are completely safe / sensical to do here.

PR #674, opened 2017-08-14
kola/test/misc/raid.go: add coreos.disk.raid.data

dgonyeo:
Adds a test that raids together the OEM-CONFIG and USR-B partitions and mounts them at `/var/lib/data`. This test does not test if the raid array still functions after a reboot, therefore `/usr/share/oem/grub.cfg` doesn't need to be overwritten and it should be able to run on all platforms.

### DIFF
--- a/kola/tests/misc/raid.go
+++ b/kola/tests/misc/raid.go
@@ -61,6 +61,39 @@ func init() {
         inline: |
             set linux_append="rd.auto"`),
 	})
+	register.Register(&register.Test{
+		Run:         DataOnRaid,
+		ClusterSize: 1,
+		Name:        "coreos.disk.raid.data",
+		UserData: conf.ContainerLinuxConfig(`storage:
+  raid:
+    - name: "DATA"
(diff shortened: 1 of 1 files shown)


dgonyeo:
After poking some more, I discovered that raid arrays are assembled at some point after root is mounted and before systemd attempts to run mount units, so this test can reboot and check that the array is still mounted just fine. I've updated this PR to do so.

kola/tests/misc/raid.go
```diff
@@ -77,6 +110,22 @@ func RootOnRaid(c cluster.TestCluster) {
 	checkIfMountpointIsRaid(c, m, "/")
 }
 
+// FIXME(dgonyeo): once `rd.auto` is set by default this test should reboot and
```
bgilbert: stale

dgonyeo: fixed

dgonyeo:
Good to merge, lack access to do so

PR #675, opened 2017-08-14
cmd/kola: Add short options for --platform and --parallel

bgilbert:


bgilbert:
I use those options frequently and would prefer to type less.

ajeddeloh:
LGTM. I would also appreciate these short options

arithx:
LGTM

PR #680, opened 2017-08-17
readme: update to be accurate

ajeddeloh:
Do no merge, still working on it

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,46 +1,102 @@
 # Mantle: Gluing CoreOS together
 
-This repository is a collection of utilities for the CoreOS SDK.
+This repository is a collection of utilities for developing Container Linux. Most of the
+tools are for uploading, running, and interacting with Container Linux instances running
+locally or in a cloud.
 
-## plume
+## Overview
+Mantle is composed of many utilities:
+ - `cork` for handling the Container Linux SDK
+ - `gangue` for downloading from Google Storage
+ - `kola` for launching instances and running tests
+ - `kolet` an agent for kola that runs on instances
+ - `ore` for interfacing with cloud providers
+ - `plume` for releasing Container Linux
+All of the utilities support the `help` command to get a full listing of their subcommands
+and options.
 
-CoreOS release utility
+## cork
+Cork is a tool that helps working with Container Linux images and the SDK.
 
-## plume index
+### cork create
+Download and unpack the Container Linux SDK.
 
-Generate and upload index.html objects to turn a Google Cloud Storage
-bucket into a publicly browsable file tree. Useful if you want something
-like Apache's directory index for your software download repository.
+`cork create`
+
+### cork enter
+Enter the SDK chroot, and optionally run a command. The command and its
+arguments can be given after `--`.
+
+`cork enter -- repo sync`
+
+### cork download-image
+Download a Container Linux image into `$PWD/.cache/images`.
+
+`cork download-image --platform=qemu`
+
+### Building Container Linux with cork
+See [Modifying CoreOS](https://coreos.com/os/docs/latest/sdk-modifying-coreos.html) for
+an example of using cork to build a Container Linux image.
+
+## gangue
+Gangue is a tool for downloading and verifying files from Google Storage with authenticated requests.
+It is primarily used by the SDK.
+
+### gangue get
+Get a file from Google Storage and verify it using GPG.
 
 ## kola
-
-Kola is a framework for testing software integration in CoreOS instances
+Kola is a framework for testing software integration in Container Linux instances
 across multiple platforms. It is primarily designed to operate within
-the CoreOS SDK for testing software that has landed in the OS image.
+the Container Linux SDK for testing software that has landed in the OS image.
 Ideally, all software needed for a test should be included by building
 it into the image from the SDK.
 
 Kola supports running tests on multiple platforms, currently QEMU, GCE,
-and AWS. In the future systemd-nspawn and other platforms may be added.
+AWS, VMware VSphere, and Packet. In the future systemd-nspawn and other
+platforms may be added.
 Local platforms do not rely on access to the Internet as a design
-principal of kola, minimizing external dependencies. Any network
+principle of kola, minimizing external dependencies. Any network
 services required get built directly into kola itself. Machines on cloud
 platforms do not have direct access to the kola so tests may depend on
 Internet services such as discovery.etcd.io or quay.io instead.
 
 Kola outputs assorted logs and test data to `_kola_temp` for later
-inspection. This directory is completely wiped and recreated every time.
+inspection.
 
 Kola is still under heavy development and it is expected that its
-interface will continue to change. Both the CLI and test registration
-interface have upcoming changes.
+interface will continue to change.
+
+By default, kola uses the `qemu` platform with the most recently built image
+(assuming it is run from within the SDK).
 
 ### kola run
-The run command invokes the main kola test harness. The harness will
-run any registered tests on all platforms unless otherwise specified. It
+The run command invokes the main kola test harness. It
 runs any tests whose registered names matches a glob pattern.
 
 `kola run <glob pattern>`
+
+### kola list
+The list command lists all of the available tests.
+
+### kola spawn
+The spawn command launches Container Linux instances. Use `-s` to open a shell on
+the spawned instance.
+
+### kola mkimage
+The mkimage command creates a copy of the input image with its primary console set
+to the serial port (/dev/ttyS0). This causes more output to be logged on the console,
+which is also logged in `_kola_temp`. This can only be used with QEMU images and must
+be used with the `coreos_*_image.bin` image, *not* the `coreos_*_qemu_image.img`.
+
+### kola bootchart
+The bootchart command launches an instance then generates an svg of the boot process
+using `systemd-analyze`.
+
+### kola updatepayload
+The updatepayload command launches a Container Linux instance then updates it by
+sending an update to its update_engine. The update is the `coreos_*_update.gz` in the
+latest build directory.
 
 ### kola test registration
 Registering kola tests currently requires that the tests are registered
@@ -50,15 +106,18 @@ the mantle codebase.
 Groups of similar tests are registered in an init() function inside the
 kola package.  `Register(*Test)` is called per test. A kola `Test`
 struct requires a unique name, and a single function that is the entry
-point into the test.
+point into the test. Additionally, userdata (such as a Container Linux
+Config) can be be supplied. See the `Test` struct in
+[kola/register/register.go](https://github.com/coreos/mantle/tree/master/kola/register/register.go)
+for a complete list of options.
 
 ### kola test writing
 A kola test is a go function that is passed a `platform.TestCluster` to
-run code against.  Its signature is `func(platform.TestCluster) error`
+run code against.  Its signature is `func(platform.TestCluster)`
 and must be registered and built into the kola binary. 
 
 A `TestCluster` implements the `platform.Cluster` interface and will
-give you access to a running cluster of CoreOS machines. A test writer
+give you access to a running cluster of Container Linux machines. A test writer
 can interact with these machines through this interface.
 
 To see test examples look under
@@ -67,7 +126,7 @@ mantle codebase.
 
 ### kola native code
 For some tests, the `Cluster` interface is limited and it is desirable to
-run native go code directly on one of the CoreOS machines. This is
+run native go code directly on one of the Container Linux machines. This is
 currently possible by using the `NativeFuncs` field of a kola `Test`
 struct. This like a limited RPC interface.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -148,44 +134,20 @@ Download a CoreOS image into `$PWD/.cache/images`.
...
 
-# build dev and prod QEMU images
-cork enter -- ${S}/image_to_vm.sh --board=${B}
-cork enter -- ${S}/image_to_vm.sh --board=${B} --prod_image
-```
+Generate and upload index.html objects to turn a Google Cloud Storage
+bucket into a publicly browsable file tree. Useful if you want something
+like Apache's directory index for your software download repository.
 
 ## TODO
```
ajeddeloh: How much of this TODO is still accurate?

README.md
```diff
@@ -82,48 +93,23 @@ For more examples, look at the
...
-### ore list-instances
-
-List running gce instances. Common usage:
-
-`ore list-instances`
+## ore
 
-### ore destroy-instances
+Related commands to launch instances on a variety of plaforms (gcloud, aws,
+azure, esx, and packet) within the latest SDK image. ore is mimics the underlying
```
euank: nit, `ore is` is a typo. I'll suggest `ore provides a low level interface for each cloud provider`

README.md
euank: I think 3 is close-enough-to-done and I don't think we care about 1 all that much.

README.md
```diff
@@ -21,27 +13,44 @@ Ideally, all software needed for a test should be included by building
 it into the image from the SDK.
 
 Kola supports running tests on multiple platforms, currently QEMU, GCE,
-and AWS. In the future systemd-nspawn and other platforms may be added.
+AWS, vmware ESX and Packet. In the future systemd-nspawn and other
```
bgilbert: `VMware`

README.md
```diff
@@ -21,27 +13,44 @@ Ideally, all software needed for a test should be included by building
...
+`kola list`
+
+### kola spawn
+The spawn command launches CoreOS instances. Use `-s` to login to the
+instance spawned.
+
+### kola mkimage
+The mkimage command sets the linux serial console to /dev/ttyS0. This
+causes more output to be logged on the console, which is also logged
+in `_kola_tmp`.
```
bgilbert: `_kola_temp`

README.md
```diff
@@ -21,27 +13,44 @@ Ideally, all software needed for a test should be included by building
...
+To get a list of all tests, run
+
+`kola list`
+
+### kola spawn
+The spawn command launches CoreOS instances. Use `-s` to login to the
+instance spawned.
+
+### kola mkimage
+The mkimage command sets the linux serial console to /dev/ttyS0. This
```
bgilbert: "...creates a copy of a `coreos_*_image.bin` with its primary console set to the serial port."

README.md
bgilbert: I tend to think TODOs belong in the issue tracker anyway, so I'd be +1 to dropping this.

ajeddeloh:
Updated to address issues, fix some bad wording. Removed the TODO section

README.md
```diff
@@ -1,47 +1,56 @@
...
+Kola is a framework for testing software integration in Container Linux instances
 across multiple platforms. It is primarily designed to operate within
-the CoreOS SDK for testing software that has landed in the OS image.
+the Container Linux SDK for testing software that has landed in the OS image.
 Ideally, all software needed for a test should be included by building
 it into the image from the SDK.
 
 Kola supports running tests on multiple platforms, currently QEMU, GCE,
-and AWS. In the future systemd-nspawn and other platforms may be added.
+AWS, VMware ESX and Packet. In the future systemd-nspawn and other
```
arithx: `VMware, ESX, and Packet`

README.md
```diff
@@ -1,47 +1,56 @@
...
-## plume
-
-CoreOS release utility
-
-## plume index
-
-Generate and upload index.html objects to turn a Google Cloud Storage
-bucket into a publicly browsable file tree. Useful if you want something
-like Apache's directory index for your software download repository.
+This repository is a collection of utilities for the Container Linux SDK. Most of the
```
dm0-: I'm not requesting changing this now since it was in the original text, but in my opinion the phrasing that these tools are "for the SDK" is wrong.  Most of these tools have little-to-nothing to do with the SDK (`cork` being the big exception), and our release process runs them all outside the SDK (except `kola` for QEMU).

Also, a newer program `gangue` is not described in the document.  It is available in the SDK and developer images as a portage URL handler to allow cryptographic verification of Gentoo binary packages.  Not sure if that warrants inclusion in the README or not.

ajeddeloh: Maybe "for developing Container Linux"? I'm all for fixing it now since most of the original readme was out of date/no longer accurate. 

I'm not familair with gangue. Where all is it used (so I can see how it's used and write a description that makes sense)?

dm0-: I think "for developing" sounds better.

Right now, the release process is the only place that uses `gangue` by default: https://github.com/coreos/scripts/blob/master/jenkins/images.sh#L14,L17

I've enabled it in my own SDK by writing the following to `/etc/env.d/99gangue` and running `env-update`:

```sh
COREOS_DEV_BUILDS='gs://builds.developer.core-os.net'
FETCHCOMMAND_GS='/usr/bin/gangue get --json-key=/dev/null "${URI}" "${DISTDIR}/${FILE}"'
RESUMECOMMAND_GS='/usr/bin/gangue get --json-key=/dev/null "${URI}" "${DISTDIR}/${FILE}"'
```

The `/dev/null` causes it to fall back to my user's authorization, but that path normally takes service account credentials.

README.md
bgilbert: Disagree.  We don't test on Workstation or Fusion, so I think we should call out ESX specifically.

arithx: We should probably call it out directly as `VMware vSphere` then, the current `kola` platform requires clone which isn't available in the free-ware version of the hypervisor, the unlocked version is `vSphere` (even though it's still an `ESX` hypervisor)

README.md
```diff
@@ -50,15 +59,17 @@ the mantle codebase.
 Groups of similar tests are registered in an init() function inside the
 kola package.  `Register(*Test)` is called per test. A kola `Test`
 struct requires a unique name, and a single function that is the entry
-point into the test.
+point into the test. Additionally, userdata (such as an container linux
```
bgilbert: `a Container Linux Config`

README.md
```diff
@@ -1,47 +1,56 @@
...
+To get a list of all tests, run
+
+`kola list`
+
+### kola spawn
+The spawn command launches Container Linux instances. Use `-s` to open a shell on
+the spawned instance.
+
+### kola mkimage
+The mkimage command creates a copy of the input image with its primary console set
```
bgilbert: It's important to point out that `kola mkimage` is only for QEMU images, and specifically `coreos_*_image.bin` (rather than `coreos_*_qemu_image.img`).

README.md
```diff
@@ -1,47 +1,56 @@
...
 ### kola run
 The run command invokes the main kola test harness. The harness will
 run any registered tests on all platforms unless otherwise specified. It
 runs any tests whose registered names matches a glob pattern.
 
 `kola run <glob pattern>`
 
+To get a list of all tests, run
+
+`kola list`
```
bgilbert: Should probably be a separate subhead for consistency.

README.md
```diff
@@ -50,15 +59,17 @@ the mantle codebase.
 Groups of similar tests are registered in an init() function inside the
 kola package.  `Register(*Test)` is called per test. A kola `Test`
 struct requires a unique name, and a single function that is the entry
-point into the test.
+point into the test. Additionally, userdata (such as an container linux
+config) can be be supplied. See the `Test` struct in [kola/register.go] for a
```
bgilbert: Unrendered link.

README.md
```diff
@@ -67,71 +78,47 @@ mantle codebase.
...
+each provider.
 
-Destroy instances on gce based by prefix or name. Images created with
-`create-instances` use a common basename as a prefix that can also be
-used to tear down the cluster. Common usage:
+Note, when uploading to some cloud providers (e.g. gce) the image may need to be packaged
+with a different --format (e.g. --format=gce) when running `image_to_vm.sh`
 
-`ore destroy-instances -prefix=$USER`
+SSH keys should be added to the project metadata before launching a cluster. All commands
```
bgilbert: Not all platforms support this, and even on platforms that do, it's probably better to pass userdata that configures an SSH key.

README.md
```diff
@@ -1,39 +1,43 @@
...
+This repository is a collection of utilities for developing Container Linux. Most of the
+tools are for uploading, running, and interacting with Container Linux instances running
+locally or in a cloud.
+
+## Overview
+Mantle is composed of many utilies:
+ - `kola` for launching instances and running tests
+ - `ore` for interfacing with cloud providers
+ - `cork` for handling the Container Linux SDK
+ - `gangue` for downloading from google storage
```
dm0-: Google Storage should probably be capitalized here, as it is below.

Also it seems a bit odd that this list is in a different order than the command sections.

README.md
```diff
@@ -144,54 +147,29 @@ arguments can be given after `--`.
...
-   `destructor` package.
- - Continue to expand the amount of data recorded in `_kola_temp`.
+### Building Container Linux with cork
+
+See [Modifying CoreOS](https://coreos.com/os/docs/latest/sdk-modifying-coreos.html) for
+an example of using cork to build a Container Linux image.
+
+## plume
+
+CoreOS release utility
```
dm0-: Every other section uses complete sentences here.

README.md
```diff
@@ -1,47 +1,72 @@
...
+This repository is a collection of utilities for developing Container Linux. Most of the
+tools are for uploading, running, and interacting with Container Linux instances running
+locally or in a cloud.
+
+## Overview
+Mantle is composed of many utilities:
+ - `kola` for launching instances and running tests
+ - `ore` for interfacing with cloud providers
+ - `cork` for handling the Container Linux SDK
+ - `gangue` for downloading from google storage
```
dm0-: Google Storage should probably be capitalized here, as it is below.

Also, it seems a bit odd that this list is sorted differently than the command sections below.

README.md
```diff
@@ -144,54 +144,36 @@ arguments can be given after `--`.
...
-   asynchronous jobs. It is more standard and flexible than our local
-   `destructor` package.
- - Continue to expand the amount of data recorded in `_kola_temp`.
+### Building Container Linux with cork
+
+See [Modifying CoreOS](https://coreos.com/os/docs/latest/sdk-modifying-coreos.html) for
+an example of using cork to build a Container Linux image.
+
+## plume
+Container Linux release utility. Releases are done in two stages, each with their own command: pre-release
```
dm0-: Not sure if it matters, but this doesn't begin with a complete sentence like the other sections.

README.md
```diff
@@ -144,54 +144,36 @@ arguments can be given after `--`.
...
+Generate and upload index.html objects to turn a Google Cloud Storage
+bucket into a publicly browsable file tree. Useful if you want something
+like Apache's directory index for your software download repository.
+
+## gangue
+Gangue is a tool for downloading and verifying files from Google Storage. It is primarily
+used by the SDK. 
+
+### gangue get
+Get a file from Google Storage and verify it using GPG.
```
dm0-: Maybe it's implicit, but it makes authenticated requests to Google Storage.  This is what allows our embargoed builds to work, so it might be worth mentioning.

ajeddeloh:
Updated for the latest round of feedback.

README.md
```diff
@@ -67,131 +123,55 @@ mantle codebase.
...
-`cork enter -- repo sync`
+## plume
+Plume is the Container Linux release utility. Releases are done in two stages,
+each with their own command: pre-release and release. Both of these commands are idempotent.
 
-### cork download-image
+### plume pre-release
+The pre-release command does as much of the release process as possible without making anything public.
+This includes uploading images to cloud providers (except those like gce which don't allow us to upload
+images without making them public) and running kola tests against the uploaded images.
```
bgilbert: plume doesn't run kola tests.  Jenkins does, as part of the pre-release pipeline.  Just drop the latter part.

README.md
```diff
@@ -67,131 +123,55 @@ mantle codebase.
...
 
-Download a CoreOS image into `$PWD/.cache/images`.
+## plume release
+Publish a new Container Linux release. This makes the images uploaded by pre-release public and uploads
+images that pre-release could not.
```
bgilbert: It also copies release artifacts to public storage buckets and updates the buckets' directory indexes.

README.md
```diff
@@ -67,131 +123,55 @@ mantle codebase.
...
-Download a CoreOS image into `$PWD/.cache/images`.
+## plume release
+Publish a new Container Linux release. This makes the images uploaded by pre-release public and uploads
+images that pre-release could not. 
 
-`cork download-image --platform=qemu`
+### plume index
+Generate and upload index.html objects to turn a Google Cloud Storage
+bucket into a publicly browsable file tree. Useful if you want something
+like Apache's directory index for your software download repository.
```
bgilbert: Doesn't need to be run as part of the release process, since this is handled by `plume release`.

README.md
```diff
@@ -67,131 +123,55 @@ mantle codebase.
...
-
-`cork create`
-
-### cork enter
+## ore
+Ore provides a low level interface for each cloud provider. It has commands 
+related to launching instances on a variety of platforms (gcloud, aws,
+azure, esx, and packet) within the latest SDK image. Ore mimics the underlying
+api for each cloud provider closely, so the interface for each cloud provider
+is different. `ore help <provider>` lists the available actions and options for
```
bgilbert: The help text is often fairly useful, so we should probably hoist this advice to the overview section.

PR #684, opened 2017-08-22
kola/tests: remove root-on-raid workaround

ajeddeloh:
bootengine now handles finding and assembling any raid device that ends
with -hasroot. The rd.auto workaround is no longer necessary.

Depends on https://github.com/coreos/bootengine/pull/130

### DIFF
--- a/kola/tests/misc/raid.go
+++ b/kola/tests/misc/raid.go
@@ -17,6 +17,8 @@ package misc
 import (
 	"encoding/json"
 
+	"github.com/coreos/go-semver/semver"
+
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
 	"github.com/coreos/mantle/platform"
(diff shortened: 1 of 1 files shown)


kola/tests/misc/raid.go
```diff
@@ -28,38 +28,23 @@ func init() {
 		Run:         RootOnRaid,
 		ClusterSize: 1,
 		Name:        "coreos.disk.raid.root",
-		// FIXME: This can only work on qemu, since it's overwriting
-		// /usr/share/oem/grub.cfg. The setting being appended (set
-		// linux_append="rd.auto") should probably be the default, and this
-		// should be removed after the OS level fix is made.
-		// https://github.com/coreos/bugs/issues/2099
```
bgilbert: Remember to drop `Platforms`, not just the comment saying that it should be dropped.

Also add a `MinVersion`.

ajeddeloh:
Was failing in CI due to no `MinVersion` and the dependant change to bootengine not being done yet.

PR #685, opened 2017-08-23
platform: add cleanup of serial file and vm folders to esx

arithx:
Currently we leak a file & folder on the datastore. This adds the
removal of these items to the API and is called inside of `Destroy`
after the serial file is downloaded.

### DIFF
--- a/platform/api/esx/api.go
+++ b/platform/api/esx/api.go
@@ -24,6 +24,7 @@ import (
 	"net"
 	"net/url"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/coreos/pkg/capnslog"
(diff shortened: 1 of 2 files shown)


platform/api/esx/api.go
```diff
@@ -282,6 +282,32 @@ func (a *API) GetConsoleOutput(name string) (string, error) {
 	return string(buf), nil
 }
 
+func (a *API) CleanupFiles(name string) error {
```
bgilbert: The method name is correct from an implementation perspective, but from the perspective of the rest of mantle, its place in the machine lifecycle is a bit unclear: CreateDevice -> TerminateDevice -> CleanupFiles.  What about calling it CleanupDevice?

Is it worth adding a check that the device has already been terminated?

arithx: For safety purposes it's probably worth implementing the check if it's ever used outside of the `platform/machine/cluster` implementation.

I'll rename it to `CleanupDevice` as it requires the context of a VM

platform/api/esx/api.go
```diff
@@ -282,6 +282,34 @@ func (a *API) GetConsoleOutput(name string) (string, error) {
...
+	if err == nil {
+		return fmt.Errorf("VM still exists")
+	}
+
+	// Remove the serial.out file
+	fm := defaults.datastore.NewFileManager(defaults.datacenter, true)
+	uri := fmt.Sprintf("%s/serial.out", name)
+	err = fm.DeleteFile(a.ctx, uri)
+	if err != nil {
+		return fmt.Errorf("couldn't delete serial.out: %v", err)
```
euank: Ideally both of this and the next delete would return `nil` if the file didn't exist (because e.g. an administrator deleted it).

Not sure if the filemanager has a thing like `os.IsNotExist` to check for.

arithx: As discussed OOB, the way to check files involves using a `Stat` method on the datastore object which we'd like to avoid.

Pushing an update to add an additional check before we print the error message that `was not found` is the suffix for the error string (they don't return an `os.PathError` but rather a string).

arithx:
Good to merge from my end.

PR #687, opened 2017-08-24
cmd/kola: Document special handling of single test name by kola run

bgilbert:
Wording suggestions welcome.

### DIFF
--- a/cmd/kola/kola.go
+++ b/cmd/kola/kola.go
@@ -43,9 +43,14 @@ var (
 	}
 
 	cmdRun = &cobra.Command{
-		Use:    "run [glob pattern]",
-		Short:  "Run kola tests by category",
-		Long:   "run all kola tests (default) or related groups",
+		Use:   "run [glob pattern]",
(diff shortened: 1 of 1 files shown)


cmd/kola/kola.go
```diff
@@ -45,7 +45,12 @@ var (
 	cmdRun = &cobra.Command{
 		Use:    "run [glob pattern]",
 		Short:  "Run kola tests by category",
-		Long:   "run all kola tests (default) or related groups",
+		Long:   `Run all kola tests (default) or related groups.
+
+If the glob pattern is exactly equal to the name of a single test, any
+restrictions on the versions of Container Linux supported by that test
+will be ignored.
```
arithx: Is it only the `MinVersion` & `MaxVersion` traits that are ignored or will it also ignore the `ExcludedPlatform` trait?

bgilbert: Just the version properties.

bgilbert:
Updated to delete one column of whitespace.

PR #688, opened 2017-08-24
kola/tests: add test that ipvsadm works

ajeddeloh:
For https://github.com/coreos/coreos-overlay/pull/2707

### DIFF
--- a/kola/tests/packages/ipvsadm.go
+++ b/kola/tests/packages/ipvsadm.go
@@ -1,0 +1,79 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package packages
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/coreos/mantle/kola/cluster"
+)
+
+func ipvsadm(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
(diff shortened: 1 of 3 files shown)


kola/tests/packages/ipvsadm.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         ipvsadm,
+		ClusterSize: 1,
+		MinVersion:  semver.Version{Major: 1520},
```
ajeddeloh: This will need to be changed if https://github.com/coreos/coreos-overlay/pull/2707 and https://github.com/coreos/portage-stable/pull/575 don't make it in today

arithx:
I think that structuring them as sub-tests would be fine from my end provided we keep any potentially destructive tests/binaries in their own test sets.

euank:
👍 for subtests with the longer term goal of letting the framework deal with.

ajeddeloh:
By subtests you mean have one overarching test that runs a bunch of functions which are each their own test but not registered with kola (so we keep the number of registered tests down), yes?

arithx:
ajeddeloh: see coreos.basic for an example

They are registered within kola and run on the same virtual machine but still function as a `test` in that a failure in one won't stop others from executing.

euank:
I specifically did mean the `c.Run(c cluster.TestCluster)` style subtest support that kola now has.

A specific example of that is the docker base tests: https://github.com/coreos/mantle/blob/934eaad14ac5687684d0296582a3a1287e38b4b0/kola/tests/docker/docker.go#L222-L229

dm0-:
Note that `coreos.basic` is written to use `kolet` to execute programs on the test system.  That won't be needed here.

You could probably do something like set up a registered test function like https://github.com/coreos/mantle/blob/master/kola/tests/docker/docker.go#L222,L229 in a `main.go` file, then each package can have its own file containing its subtest function and add a `c.Run()` line to `main.go`.

dm0-:
Yes, what @euank said, I was apparently typing that reference at the same time.

ajeddeloh:
Updated to use sub-tests

PR #689, opened 2017-08-25
platform/machine/qemu/cluster: create machine with multiple devices

arithx:
Adds the ability to create a machine with multiple disk drives
attached. Default functionality is unchanged as the original
`NewMachine` function has been renamed to `CreateMachine` which
takes an additional argument of type `QEMUOptions`. A wrapper
function now exists at `NewMachine` which calls `CreateMachine` with
an empty `QEMUOptions` struct.

This only implements a `Disk` option on the `QEMUOptions` struct,
which itself is a struct containing a `Size` string.

Whenever additional disks are present they will have a new `qcow2`
image created for them which is immediately cleaned up similar to
how we are doing for the primary disk currently.

### DIFF
--- a/cmd/kola/spawn.go
+++ b/cmd/kola/spawn.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -24,6 +25,7 @@ import (
 	"github.com/coreos/mantle/kola"
 	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/platform/conf"
+	"github.com/coreos/mantle/platform/machine/qemu"
 )
 
 var (
@@ -34,11 +36,12 @@ var (
 		Short:  "spawn a CoreOS instance",
 	}
 
-	spawnNodeCount int
-	spawnUserData  string
-	spawnShell     bool
-	spawnRemove    bool
-	spawnVerbose   bool
+	spawnNodeCount      int
+	spawnUserData       string
+	spawnShell          bool
+	spawnRemove         bool
+	spawnVerbose        bool
+	spawnMachineOptions string
 )
 
 func init() {
@@ -47,6 +50,7 @@ func init() {
 	cmdSpawn.Flags().BoolVarP(&spawnShell, "shell", "s", false, "spawn a shell in an instance before exiting")
 	cmdSpawn.Flags().BoolVarP(&spawnRemove, "remove", "r", true, "remove instances after shell exits")
 	cmdSpawn.Flags().BoolVarP(&spawnVerbose, "verbose", "v", false, "output information about spawned instances")
+	cmdSpawn.Flags().StringVar(&spawnMachineOptions, "qemu-options", "", "experimental: path to QEMU machine options json")
 	root.AddCommand(cmdSpawn)
 }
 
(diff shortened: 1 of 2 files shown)


platform/machine/qemu/cluster.go
```diff
@@ -203,6 +220,21 @@ func (qc *Cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error)
...
+		additionalDiskFile, err := setupDisk(disk.Size, "")
+		if err != nil {
+			return nil, err
+		}
+		defer additionalDiskFile.Close()
+		disks = append(disks, additionalDiskFile)
+		fdnum := i + 5 // fd 4 is the primary drive
+		fdset := i + 2 // fdset 1 is used by the primary drive
+		id := fmt.Sprintf("d%d", fdnum)
+		qmCmd = append(qmCmd, "-add-fd", fmt.Sprintf("fd=%d,set=%d", fdnum, fdset),
```
bgilbert: Code duplication here is not ideal.  Maybe create a closure which allocates the fd and fdset numbers and appends the args, then call it for both the primary and secondary disks.

cmd/kola/spawn.go
```diff
@@ -87,7 +91,20 @@ func doSpawn(cmd *cobra.Command, args []string) error {
...
-		mach, err := cluster.NewMachine(userdata)
+		var mach platform.Machine
+		if kolaPlatform == "qemu" && spawnMachineOptions != "" {
+			b, err := ioutil.ReadFile(spawnMachineOptions)
+			if err != nil {
+				return fmt.Errorf("Could not read machine options: %v", err)
+			}
+
+			var machineOpts qemu.MachineOptions
+			json.Unmarshal(b, &machineOpts)
```
ajeddeloh: Check for errors otherwise this will silently fail and not apply the options

platform/machine/qemu/cluster.go
```diff
@@ -186,11 +181,23 @@ func (qc *Cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error)
 			"-device", qc.virtio("9p", "fsdev=cfg,mount_tag=config-2"))
 	}
 
+	fdnum := 3
```
ajeddeloh: nit: I would move these declarations down below the declaration of addDisk to make it clear they aren't being closed on by the closure.

cmd/kola/spawn.go
```diff
@@ -87,7 +91,23 @@ func doSpawn(cmd *cobra.Command, args []string) error {
...
+
+			var machineOpts qemu.MachineOptions
+			err = json.Unmarshal(b, &machineOpts)
+			if err != nil {
+				return fmt.Errorf("Could not unmarshal machine options: %v", err)
+			}
+
+			mach, err = cluster.(*qemu.Cluster).NewMachineWithOptions(userdata, machineOpts)
+		} else {
+			mach, err = cluster.NewMachine(userdata)
```
ajeddeloh: If you specify bad machine options (e.g. an invalid suffix) mach and err are both set to nil and it segfaults later.

arithx: Forgot to setup the err var outside the if scope. Pushed a change to fix this.

cmd/kola/spawn.go
```diff
@@ -87,7 +91,24 @@ func doSpawn(cmd *cobra.Command, args []string) error {
 
 	var someMach platform.Machine
 	for i := 0; i < spawnNodeCount; i++ {
-		mach, err := cluster.NewMachine(userdata)
+		var mach platform.Machine
+		var err error
+		if kolaPlatform == "qemu" && spawnMachineOptions != "" {
+			b, err := ioutil.ReadFile(spawnMachineOptions)
```
ajeddeloh: err is getting redefined here I think.

ajeddeloh:
Tested via it's use in https://github.com/coreos/mantle/pull/690. LGTM

arithx:
Nits fixed and repushed

PR #690, opened 2017-08-25
 kola/tests: update root on raid to use uuids

ajeddeloh:
Use new gpt partition type uuids from https://github.com/coreos/bootengine/pull/132

Depends on https://github.com/coreos/mantle/pull/689

### DIFF
--- a/kola/tests/misc/raid.go
+++ b/kola/tests/misc/raid.go
@@ -23,30 +23,61 @@ import (
 	"github.com/coreos/mantle/kola/register"
 	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/platform/conf"
+	"github.com/coreos/mantle/platform/machine/qemu"
 )
 
-func init() {
-	register.Register(&register.Test{
-		Run:         RootOnRaid,
-		ClusterSize: 1,
-		MinVersion:  semver.Version{Major: 1520},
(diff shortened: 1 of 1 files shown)


kola/tests/misc/raid.go
```diff
@@ -84,12 +117,20 @@ systemd:
 }
 
 func RootOnRaid(c cluster.TestCluster) {
-	m := c.Machines()[0]
+	options := qemu.MachineOptions{
+		AdditionalDisks: []qemu.QEMUDisk{
```
arithx: FYI: This was changed from `QEMUDIsk` to just `Disk` in the upstream PR.

ajeddeloh:
Rebased onto master, fixed QEMUDisk -> Disk

kola/tests/misc/raid.go
```diff
@@ -23,30 +23,63 @@ import (
...
             - "-L"
-            - "ROOT"`),
+            - "wasteland"
+          force: true`)
+)
+
+func init() {
+	register.Register(&register.Test{
+		// This test needs additional disks which is only supported on qemu since Ignition
+		// does not support creating partitions on disks without wiping the partition table
```
bgilbert: It does.  It just doesn't support deleting the ones that are currently filling up the disk.

PR #691, opened 2017-08-25
platform/base: Render CLCs with platform

euank:
This updates the BaseCluster to know what ctplatform it's for, and for
each of those platforms render Container Linux configs appropriately.

Platforms CT does not have specific support for pass in the empty
string, which results in them working fine for any CLCs that do not use
special variables.

These platforms can still use CLCs for userdata as long as they don't
reference special platform-specific variables.

cc @bgilbert

### DIFF
--- a/cmd/ore/packet/create-device.go
+++ b/cmd/ore/packet/create-device.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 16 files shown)


euank:
Green now; having a dirty gopath turned out to be the culprit for the glide mess.

PR #693, opened 2017-08-28
kola/etcd: add etcd-member and etcdctl tests

euank:
I promised a forthcoming test in https://github.com/coreos/coreos-overlay/pull/2720#issuecomment-324794069, so here's a shot at it.

These both pass on images built off master (tested against gce). Neither passes on the current alpha due to them depending on etcdctl v3.

### DIFF
--- a/kola/tests/etcd/discovery.go
+++ b/kola/tests/etcd/discovery.go
@@ -15,6 +15,9 @@
 package etcd
 
 import (
+	"encoding/json"
+
+	"github.com/coreos/go-semver/semver"
 	"github.com/coreos/pkg/capnslog"
 
 	"github.com/coreos/mantle/kola/cluster"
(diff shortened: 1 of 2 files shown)


kola/tests/etcd/discovery.go
```diff
@@ -72,4 +112,124 @@ func Discovery(c cluster.TestCluster) {
...
+	fi
+
+	backup_to="$(mktemp -d)"
+
+	sudo -E etcdctl snapshot save "${backup_to}/snapshot.db"
+	sudo -E etcdctl snapshot status "${backup_to}/snapshot.db"
+`)
+
+	if err != nil {
+		c.Fatalf("unable to put, get, and snapshot with etcdctl 3")
```
bgilbert: Should log output and `err`.

euank:
Bumped and rebased, now uses `MustSSH`; I verified it still passes on aws at least.

PR #697, opened 2017-08-30
kola/harness: add nondestructive flag for tests

arithx:
Adds a NonDestructive flag for tests similar to Min/MaxVersion
flags which allows a test to denote that it is safe to run on the
same machine as other non-destructive tests. Currently for a test
to properly be non-destructive they must leave the machine in the
exact state that they received it in, even in the case of an error
during the test.

These tests can still be filtered down as usual, they will just be
ran on a cluster that is shared with any other tests which require
the same cluster setup.

In the `_kola_tmp` output a symlink is created for each test which
points to the output directory for the cluster that it ran on.

### DIFF
--- a/harness/harness.go
+++ b/harness/harness.go
@@ -298,6 +298,23 @@ func (h *H) mkOutputDir() (dir string, err error) {
 	return
 }
 
+// Used by Non-Destructive Tests to create a symlink from the test name
+// to the cluster output directory. Only test frameworks should ever
+// use this.
+func (h *H) LinkOutputDir() (dir string, err error) {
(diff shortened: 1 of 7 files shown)


cgwalters:
I think this is a good idea, but it'd be a lot more reliable if we *enforced* the nondestructive nature.  For example, by running as a systemd unit with `DynamicUser=yes` that didn't have any additional privileges.

(I'd also like to make it more natural to use kola to run fully unprivileged containers which mount the host)

arithx:
Closing as this is extremely old and would need to be re-written.

PR #698, opened 2017-08-30
sdk: Retry GCS API failures same as downloads

dm0-:
This should fix most build failures caused by GCS, since bad HTTP responses seem to be random incidents.

### DIFF
--- a/sdk/download.go
+++ b/sdk/download.go
@@ -60,7 +60,7 @@ func DownloadFile(file, fileURL string, client *http.Client) error {
 	if err != nil {
 		return err
 	}
-	if parseURL.Scheme == "gs" {
+	resolveGS := func() error {
 		if client == nil {
 			client = http.DefaultClient
 		}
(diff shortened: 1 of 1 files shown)


cmd/gangue/gangue.go
```diff
@@ -126,13 +130,15 @@ func run(cmd *cobra.Command, args []string) {
 	}
 
 	// Download the file and verify it (unless disabled)
-	if verify {
-		err = sdk.UpdateSignedFile(output, source, client, gpgKeyFile)
-		if err == nil && !keepSig {
-			err = os.Remove(output + ".sig")
+	for i, err = 0, fmt.Errorf("dummy"); i <= retry && err != nil; i++ {
```
euank: Any reason not to use the `util.Retry` helper?

dm0-: I don't think its wait time is useful here.  The HTTP failure responses seem to happen one at a time, not through some period of downtime.

arithx: Would it not make sense to just use a duration of 0 in that case?

dm0-:
Looking into it further, I think I'd prefer this approach.  I believe the backend failures originated at this API call, and retrying here also protects other commands (e.g. when `cork` downloads the SDK).

PR #699, opened 2017-08-31
kola/tests: add nested partitions and raid test

ajeddeloh:
Depends on https://github.com/coreos/ignition/pull/434


kola/tests/misc/raid.go
```diff
@@ -80,6 +152,18 @@ func init() {
...
 	register.Register(&register.Test{
+		// This test needs additional disks which is only supported on qemu since Ignition
+		// does not support creating partitions on disks without wiping the partition table
+		// first.
+		// TODO(ajeddeloh): change this to create partitions 9 and 10 once Ignition supports
+		// creating them
+		Run:         NestedRootOnRaid,
+		ClusterSize: 0,
+		Platforms:   []string{"qemu"},
+		MinVersion:  semver.Version{Major: 1520},
```
bgilbert: Will need to be updated, of course.

kola/tests/misc/raid.go
```diff
@@ -64,6 +64,78 @@ var (
             - "-L"
             - "wasteland"
           force: true`)
+	nestedRaidRootUserData = conf.ContainerLinuxConfig(`storage:
+  disks:
+    - device: /dev/vdb
+      wipe_table: true
```
bgilbert: None of the `wipe_table`s should be necessary here.

kola/tests/misc/raid.go
```diff
@@ -80,6 +152,18 @@ func init() {
 		Name:        "coreos.disk.raid.root",
 	})
 	register.Register(&register.Test{
+		// This test needs additional disks which is only supported on qemu since Ignition
+		// does not support creating partitions on disks without wiping the partition table
```
bgilbert: stale copypaste

kola/tests/misc/raid.go
ajeddeloh: 1534 good? It's dependent on Ignition changes (which will require a ignition bump in the overlay).

bgilbert: Seems like a reasonable guess.  If it turns out to be wrong, the test will start failing and we'll bump again.

PR #702, opened 2017-09-07
cork: allow repo syncing forcibly

euank:
This is helpful for, e.g., automated builders which are fine with
discarding git repos.

### DIFF
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -52,6 +52,7 @@ var (
 
 	// only for `update` command
 	allowCreate      bool
+	forceSync        bool
 	downgradeInPlace bool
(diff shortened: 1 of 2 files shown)


euank:
Manually tested that this works as I'd expect.

The two cases this happens in:

1. `local manifest` patches that point to different remotes -- @dm0- pointed out that those can usually just use PR refs instead
1. embargoed bulids -- that could better be fixed by not sharing release build workers between regular and embargoed builds.

I'm not sure whether this is an option we'd actually use or not after thinking about it.

euank:
I think I'm weakly in favor of merging even if we don't immediately use it in jenkins.

PR #705, opened 2017-09-08
kola: add coreos.locksmith.reboot test

dgonyeo:
Adds a test that ensures that locksmith can successfully reboot without
etcd in the default configuration.

Passes with https://github.com/coreos/locksmith/pull/130

### DIFF
--- a/kola/tests/locksmith/locksmith.go
+++ b/kola/tests/locksmith/locksmith.go
@@ -66,7 +66,11 @@ func init() {
       "mode": 420
     }]
   }
-}`),
+}`)})
+	register.Register(&register.Test{
+		Name:        "coreos.locksmith.reboot",
+		Run:         locksmithReboot,
+		ClusterSize: 1,
 	})
 	register.Register(&register.Test{
 		Name:        "coreos.locksmith.tls",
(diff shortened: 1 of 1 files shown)


bgilbert:
Thursday release, so 1535.

kola/tests/locksmith/locksmith.go
```diff
@@ -117,6 +121,27 @@ func init() {
...
+	m := c.Machines()[0]
+
+	output, err := m.SSH("locksmithctl send-need-reboot")
+	if _, ok := err.(*ssh.ExitMissingError); ok {
+		err = nil // A terminated session is perfectly normal during reboot.
+	} else if err == io.EOF {
+		err = nil // Sometimes copying command output returns EOF here.
+	}
+	if err != nil {
+		c.Fatalf("failed to run \"locksmothctl send-need-reboot\": output: %q status: %q", output, err)
```
bgilbert: `s/smoth/smith/` (sorry)

kola/tests/locksmith/locksmith.go
```diff
@@ -117,6 +121,27 @@ func init() {
...
+	}
+
+	err = platform.CheckMachine(m)
```
bgilbert: This doesn't confirm that the machine has rebooted yet, or in fact will reboot at all.  `/proc/sys/kernel/random/boot_id` may help here.

dgonyeo:
I just changed this to cat `/proc/sys/kernel/random/boot_id` before and after the reboot and ensure that they're different. How's that look?

kola/tests/locksmith/locksmith.go
```diff
@@ -117,6 +121,41 @@ func init() {
...
+	if _, ok := err.(*ssh.ExitMissingError); ok {
+		err = nil // A terminated session is perfectly normal during reboot.
+	} else if err == io.EOF {
+		err = nil // Sometimes copying command output returns EOF here.
+	}
+	if err != nil {
+		c.Fatalf("failed to run \"locksmithctl send-need-reboot\": output: %q status: %q", output, err)
+	}
+
+	err = platform.CheckMachine(m)
```
bgilbert: This still has a race between `send-need-reboot` and `CheckMachine`.  I checked what `RebootMachine` does and it stops `sshd.socket` first to avoid races; so does `locksmithCluster` in this file.  If you make that change, the `boot_id` checks shouldn't be necessary after all.

dgonyeo: I updated this to change the command to `sudo systemctl stop sshd.socket && locksmithctl send-need-reboot` and removed the boot_id stuff.

bgilbert:
Re `MinVersion`, I think we're okay without, since `best-effort` mode should also succeed without etcd.

dgonyeo:
Good to merge

PR #709, opened 2017-09-12
kola: atomically update _kola_temp platform-latest symlink

bgilbert:
Avoid races when multiple kola runs are started in close succession.

cc @ajeddeloh

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -460,19 +460,21 @@ func SetupOutputDir(outputDir, platform string) (string, error) {
 	}
 
 	if defaulted {
+		tempLinkPath := filepath.Join(outputDir, "latest")
 		linkPath := filepath.Join(defaultBaseDirName, platform+"-latest")
+		// don't clobber existing files that are not symlinks
 		st, err := os.Lstat(linkPath)
-		if err == nil {
-			if (st.Mode() & os.ModeType) != os.ModeSymlink {
-				return "", fmt.Errorf("%v exists and is not a symlink", linkPath)
-			}
-			if err := os.Remove(linkPath); err != nil {
-				return "", err
-			}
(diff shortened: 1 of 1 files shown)


kola/harness.go
```diff
@@ -460,19 +460,21 @@ func SetupOutputDir(outputDir, platform string) (string, error) {
...
-			if err := os.Remove(linkPath); err != nil {
-				return "", err
-			}
-		} else if !os.IsNotExist(err) {
+		if err == nil && (st.Mode()&os.ModeType) != os.ModeSymlink {
+			return "", fmt.Errorf("%v exists and is not a symlink", linkPath)
+		} else if err != nil && !os.IsNotExist(err) {
+			return "", err
+		}
+		if err := os.Symlink(defaultDirName, tempLinkPath); err != nil {
```
arithx: Currently `tempLinkPath` is only removed if the Rename succeeds, is this intended or should we move it to a defer'd cleanup?

bgilbert: It's being removed only if the rename fails.  If it succeeds, the link is now at its final location so we don't need to clean it up.

arithx: Whoops, brain not working this morning need more caffeine

kola/harness.go
```diff
@@ -460,19 +460,21 @@ func SetupOutputDir(outputDir, platform string) (string, error) {
 	}
 
 	if defaulted {
+		tempLinkPath := filepath.Join(outputDir, "latest")
 		linkPath := filepath.Join(defaultBaseDirName, platform+"-latest")
+		// safety check
```
ajeddeloh: Maybe something more descriptive like `do not clobber existing files that are not symlinks` ?

PR #712, opened 2017-09-13
kola/tests: fix race with raid test

ajeddeloh:
The root device is not always vda. Use names that are stable instead.

Unfortunately this trades one race for another until https://github.com/coreos/ignition/pull/446 makes it into a release, since it uses the `/dev/disk/by-path` symlinks which has a race with udev. It might be better to wait until that is merged and Ignition is updated in the overlay before merging this. In my experience testing it, without the Ignition patch it fails ~10% of the time.

Maybe we should disable this test until then?

### DIFF
--- a/kola/tests/misc/raid.go
+++ b/kola/tests/misc/raid.go
@@ -29,7 +29,7 @@ import (
 var (
 	raidRootUserData = conf.ContainerLinuxConfig(`storage:
   disks:
-    - device: /dev/vdb
+    - device: "/dev/disk/by-id/virtio-secondary"
       wipe_table: true
       partitions:
        - label: root1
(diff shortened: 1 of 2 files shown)


bgilbert:
Do those device paths also work on ARM64?

If the failure rate is only 10%, I'd be inclined to just live with it until the next alpha.

ajeddeloh:
Good catch - they do not.

arithx:
@ajeddeloh https://github.com/coreos/mantle/blob/master/platform/machine/qemu/cluster.go#L266

ajeddeloh:
Updated to expand AdditionalOptions.Disks to allow specifying a serial number, then used that to switch to `/dev/disk/by-id/` symlinks instead. Tested on arm64 and arm64.

PR #721, opened 2017-09-22
kola/tests/docker: add test for /etc/coreos/docker-1.12

bgilbert:


kola/tests/docker/torcx_docker_flag.go
```diff
@@ -0,0 +1,72 @@
...
+	checkVersions(`^1\.12$`, `^1\.12\.`)
+
+	// flag=no
+	if _, err := m.SSH("echo no | sudo tee /etc/coreos/docker-1.12"); err != nil {
+		c.Fatalf("couldn't set docker flag to no: %s", err)
+	}
+	if err := m.Reboot(); err != nil {
+		c.Fatalf("could not reboot: %v", err)
+	}
+	if _, err := m.SSH(`sudo rm -rf /var/lib/docker`); err != nil {
```
euank: since we're only upgrading, this shouldn't be necessary.

dm0-: It shouldn't hurt, though.  Otherwise it would still detect the `overlay` driver from starting at 1.12, which would fail the driver test (if it ran).

euank: Yeah, no preference. I guess since we didn't run a container it won't even take seconds to delete.

This LGTM as well.

bgilbert:
Even if we're still running 1.12 in two years, it won't be bundled in the image, so this test won't be relevant anymore.

PR #723, opened 2017-09-25
kola/spawn: do not leak dnsmasq processes

ajeddeloh:
Remember to call cluster.Destroy. Running locally this ensures dnsmasq gets killed and doesn't leak processes.

### DIFF
--- a/cmd/kola/spawn.go
+++ b/cmd/kola/spawn.go
@@ -89,6 +89,10 @@ func doSpawn(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("Cluster failed: %v", err)
 	}
 
+	if spawnRemove {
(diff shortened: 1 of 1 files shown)


cmd/kola/spawn.go
```diff
@@ -88,6 +88,7 @@ func doSpawn(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("Cluster failed: %v", err)
 	}
+	defer cluster.Destroy()
```
bgilbert: This will destroy machines even if `spawnRemove` is false.

ajeddeloh:
Switched to destroying the cluster (which should tear down the machines as well) instead of tearing down each machine individually.

PR #724, opened 2017-09-26
platform/api/oci/types: add sdk types

arithx:
Adds Types used by the Oracle OCI SDK and their related unit tests.

cc @bgilbert


platform/api/oci/types/image.go
```diff
@@ -0,0 +1,64 @@
...
+package types
+
+type CreateImageInput struct {
+	CompartmentID string             `json:"compartmentId"`
+	DisplayName   *string            `json:"displayName,omitempty"`
+	ImageSource   ImageSourceDetails `json:"imageSourceDetails,omitempty"`
+	InstanceID    *string            `json:"instanceId,omitempty"`
+}
+
+type ImageSourceDetails interface{}
```
euank: Why is this an `interface{}`?

platform/api/oci/types/image.go
```diff
@@ -0,0 +1,64 @@
...
+type ImageSourceViaObjectStorageURI struct {
+	SourceURI  string `json:"sourceUri"`
+	SourceType string `json:"sourceType"`
+
+	ImageSourceDetails `json:"-"`
+}
+
+type ExportImageInput struct {
+	DestinationType string  `json:"destinationType"`
+	BucketName      *string `json:"bucketName,omitempty"`
```
euank: should each of these be `omitempty`, or does `null` actually have a special meaning?

platform/api/oci/types/volume_test.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			marshal, err := json.Marshal(test.in.input)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if string(marshal) != test.out.jsonStr {
```
euank: If we're using testify, this should be `asesrt.Equal`

platform/api/oci/types/instance_test.go
```diff
@@ -0,0 +1,399 @@
...
+				DisplayName:      StrToPtr("example-instance"),
+				HostnameLabel:    StrToPtr("bminstance-1"),
+				ImageID:          "ocid1.image...",
+				IPXEScript:       StrToPtr("example-ipxe-script"),
+				Metadata:         &map[string]string{"foo": "bar"},
+				Shape:            "VM.Standard1.1",
+				SubnetID:         StrToPtr("example-subnet-id"),
+				ExtendedMetadata: RawMessageToPtr(json.RawMessage(`{"extended":{"metadata":"example"}}`)),
+			}},
+			out:  out{jsonStr: `{"availabilityDomain":"Uocm:PHX-AD-1","compartmentId":"ocid1.compartment.oc1..aaaaaaaayzfqeibduyox6iib3olcmdar3ugly4fmameq4h7lcdlihrvur7xq","createVnicDetails":{"assignPublicIp":false,"displayName":"example-vnic-details","hostnameLabel":"bminstance-1","privateIp":"10.0.3.3","skipSourceDestCheck":true,"subnetId":"example-subnet-id"},"displayName":"example-instance","hostnameLabel":"bminstance-1","imageId":"ocid1.image...","ipxeScript":"example-ipxe-script","metadata":{"foo":"bar"},"shape":"VM.Standard1.1","subnetId":"example-subnet-id","extendedMetadata":{"extended":{"metadata":"example"}}}`},
```
euank: This style of testing the marshalled string doesn't seem great.

Notably, things like non-deterministic ordering of maps could easily bite these tests.

platform/api/oci/types/instance.go
```diff
@@ -0,0 +1,94 @@
...
+
+type Instance struct {
+	AvailabilityDomain string             `json:"availabilityDomain"`
+	CompartmentID      string             `json:"compartmentId"`
+	DisplayName        *string            `json:"displayName"`
+	ID                 string             `json:"id"`
+	ImageID            *string            `json:"imageId"`
+	IPXEScript         *string            `json:"ipxeScript"`
+	LifecycleState     string             `json:"lifecycleState"`
+	Metadata           *map[string]string `json:"metadata"`
```
euank: a `map` is already a pointer.

platform/api/oci/types/instance.go
```diff
@@ -0,0 +1,94 @@
...
+	Region             string             `json:"region"`
+	Shape              string             `json:"shape"`
+	TimeCreated        string             `json:"timeCreated"`
+
+	// The ExtendedMetadata property allows nested JSON objects.
+	// It is strongly advised that the Metadata attribute is used
+	// instead.
+	ExtendedMetadata *json.RawMessage `json:"extendedMetadata"`
```
euank: `json.RawMessage` is already a pointer. Any reason this isn't `omitempty`?

platform/api/oci/types/image.go
```diff
@@ -0,0 +1,64 @@
...
+	ObjectName      *string `json:"objectName,omitempty"`
+	DestinationURI  *string `json:"destinationUri,omitempty"`
+}
+
+type UpdateImageInput struct {
+	DisplayName *string `json:"displayName,omitempty"`
+}
+
+type Image struct {
+	BaseImageID            *string `json:"baseImageId"`
```
euank: should each of these be `omitempty`, or does `null` actually have a special meaning?

arithx: This is a response model, it is never Marshal'd

platform/api/oci/types/image.go
```diff
@@ -0,0 +1,64 @@
...
+type ImageSourceViaObjectStorageTuple struct {
+	BucketName    string `json:"bucketName"`
+	NamespaceName string `json:"namespaceName"`
+	ObjectName    string `json:"objectName"`
+	SourceType    string `json:"sourceType"`
+
+	ImageSourceDetails `json:"-"`
+}
+
+type ImageSourceViaObjectStorageURI struct {
```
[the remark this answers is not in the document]
arithx: I believe they are equivalent, I'll drop `ImageSourceViaObjectStorageTuple` in the next round.

platform/api/oci/types/image.go
arithx: It's used as a base for `ImageSourceViaObjectStorageTuple` & `ImageSourceViaObjectStorageURI` because `CreateImageInput` can return either. Per other discussions in the thread though I'm going to drop `ImageStorageViaObjectStorageTuple` and will drop the Interface as a result.

platform/api/oci/types/instance_test.go
arithx: Would it be preferred to `Unmarshal` the output back into an object and compare the objects or build regular expression based asserts on the JSON output?

platform/api/oci/types/volume_test.go
arithx: :+1:

platform/api/oci/types/image.go
euank: For that case, IMO, you'd want to use `json.RawMessage` and have the type it's in have a clarifying name (e.g. `XMessageWrapper`).. and then have the actual proper types specified out as non-wrappers.

PR #714, opened 2017-09-15
travis: update go version

dvrkps:

PR #729, opened 2017-10-04
Update to go 1.9

euank:
Replaces #714 

I'd like to use a shiny new go 1.9 feature, this seems like a good first step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ matrix:
 
 go_import_path: github.com/coreos/mantle
 
+install:
+  # './build' uses 'git describe' which needs tags
+  - git fetch --unshallow --tags
+
 script:
   - ./build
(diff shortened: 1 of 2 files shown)


euank:
@dm0- `t.Helper`. I was going to update the ebuild too, and hopefully everyone else will be willing to bump to 1.9

bgilbert:
@euank Go 1.9 isn't in a Fedora stable release yet.

euank:
Seems reasonable to wait for it to be in fedora before I use the feature. I think it's okay to bump it here in the meanwhile though. Presumably it won't be long before it's in fedora.

While I was at it, I tossed on one more travis commit which actually lets it build in my fork. I'm kinda surprised the `./build` step didn't fail on this repo too without that.

PR #730, opened 2017-10-06
*: add oci platform

arithx:
Adds the OCI platform to `kola` & `ore`.

Replaces #719


bgilbert:
Commit messages:
- Reversion messages should always explain why the reversion is occurring.
- `vendor: add oracle/bmcs-go-sdk ` should explain why the glide configs are not being updated.
- `vendor: update go-ini to 1.28.2` sould explain why 1.8 isn't good enough for `auth/oci.go`.  That information could be useful if the update causes a regression elsewhere and someone is trying to figure out how to avoid it while satisfying `oci.go`.
- Commit message bodies shouldn't substantially just repeat the title.  Ideally the body contains more information about the commit that one can't get from reading the diff (and/or a high-level overview if one can't easily get that from reading the diff), but if there's really nothing more to say, it's better to omit the body than to add noise.  If tailor is encouraging the latter, we should change tailor's config.

arithx:
Commit messages are updated

cmd/ore/oci/setup.go
```diff
@@ -0,0 +1,84 @@
...
+
+	secList, err := API.CreateDefaultSecurityList(vcn.ID)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Creating default Security List: %v\n", err)
+		os.Exit(1)
+	}
+
+	igw, err := API.CreateInternetGateway(vcn.ID)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Creating Internet Gateway: %v\n", err)
```
bgilbert: Should this be fatal?

cmd/ore/oci/setup.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	rt, err := API.CreateDefaultRouteTable(vcn.ID, igw.ID)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Creating default Route Table: %v\n", err)
```
bgilbert: Likewise.

cmd/ore/oci/setup.go
```diff
@@ -0,0 +1,84 @@
...
+
+	ads, err := API.ListAvailabilityDomains()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Listing Availability Domains: %v\n", err)
+		os.Exit(1)
+	}
+
+	ad := ads[0]
+
+	_, err = API.CreateSubnet("kolasubnet", ad.Name, "10.0.0.0/16", vcn.ID, secList.ID, rt.ID)
```
bgilbert: Maybe call it `kola-1`, in case we want to support multiple subnets down the road?

platform/api/oci/instance.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+		CreateVnicOptions: &baremetal.CreateVnicOptions{
+			AssignPublicIp: boolToPtr(true),
+			SubnetID:       subnet.ID,
+		},
+	}
+
+	inst, err := a.client.LaunchInstance(subnet.AvailabilityDomain, a.opts.CompartmentID, a.opts.Image, a.opts.Shape, subnet.ID, &opts)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
arithx: The error messages while a bit verbose convey the relevant information very well.

platform/conf/conf.go
```diff
@@ -262,6 +262,110 @@ func (c *Conf) addSystemdUnitCloudConfig(name, contents string, enable bool) {
...
+			{
+				Name:    name,
+				Content: contents,
+			},
+		},
+		Enable: enable,
+	})
+}
+
+func (c *Conf) AddSystemdUnitDropin(service, name, contents string, enable bool) {
```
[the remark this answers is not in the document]
arithx: Yeah, I only added it to have parity with the AddSystemdUnit method.

platform/machine/oci/cluster.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, err
+	}
+
+	// coreos-metadata doesn't produce the private ipv4, adding an override which pulls
+	// it from the vnics portion fo the metadata service
+	conf.AddSystemdUnitDropin("coreos-metadata.service", "oci_override.conf", `[Service]
+Type=oneshot
+Environment=OUTPUT=/run/metadata/coreos
+ExecStart=/usr/bin/mkdir --parent /run/metadata
+ExecStart=/usr/bin/bash -c 'echo "COREOS_OCI_IPV4_PRIVATE_0=$(curl http://169.254.169.254/opc/v1/vnics/ | jq .[].privateIp -r)" > ${OUTPUT}'`, false)
```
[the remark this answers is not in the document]
arithx: I'll add the `ExecStart=` line to clobber it because once the information is available on the `vnics` API we can update `coreos-metadata` and drop this dropin.

cmd/ore/oci/clear.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdClear = &cobra.Command{
+		Use:   "clear-account",
```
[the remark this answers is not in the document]
arithx: Would `clear-networking` or `clear-region-networking` convey the actions more clearly?

cmd/ore/oci/clear.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	rts, err := API.ListRouteTables(vcn.ID)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Getting Route Tables: %v\n", err)
+		os.Exit(1)
+	}
+
+	for _, rt := range rts {
+		if rt.DisplayName != fmt.Sprintf("Default Route Table for %s", vcn.DisplayName) {
```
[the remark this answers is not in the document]
arithx: I could potentially check if the `TimeCreated` field matches the `TimeCreated` of the `VirtualNetwork` but there is no direct identifier (other than the display name) that a unit is the default.

cmd/ore/oci/setup.go
arithx: :+1:

platform/api/oci/instance.go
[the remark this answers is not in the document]
bgilbert: OK.

cmd/ore/oci/clear.go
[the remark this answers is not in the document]
bgilbert: Better, but "clear" implies completely emptying and it's not doing that.  To my mind, the opposite of "set up" is "tear down", but that's slightly awkward.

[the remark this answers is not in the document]
arithx: Per OOB discussions going with `create-kola-vcn` & `delete-kola-vcn`

arithx:
Updates pushed

platform/api/oci/instance.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	content, err := a.client.ShowConsoleHistoryData(consoleHistoryStatus.ID, &baremetal.ConsoleHistoryDataOptions{
+		Length: 2 * 1024 * 1024,
+		Offset: 0,
+	})
+	if err != nil {
+		return "", fmt.Errorf("getting console history data: %v", err)
+	}
+
```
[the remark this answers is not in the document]
arithx: Added a comment to the top of the function.

auth/oci.go
```diff
@@ -0,0 +1,101 @@
...
+	// attempt to parse the mantle config
+	cfg, err = ini.InsensitiveLoad(fmt.Sprintf("%s.mantle", path))
+	if err == nil {
+		for _, section := range cfg.Sections() {
+			p, _ := profiles[strings.ToLower(section.Name())]
+			err = section.MapTo(&p)
+			profiles[strings.ToLower(section.Name())] = p
+		}
+	} else {
+		plog.Warningf("could not find %s", fmt.Sprintf("%s.mantle", path))
```
bgilbert: Should log `err`, since the file might exist but have bad permissions etc.

platform/api/oci/instance.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		return true, nil
+	})
+	if err != nil {
+		a.TerminateInstance(id)
+		return nil, fmt.Errorf("waiting for machine to become active: %v", err)
+	}
+
+	mach := Machine{}
```
bgilbert: Just put `Name` and `ID` in the `Machine{}` initializer.  You could even move this to the end of the function and set all the fields at once.

platform/api/oci/instance.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		return true, nil
+	})
+	if err != nil {
+		return "", fmt.Errorf("waiting for console history data to be ready: %v", err)
+	}
+
+	content, err := a.client.ShowConsoleHistoryData(consoleHistoryStatus.ID, &baremetal.ConsoleHistoryDataOptions{
+		Length: 2 * 1024 * 1024,
```
bgilbert:
```
// OCI limits console history to 1 MB; request 2 to be safe
```

platform/util.go
```diff
@@ -129,3 +130,26 @@ func StartMachine(m Machine, j *Journal, c RuntimeConfig) error {
 	}
 	return nil
 }
+
+func WaitForActive(timeout, delay time.Duration, checkFunction func() (bool, error)) error {
```
bgilbert: Something like `WaitUntilReady` would be a better name, since "active" is particular to one use case.

Also, please put this in the top-level `util` package, maybe in `retry.go`.  It's not specific to `platform`.

cmd/ore/oci/delete.go
```diff
@@ -0,0 +1,114 @@
...
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdDelete = &cobra.Command{
+		Use:   "delete-kola-vcn",
+		Short: "Delete OCI networking",
+		Long:  "Removes kola virtual cloud network from an OCI account",
```
bgilbert: `Remove`

platform/api/aws/ec2.go
```diff
@@ -104,35 +105,28 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
 
 	// loop until all machines are online
 	var insts []*ec2.Instance
-	// 5 minutes is a pretty reasonable timeframe for AWS instances to work.
```
bgilbert: I don't think we want to lose this comment.

platform/api/aws/ec2.go
```diff
@@ -104,35 +105,28 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
...
-	after := time.After(5 * time.Minute)
-	for done := false; !done; {
-		select {
-		case <-after:
-			a.TerminateInstances(ids)
-			return nil, fmt.Errorf("timed out waiting for instances to run")
-		default:
-		}
-
-		// don't make api calls too quickly, or we will hit the rate limit
```
bgilbert: and probably not this one either.

platform/api/aws/ec2.go
```diff
@@ -104,35 +105,28 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
...
-			return nil, err
+			return false, err
 		}
 		insts = desc.Reservations[0].Instances
 
-		done = true
+		done := true
 		for _, i := range insts {
 			if *i.State.Name != ec2.InstanceStateNameRunning || i.PublicIpAddress == nil {
 				done = false
```
bgilbert: Could return `false, nil` immediately.

cmd/kola/options.go
```diff
@@ -93,6 +93,11 @@ func init() {
 	sv(&kola.ESXOptions.Server, "esx-server", "", "ESX server")
 	sv(&kola.ESXOptions.Profile, "esx-profile", "", "ESX profile (default \"default\")")
 	sv(&kola.ESXOptions.BaseVMName, "esx-base-vm", "", "ESX base VM name")
+
+	// oci-specific options
+	sv(&kola.OCIOptions.Region, "oci-region", "", "OCI region")
+	sv(&kola.OCIOptions.Image, "oci-image-id", "", "OCI image id")
```
bgilbert: I meant `OCI image ID` for the help text, not the parameter name.  We don't use an `-id` suffix for any other parameters.

platform/api/oci/instance.go
```diff
@@ -0,0 +1,144 @@
...
+	"encoding/base64"
+	"fmt"
+	"time"
+
+	"github.com/oracle/bmcs-go-sdk"
+
+	"github.com/coreos/mantle/platform"
+)
+
+func (a *API) CreateInstance(name, userdata string) (*Machine, error) {
```
bgilbert: `name` isn't actually used.  It should probably go in both the `hostnameLabel` and the instance display name.

bgilbert:
`coreos.network.initramfs.second-boot` should be blacklisted on OCI because of iSCSI.

platform/machine/oci/cluster.go
```diff
@@ -0,0 +1,117 @@
...
+	return oc, nil
+}
+
+func (oc *cluster) vmname() string {
+	b := make([]byte, 5)
+	rand.Read(b)
+	return fmt.Sprintf("%s-%x", oc.Name(), b)
+}
+
+func (oc *cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error) {
```
bgilbert: To avoid confusion, this should probably fail if `userdata` is not Ignition-compatible.  `userdata.IsIgnition()` is almost the right test, except it returns `false` for CLCs, which is arguably a bug.  (Its only other caller should probably handle CLCs too.)  Maybe fix it and rename to `IsIgnitionCompatible()`?

arithx: The way other platforms seem to be doing ignition detection is to call the `IsIgnition` function on the `conf.Conf` object after the `conf.UserData` has been rendered. Do we want to change the underlying `IsIgnition` function of `conf.UserData` and change the way those platforms are doing their detection or mimic that pattern?

bgilbert: As discussed OOB, we can do the detection either on the `UserData` or the `Conf`, but `UserData.IsIgnition()` should be fixed regardless.

arithx:
updates pushed

arithx:
Good to merge on my end once travis passes

PR #732, opened 2017-10-13
harness: add support for additional reporters

arithx:
Adds a new mechanism into the harness to provide support for
additional reporters. New reporters can be defined inside of
`harness/reporters` & should implement the `Reporter` interface.

The first additional reporter provided is the JSONReporter. `kola`
runs will now produce an additional file inside of the `_kola_tmp`
logs named `report.json`.

### DIFF
--- a/harness/harness.go
+++ b/harness/harness.go
@@ -29,6 +29,9 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/coreos/mantle/harness/reporters"
+	"github.com/coreos/mantle/harness/testresult"
 )
 
 // H is a type passed to Test functions to manage test state and support formatted test logs.
@@ -67,6 +70,8 @@ type H struct {
 	sub      []*H      // Queue of subtests to be run in parallel.
 
 	isParallel bool
+
+	reporters reporters.Reporters
 }
 
 func (c *H) parentContext() context.Context {
@@ -81,6 +86,15 @@ func (h *H) Verbose() bool {
 	return h.suite.opts.Verbose
 }
 
+func (c *H) status() testresult.TestResult {
+	if c.Failed() {
+		return testresult.Fail
+	} else if c.Skipped() {
+		return testresult.Skip
+	}
+	return testresult.Pass
+}
+
 // flushToParent writes c.output to the parent after first writing the header
 // with the given format and arguments.
 func (c *H) flushToParent(format string, args ...interface{}) {
@@ -90,12 +104,14 @@ func (c *H) flushToParent(format string, args ...interface{}) {
 
 	fmt.Fprintf(p.w, format, args...)
 
+	status := c.status()
+
 	// TODO: include test numbers in TAP output.
 	if p.tap != nil {
 		name := strings.Replace(c.name, "#", "", -1)
-		if c.Failed() {
+		if status == testresult.Fail {
 			fmt.Fprintf(p.tap, "not ok - %s\n", name)
-		} else if c.Skipped() {
+		} else if status == testresult.Skip {
 			fmt.Fprintf(p.tap, "ok - %s # SKIP\n", name)
 		} else {
 			fmt.Fprintf(p.tap, "ok - %s\n", name)
@@ -104,7 +120,8 @@ func (c *H) flushToParent(format string, args ...interface{}) {
 
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	io.Copy(p.w, &c.output)
+	outputBufferCopy := c.output
+	outputBufferCopy.WriteTo(p.w)
 }
 
 type indenter struct {
@@ -429,12 +446,13 @@ func (t *H) Run(name string, f func(t *H)) bool {
 		return true
 	}
 	t = &H{
-		barrier: make(chan bool),
-		signal:  make(chan bool),
-		name:    testName,
-		suite:   t.suite,
-		parent:  t,
-		level:   t.level + 1,
+		barrier:   make(chan bool),
+		signal:    make(chan bool),
+		name:      testName,
+		suite:     t.suite,
+		parent:    t,
+		level:     t.level + 1,
+		reporters: t.reporters,
 	}
 	t.w = indenter{t}
 	// Indent logs 8 spaces to distinguish them from sub-test headers.
@@ -464,15 +482,25 @@ func (t *H) report() {
 	}
 	dstr := fmtDuration(t.duration)
 	format := "--- %s: %s (%s)\n"
-	if t.Failed() {
-		t.flushToParent(format, "FAIL", t.name, dstr)
-	} else if t.suite.opts.Verbose {
-		if t.Skipped() {
-			t.flushToParent(format, "SKIP", t.name, dstr)
-		} else {
-			t.flushToParent(format, "PASS", t.name, dstr)
-		}
+
+	status := t.status()
+	if status == testresult.Fail || t.suite.opts.Verbose {
+		t.flushToParent(format, status, t.name, dstr)
 	}
+
+	// TODO: store multiple buffers for subtests without indentation
+	// potentially add a TeeWriter which will output to both buffers
+	//
+	// original comment from PR ---
+	// euank: As a followup, we could potentially use something other
+	// than c.output here.  c.output came from an indenter. We don't
+	// need the indentation for reporters.  One way we could do that is
+	// by replacing H.w with a TeeWriter which tees to an indented sink
+	// and a normal sink, and then use whichever is appropriate.  We
+	// could also write verbosely to the 'reporter sink'.  I'm fine with
+	// this being a TODO if you don't want to tackle it in this initial
+	// PR.
+	t.reporters.ReportTest(t.name, status, t.duration, t.output.Bytes())
 }
 
 // CleanOutputDir creates/empties an output directory and returns the cleaned path.
--- a/harness/reporters/json.go
+++ b/harness/reporters/json.go
@@ -1,0 +1,72 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package reporters
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/coreos/mantle/harness/testresult"
+)
+
+type jsonReporter struct {
+	Tests    []jsonTest            `json:"tests"`
+	Result   testresult.TestResult `json:"result"`
+	filename string
+
+	// Context variables
+	Platform string `json:"platform"`
+	Version  string `json:"version"`
+}
+
+type jsonTest struct {
+	Name     string                `json:"name"`
+	Result   testresult.TestResult `json:"result"`
+	Duration time.Duration         `json:"duration"`
+	Output   string                `json:"output"`
+}
+
+func NewJSONReporter(filename, platform, version string) *jsonReporter {
+	return &jsonReporter{
+		Platform: platform,
+		Version:  version,
+		filename: filename,
+	}
+}
+
+func (r *jsonReporter) ReportTest(name string, result testresult.TestResult, duration time.Duration, b []byte) {
+	r.Tests = append(r.Tests, jsonTest{
+		Name:     name,
+		Result:   result,
+		Duration: duration,
+		Output:   string(b),
+	})
+}
+
+func (r *jsonReporter) Output(path string) error {
+	f, err := os.Create(filepath.Join(path, r.filename))
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	return json.NewEncoder(f).Encode(r)
+}
+
+func (r *jsonReporter) SetResult(result testresult.TestResult) {
+	r.Result = result
+}
--- a/harness/reporters/reporter.go
+++ b/harness/reporters/reporter.go
@@ -1,0 +1,51 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package reporters
+
+import (
+	"time"
+
+	"github.com/coreos/mantle/harness/testresult"
+)
+
+type Reporters []Reporter
+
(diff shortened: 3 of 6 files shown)


arithx:
cc @euank

harness/suite.go
```diff
@@ -203,6 +208,26 @@ func (s *Suite) Run() (err error) {
 		return err
 	}
 
+	for _, reporter := range s.reporters {
```
euank: We're looping through 'reporters' multiple times, and it feels rather fragile.

I think encapsulating a slice of reporters as its own type (`type Reporters []Reporter`) and then defining methods on it (like 'Create` and `WriteOutput` and such). I think that would make all of this code cleaner.

harness/harness.go
```diff
@@ -464,14 +489,24 @@ func (t *H) report() {
...
+		status = "PASS"
+	}
+
+	if t.suite.opts.Verbose {
+		t.flushToParent(format, status, t.name, dstr)
+	} else {
+		t.mu.Lock()
+		defer t.mu.Unlock()
+		b := t.output.Bytes()
+		t.publishToReporters(t.name, status, t.duration, b)
```
euank: `publishToReporters` being called both here and in `flushToParent` seems awkward.

Couldn't we call it only once and have it deal with output handling itself?

As is, having to realize that `flushToParent` is implicitly calling it makes this code harder to follow and seems like a possible footgun.

harness/reporters/json.go
```diff
@@ -0,0 +1,80 @@
...
+
+package reporters
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
+type JSONReporter struct {
```
euank: doesn't need to be exported

harness/reporters/json.go
```diff
@@ -0,0 +1,80 @@
...
+
+func NewJSONReporter(context map[string]string, filename string) *JSONReporter {
+	return &JSONReporter{
+		context:  context,
+		filename: filename,
+	}
+}
+
+func (r *JSONReporter) MarshalJSON() ([]byte, error) {
+	a := map[string]interface{}{}
```
euank: I would rather have a proper struct defined (and possibly exported) for this json structure.

harness/reporters/json.go
```diff
@@ -0,0 +1,80 @@
...
+func (r *JSONReporter) ReportTest(name, result string, duration time.Duration, b []byte) {
+	r.tests = append(r.tests, JSONTest{
+		Name:     name,
+		Result:   result,
+		Duration: duration,
+		Output:   fmt.Sprintf("%s", string(b)),
+	})
+}
+
+func (r *JSONReporter) GetOutput() ([]byte, error) {
```
euank: Seems like the caller should just do this themselves

harness/reporters/json.go
```diff
@@ -0,0 +1,80 @@
...
+	"time"
+)
+
+type JSONReporter struct {
+	tests  []JSONTest
+	result string
+
+	// values in context will be present in the root level of the
+	// resulting json output
+	context map[string]string
```
euank: I'm not a fan of the loosely-typed map thing; I'd rather have a proper struct defining this, otherwise parsing it later is a chore.

harness/reporters/json.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (r *JSONReporter) SetResult(result string) {
+	r.result = result
+}
+
+func (r *JSONReporter) GetFilename() string {
+	return r.filename
+}
+
+type JSONTest struct {
```
euank: doesn't need to be exported

harness/reporters/reporter.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"time"
+)
+
+type Reporter interface {
+	ReportTest(string, string, time.Duration, []byte)
+	GetOutput() ([]byte, error)
+	SetResult(string)
+	GetFilename() string
```
euank: Nit, it's idiomatic to leave off `Get` here.

arithx:
changes pushed

harness/harness.go
```diff
@@ -90,12 +94,21 @@ func (c *H) flushToParent(format string, args ...interface{}) {
 
 	fmt.Fprintf(p.w, format, args...)
 
+	var status string
```
euank: I think this is done often enough to justify a `c.status()` helper

harness/harness.go
```diff
@@ -464,15 +485,22 @@ func (t *H) report() {
 	}
 	dstr := fmtDuration(t.duration)
 	format := "--- %s: %s (%s)\n"
+
+	var status string
 	if t.Failed() {
```
euank: I think this is more readable if we split up the status setting (or make a helper) and the flushing.

Making it `if t.Failed() || t.suite.opts.Verbose { flushToParent }` by itself makes the reason for that much more obvious.

What you have now  is already an improvement over what it was before, but I think it can still be a little better.

harness/reporters/json.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (r *jsonReporter) SetResult(result string) {
+	r.Result = result
+}
+
+func (r *jsonReporter) Filename() string {
+	return r.filename
+}
+
+type jsonTest struct {
```
euank: I'd rather this be near the top of the file, close to the reporter struct it's used in.

harness/reporters/json.go
```diff
@@ -0,0 +1,87 @@
...
+	Name     string        `json:"name"`
+	Result   string        `json:"result"`
```
euank: nit, but I'd probably split this into its own type (e.g.
```go
type testResult string

const (
	TestResultFail testResult = "FAIL"
	TestResultSkip            = "SKIP"
        ....
)
```
)

It is a bit of a matter of taste, up to you.

harness/reporters/reporter.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+func (reps *Reporters) Cleanup() error {
+	var err multierror.Error
+
+	for _, r := range *reps {
+		e := r.Cleanup()
+		if e != nil {
+			fmt.Println(e)
```
euank: debug code?

harness/reporters/reporter.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	return err.AsError()
+}
+
+func (reps *Reporters) Output() error {
+	for _, r := range *reps {
+		err := r.Output()
+		if err != nil {
+			fmt.Println(err)
```
euank: debug code?

harness/harness.go
```diff
@@ -104,7 +117,14 @@ func (c *H) flushToParent(format string, args ...interface{}) {
 
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	io.Copy(p.w, &c.output)
+	b := c.output.Bytes()
+	io.Copy(p.w, bytes.NewBuffer(b))
+}
+
+func (c *H) publishToReporters(name, status string) {
+	if !c.hasSub {
```
euank: Why skip tests-that-have-subtests? That is what this is doing, right?

harness/reporters/reporter.go
```diff
@@ -0,0 +1,80 @@
...
+	return nil
+}
+
+func (reps *Reporters) SetResult(s string) {
+	for _, r := range *reps {
+		r.SetResult(s)
+	}
+}
+
+func (reps *Reporters) OpenFile(pathFunc func(string) string) error {
```
euank: I'd much rather this be passed a `outputRoot` or something and then figure out the right name based on that.

I'm also not a fan of `OpenFile` as the name. Something like `SetupOutputDir` might be better?

I'm also dubious of the value of not just opening the file in the `Output` function. The file handle isn't being used between this "OpenFile" and the eventual write of output.

I'm not sure that the earlier error reporting / failure is worth this additional complexity

harness/harness.go
arithx: All of the tests are at the same level, this makes it so that the test groupings don't get reported as individual tests (i.e. `coreos.flannel.vxlan` & `coreos.flannel.udp` are reported but `coreos.flannel` is not)

harness/reporters/reporter.go
arithx: I don't have a particular strong opinion either way, the reason that I have the `OpenFile` seperated from the `Output` function is to be able to provide similar functionality to what happens with the Profile's. Inside of the `harness/suite` they have their file handlers opened directly inside of the `Run` function while the actual output is defer'd (see https://github.com/coreos/mantle/blob/master/harness/suite.go#L208).

If we are going to change these to open the file handler inside of the `Output` method do we want to move the profiles to also opening their files inside of the defer'd function?

harness/reporters/reporter.go
arithx: Yup, whoops forgot to remove it.

arithx:
updates pushed

harness/reporters/json.go
```diff
@@ -0,0 +1,81 @@
...
+		Output:   fmt.Sprintf("%s", string(b)),
+	})
+}
+
+func (r *jsonReporter) Output(path string) error {
+	f, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+
```
euank: Right now, if `json.Marshal` returns an error the file won't be closed.

Using `defer` to close the file should fix that, and if you want to still bubble up the error we cna use a named return for that.

harness/reporters/json.go
```diff
@@ -0,0 +1,81 @@
...
+	b, err := json.Marshal(r)
```
euank: Since `f` is a writer, this can be written more simply (without `fprintf`) by using `json.NewEncoder(f).Encode(r)`

harness/reporters/reporter.go
```diff
@@ -0,0 +1,58 @@
...
+	TestResultFail TestResult = "FAIL"
+	TestResultSkip TestResult = "SKIP"
+	TestResultPass TestResult = "PASS"
+)
+
+type TestResult string
+
+type Reporters []Reporter
+
+func (reps *Reporters) ReportTest(name string, result TestResult, duration time.Duration, b []byte) {
```
euank: Is there a reason this is a pointer receiver? Since it's interacted with via `append` below, I don't think it needs to be a `*`

harness/reporters/json.go
```diff
@@ -0,0 +1,81 @@
...
+	r.Tests = append(r.Tests, jsonTest{
+		Name:     name,
+		Result:   result,
+		Duration: duration,
+		Output:   fmt.Sprintf("%s", string(b)),
+	})
+}
+
+func (r *jsonReporter) Output(path string) error {
+	f, err := os.Create(path)
```
euank: Should the reporter append its own name to it, e.g. `filepath.Join(path, "report.json")` and treat the path as a directory?

Given this attempts to support multiple possible reporters, it seems like they'd each either need their own directory or their own file within a provided directory.

harness/suite.go
```diff
@@ -203,6 +211,8 @@ func (s *Suite) Run() (err error) {
 		return err
 	}
 
+	defer s.reporters.Output(s.outputPath)
```
euank: I'm still not a fan of passing in the "outputPath" function for the reporter to call vs passing in the `s.opts.OutputDir` directory here and having the reporter call it vs, say, doing 

```
reportDir := s.outputPath("reports")
os.Mkdir(reportDir)
...
reporters.Output(reportDir)
```

and then having each reporter append its own filename to that given directory.

It feels weird to leak a private function into the reporters abstraction when all that function does is a `filepath.Join`.

harness/harness.go
```diff
@@ -104,7 +119,14 @@ func (c *H) flushToParent(format string, args ...interface{}) {
 
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	io.Copy(p.w, &c.output)
+	b := c.output.Bytes()
```
euank: Technically we can save some memory by writing these two lines as:

```
outputBufferCopy := c.output
outputBufferCopy.WriteTo(p.w)
```

I think that works; It's okay if you don't want to change it from what you have.

harness/suite.go
```diff
@@ -203,6 +211,12 @@ func (s *Suite) Run() (err error) {
 		return err
 	}
 
+	reportDir := s.outputPath("reports")
+	if err := os.Mkdir(reportDir, 0777); err != nil {
+		return err
+	}
+	defer s.reporters.Output(reportDir)
```
euank: this is squashing the error. Since we have a named return error for this function already, this could be...

```
defer func() {
    if reportErr := s.reporters.Output(reportDir); reportErr != nil && err != nil {
        err = reportErr
    }
}()

Comment (arithx) [reply harness/suite.go]: Pushed update with this change.

**Comment by arithx** [comment]
updates pushed

File: harness/reporters/json.go [hunk 24]
```diff
@@ -0,0 +1,71 @@
...
+		filename: filename,
+	}
+}
+
+func (r *jsonReporter) ReportTest(name string, result TestResult, duration time.Duration, b []byte) {
+	r.Tests = append(r.Tests, jsonTest{
+		Name:     name,
+		Result:   result,
+		Duration: duration,
+		Output:   fmt.Sprintf("%s", string(b)),
```
euank: redundant `fmt` call.

harness/harness.go
```diff
@@ -104,7 +119,14 @@ func (c *H) flushToParent(format string, args ...interface{}) {
...
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	io.Copy(p.w, &c.output)
+	outputBufferCopy := c.output
+	outputBufferCopy.WriteTo(p.w)
+}
+
+func (c *H) publishToReporters(name string, status reporters.TestResult) {
+	if !c.hasSub {
+		c.reporters.ReportTest(name, status, c.duration, c.output.Bytes())
```
euank: As a followup, we could potentially use something other than `c.output` here.

`c.output` came from an `indenter`. We don't need the indentation for reporters.

One way we could do that is by replacing `H.w` with a `TeeWriter` which tees to an indented sink and a normal sink, and then use whichever is appropriate.

We could also write verbosely to the 'reporter sink'.

I'm fine with this being a TODO if you don't want to tackle it in this initial PR.

harness/harness.go
```diff
@@ -104,7 +119,14 @@ func (c *H) flushToParent(format string, args ...interface{}) {
 
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	io.Copy(p.w, &c.output)
+	outputBufferCopy := c.output
+	outputBufferCopy.WriteTo(p.w)
+}
+
+func (c *H) publishToReporters(name string, status reporters.TestResult) {
+	if !c.hasSub {
```
euank: I don't think we actually want to skip this. It's possible to miss failures and output as-is.

For example, in the `docker.torcx-manifest-pkgs` tests:

https://github.com/coreos/mantle/blob/2a5b8f93cf2b2d38a303f655b7563e0c7f890f77/kola/tests/docker/torcx_manifest_pkgs.go#L142

If that MustSSH Fails, it will hit this condition of having a subtest (the `install-torcx-profile` above), and thus not be reported as failing.

arithx:
Added TODO & removed `!c.hasSub` check.

harness/reporters/reporter.go
```diff
@@ -0,0 +1,57 @@
...
+// limitations under the License.
+
+package reporters
+
+import (
+	"time"
+)
+
+const (
+	TestResultFail TestResult = "FAIL"
```
euank: based on this also being used in the `harness` pkg, it seems plausible this should be in `harness/testresult` and have the consts renamed to `Fail`/`Skip`/`Pass`.

As-is, it's not too bad. I'm okay with it just living here too.

arithx: I don't have a particular preference on this, so pushing up the changes. I initially left this out because it felt a bit wonky having a package that only had a struct definition and three constants.

Good to merge on my end.

euank:
I'll leave this for @bgilbert to merge or do his own review pass.

bgilbert:
LGTM on a cursory skim.

PR #734, opened 2017-10-16
aws: tag instances on creation, add 'createdby' tag

euank:
This will make cleaning up aws instances reliably easier.

See the individual commits. If desired, I can break out the second one to a separate PR since these are only tangentially related.

### DIFF
--- a/platform/api/aws/ec2.go
+++ b/platform/api/aws/ec2.go
@@ -85,6 +85,10 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
 						Key:   aws.String("Name"),
 						Value: aws.String(name),
 					},
+					&ec2.Tag{
+						Key:   aws.String("CreatedBy"),
+						Value: aws.String("mantle"),
+					},
 				},
 			},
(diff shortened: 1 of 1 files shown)


euank:
Updated to just be a hardcoded "kola" tag value.

arithx:
Do we want this still split into two commits one of which adds the `name` and the other which adds the `createdBy`?

platform/api/aws/ec2.go
```diff
@@ -85,6 +85,10 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
 						Key:   aws.String("Name"),
 						Value: aws.String(name),
 					},
+					&ec2.Tag{
+						Key:   aws.String("CreatedBy"),
+						Value: aws.String("kola"),
```
bgilbert: This is platform code, so should be `mantle`.

euank:
Agreed that I'd prefer them as separate commits since they are distinct changes.. and you can always operate on the merge commit regardless of whether I squash them or not.

PR #735, opened 2017-10-17
platform/conf: update userdata isignition to check for clc's

arithx:
Currently the `IsIgnition` function on the `UserData` object only
returns `true` if the `UserData` is already in `Ignition` form.
CLC's are translated to `Ignition` configs when `Render` is called,
so they should also be marked as `Ignition` in this function.

Renames the function from `IsIgnition` to `IsIgnitionCompatible` to
reflect the change.

More discussion [here](https://github.com/coreos/mantle/pull/730#discussion_r144758775)


arithx:
Good to merge on my end

PR #737, opened 2017-10-18
auth/oci: update config location

arithx:
When the command-line tool for OCI switched from `bmcs` to `oci` the
default config location moved from `~/.oraclebmc` to `~/.oci`.


arithx:
cc @bgilbert

arithx:
Good to merge

PR #741, opened 2017-10-20
kola/tests: rewrite linux.nfs.* to use CLC

arithx:
`linux.nfs.v3` & `linux.nfs.v4` use cloudconfig. This ports them
to Container Linux Configs.

### DIFF
--- a/kola/tests/misc/nfs.go
+++ b/kola/tests/misc/nfs.go
@@ -19,8 +19,6 @@ import (
 	"path"
 	"time"
 
-	"github.com/coreos/coreos-cloudinit/config"
-
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
 	"github.com/coreos/mantle/platform/conf"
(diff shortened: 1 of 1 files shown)


ajeddeloh:
Did you figure out why it was flaking on `aws` (not that it wasn't flaky before) or is that still an open question?

arithx:
The issue I was seeing on AWS was `systemd-hostnamed` failing correctly because it didn't have the permissions to read the `/etc/hostname` file (the CLC wasn't specifying the mode so it was set to [mode 0](https://github.com/coreos/bugs/issues/2208))

kola/tests/misc/nfs.go
```diff
@@ -19,54 +19,28 @@ import (
...
+	nfsserverconf = conf.ContainerLinuxConfig(`storage:
+  files:
+    - filesystem: "root"
+      path: "/etc/hostname"
+      contents:
+        inline: "nfs1"
+      mode: 0644
+    - filesystem: "root"
+      path: "/etc/exports"
+      contents:
```
bgilbert: Also mode 0644 for future-proofing.

arithx: Pushed a change to add this. Good to merge on my end.

PR #742, opened 2017-10-20
platform/oci: update tests

arithx:
Blacklists cloudconfig tests, adds an `IsEmpty` function to `platform/conf`, & implements the `authorized_ssh_key` metadata key.

See individual commit messages for context.

### DIFF
--- a/platform/api/oci/instance.go
+++ b/platform/api/oci/instance.go
@@ -35,8 +35,9 @@ func (a *API) CreateInstance(name, userdata, sshKey string) (*Machine, error) {
 		return nil, err
 	}
 
-	metadata := map[string]string{
-		"user_data": base64.StdEncoding.EncodeToString([]byte(userdata)),
+	metadata := map[string]string{}
+	if userdata != "" {
+		metadata["user_data"] = base64.StdEncoding.EncodeToString([]byte(userdata))
 	}
 	if sshKey != "" {
 		metadata["ssh_authorized_keys"] = sshKey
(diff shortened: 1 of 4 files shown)


platform/api/oci/instance.go
```diff
@@ -37,7 +37,8 @@ func (a *API) CreateInstance(name, userdata string) (*Machine, error) {
 
 	opts := baremetal.LaunchInstanceOptions{
 		Metadata: map[string]string{
-			"user_data": base64.StdEncoding.EncodeToString([]byte(userdata)),
+			"user_data":           base64.StdEncoding.EncodeToString([]byte(userdata)),
+			"ssh_authorized_keys": sshKey,
```
bgilbert: Do we need to omit this field if `sshKey` is empty?

arithx: Pushed a change to omit it when empty

platform/api/oci/instance.go
```diff
@@ -35,10 +35,15 @@ func (a *API) CreateInstance(name, userdata string) (*Machine, error) {
 		return nil, err
 	}
 
+	metadata := map[string]string{
+		"user_data": base64.StdEncoding.EncodeToString([]byte(userdata)),
```
bgilbert: Come to think of it, could you add another commit which sets this conditionally as well?  Users will probably not explicitly set a zero-length `user_data` field.

arithx: New commit pushed addressing this.

arithx:
Good to merge

PR #743, opened 2017-10-20
cmd/kola: add option to specify config file for all platforms

arithx:
Currently only GCE & Packet have the ability to specify the config
file location. This adds the ability to set it for AWS, ESX & OCI.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -72,7 +72,7 @@ func init() {
 	if defaultRegion == "" {
 		defaultRegion = "us-west-2"
 	}
-	// Container Linux 1430.0.0 (alpha) on us-west-2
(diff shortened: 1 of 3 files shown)


cmd/kola/options.go
```diff
@@ -72,7 +72,7 @@ func init() {
 	if defaultRegion == "" {
 		defaultRegion = "us-west-2"
 	}
-	// Container Linux 1430.0.0 (alpha) on us-west-2
+	sv(&kola.AWSOptions.CredentialsFile, "aws-config-file", "", "AWS config file (default \"~/.aws/credentials\")")
```
bgilbert: The `aws` tool distinguishes between config and credentials files, and we want the latter.  We should probably use their terminology here.

arithx: Update pushed addressing this.

arithx:
Good to merge

PR #746, opened 2017-10-22
kola: Support DigitalOcean

bgilbert:
Add ore commands to create and delete custom images.  Allow running kola on DigitalOcean-provided images or on custom images.

All kola tests pass.


platform/machine/do/machine.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (pm *machine) PasswordSSHClient(user string, password string) (*ssh.Client, error) {
+	return pm.cluster.PasswordSSHClient(pm.IP(), user, password)
+}
+
+func (pm *machine) SSH(cmd string) ([]byte, []byte, error) {
+	return pm.cluster.SSH(pm, cmd)
+}
+
+func (m *machine) Reboot() error {
```
arithx: I noticed this pattern elsewhere but figured I'd ask here. Is there a reason that the `Reboot` function names the `*machine` variable as `m` rather than the name being used in all of the others (`pm` in this instance)?

platform/machine/do/cluster.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return &cluster{
+		BaseCluster: bc,
+		api:         api,
+		sshKeyID:    keyID,
+	}, nil
+}
+
+func (pc *cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error) {
```
arithx: Would `dc` / `dm` more applicable than `pc` / `pm`?

platform/api/do/api.go
```diff
@@ -199,6 +210,43 @@ func (a *API) DeleteDroplet(ctx context.Context, dropletID int) error {
 	return nil
 }
 
+func (a *API) GetUserImage(ctx context.Context, imageName string) (*godo.Image, error) {
+	var ret *godo.Image
+	page := godo.ListOptions{
+		PerPage: 200,
+	}
+	for {
+		images, _, err := a.c.Images.ListUser(ctx, &page)
```
arithx: Does this call return an empty list with no error in the case that the page you're requesting doesn't exist? Otherwise is it possible to fall into the error below if the user has exactly 200 images?

bgilbert: Good question; will check.

platform/machine/do/cluster.go
bgilbert: Good catch.  I forgot to update those.

platform/machine/do/machine.go
bgilbert: Probably it happened once and then was copypasted everywhere.  I'll add a patch to fix up the other instances.

platform/api/do/api.go
bgilbert: Empty list. It also turns out that the pages are 1-indexed and not 0-indexed; fixed.

bgilbert:
Addressed comments and rebased.

PR #749, opened 2017-10-23
cmd/ore: add upload-image oci command

arithx:
Adds an `upload-image` command for the OCI platform. This only
uploads the image to objectstorage, it does not import the image.
Currently the `go-bmcs-sdk` does not support importing the image
from anything other than a running instance.

Until the support is added you'll need to manually create the image
via the `oci` command-line tool or the web UX.

### DIFF
--- a/platform/api/oci/image.go
+++ b/platform/api/oci/image.go
@@ -1,0 +1,44 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


arithx:
cc @bgilbert

platform/api/oci/image.go
```diff
@@ -0,0 +1,44 @@
...
+package oci
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/oracle/bmcs-go-sdk"
+)
+
+func (a *API) UploadImage(name string, bucketName, filePath string) (*baremetal.Object, error) {
```
bgilbert: nit: `bucketName` before `name`.

arithx:
Nits addressed and tested. Good on my end

PR #750, opened 2017-10-24
kola/tests/ignition: test all documented resource URL schemes

bgilbert:


cmd/kola/options.go
```diff
@@ -63,6 +63,7 @@ func init() {
 	sv(&kola.AWSOptions.AMI, "aws-ami", "alpha", `AWS AMI ID, or (alpha|beta|stable) to use the latest image`)
 	sv(&kola.AWSOptions.InstanceType, "aws-type", "t2.small", "AWS instance type")
 	sv(&kola.AWSOptions.SecurityGroup, "aws-sg", "kola", "AWS security group name")
+	sv(&kola.AWSOptions.IAMInstanceProfile, "aws-iam-profile", "kola", "AWS IAM instance profile name")
```
ajeddeloh: Should mention you added this flag in the commit message.

kola/tests/ignition/resource.go
```diff
@@ -0,0 +1,141 @@
...
+			  }
+		      ]
+		  }
+	      }`),
+	})
+	register.Register(&register.Test{
+		Name:             "coreos.ignition.v2_1.resource.remote",
+		Run:              resourceRemote,
+		ClusterSize:      1,
+		ExcludePlatforms: []string{"qemu"},
```
ajeddeloh: Why exclude `qemu`?

kola/tests/ignition/resource.go
```diff
@@ -0,0 +1,141 @@
...
+		  "ignition": {
+		      "version": "2.1.0"
+		  },
+		  "storage": {
+		      "files": [
+			  {
+			      "filesystem": "root",
+			      "path": "/resource/s3-auth",
+			      "contents": {
+				  "source": "s3://kola-fixtures/resources/authenticated"
```
ajeddeloh: How are we authenticated to this? Is it by virtue of the `kola` IAM profile? We should document it.

kola/tests/ignition/resource.go
bgilbert: Jenkins runs it without Internet access.

kola/tests/ignition/resource.go
bgilbert: The fixture allows read access from any authenticated user, which we get from the `kola` role.  I'll add a comment.

cmd/kola/options.go
bgilbert: For what purpose?

kola/tests/ignition/resource.go
bgilbert: Actually, the test should also confirm that this is true.

cmd/kola/options.go
ajeddeloh: Makes it easy to tell when the flag was added. Also, the current commit message implies it is hardcoded and not configurable. Adding a comment that there is a flag to change it clarifies the nature of the change.

bgilbert: > Makes it easy to tell when the flag was added.

That's what `git blame` is for.

> Also, the current commit message implies it is hardcoded and not configurable. Adding a comment that there is a flag to change it clarifies the nature of the change.

The flag is pro forma, same as `--aws-sg`.  In practice no one should ever need to use it.

ajeddeloh: > That's what git blame is for

Sure, but if you don't know where these options are (think about new hires, drive by contributers) that doesn't work. 

I'm in general in favor of the commit message including any interface changes, regardless of if we think anyone will use it.

bgilbert:
Updated for comments and rebased.

PR #752, opened 2017-10-31
ore: implement 'gc' subcommand for aws

euank:
The intent is that running this in a distributed-cron-like-thing should be enough to generally catch leaks.

### DIFF
--- a/cmd/ore/aws/gc.go
+++ b/cmd/ore/aws/gc.go
@@ -1,0 +1,48 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


cmd/ore/aws/gc.go
```diff
@@ -0,0 +1,48 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdGC = &cobra.Command{
+		Use:   "gc",
+		Short: "GC resources in AWS",
+		Long:  `Delete GC instances created over the given duration ago`,
```
bgilbert: `s/GC //`

cmd/ore/aws/gc.go
```diff
@@ -0,0 +1,48 @@
...
+		RunE:  runGC,
+	}
+
+	gcDuration time.Duration
+)
+
+func init() {
+	AWS.AddCommand(cmdGC)
+	cmdGC.Flags().DurationVar(&gcDuration, "duration", 2*time.Hour, "how old resources must be before they're considered garbage")
```
bgilbert: Defaulting to two hours seems a bit aggressive.  6?  24?

euank: I split the difference and picked 5

euank:
@arithx They won't for `kola spawn` in the general case, only in the case where you use an account that's cronning `ore gc`... that's more a "this aws account does this" documentation thing I guess?

bgilbert:
Also, that really only applies to folks using the team AWS account.

PR #756, opened 2017-11-02
cmd/ore/gcloud: add delete-images subcommand

bgilbert:
For Jenkins.

### DIFF
--- a/cmd/ore/gcloud/delete-images.go
+++ b/cmd/ore/gcloud/delete-images.go
@@ -1,0 +1,62 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


cmd/ore/gcloud/delete-images.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+)
+
+func init() {
+	GCloud.AddCommand(cmdDeleteImage)
+}
+
+func runDeleteImage(cmd *cobra.Command, args []string) {
+	if len(args) == 0 {
+		fmt.Fprint(os.Stderr, "Specify image name.\n")
```
arithx: Given the command is plural would `Specify image name(s).\n` fit more?

bgilbert: Sure.

PR #759, opened 2017-11-07
cluster: Fix qemu drive warning

glevand:
The serial number is now a parameter of the device.  Fixes run time
warnings like these:

  qemu: -drive 'serial' is deprecated, please use the corresponding option of '-device' instead


glevand:
Rebased to latest.

PR #766, opened 2017-11-21
kola/tests: add test for remote s3 config

arithx:
Adds a new test targeting a remote S3 config which is not viewable
anonymously. It has read access granted to the AuthenticatedUsers
AWS group (similar to the other S3 resource tests).

### DIFF
--- a/kola/tests/ignition/resource.go
+++ b/kola/tests/ignition/resource.go
@@ -113,7 +113,12 @@ func init() {
 		Platforms:   []string{"aws"},
 		UserData: conf.Ignition(`{
 		  "ignition": {
-		      "version": "2.1.0"
+		      "version": "2.1.0",
+		      "config": {
+		          "append": [{
+		              "source": "s3://kola-fixtures/resources/authenticated.ign"
+		          }]
+		      }
 		  },
 		  "storage": {
 		      "files": [
@@ -164,12 +169,17 @@ func resourceS3(c cluster.TestCluster) {
 		// object accessible by any authenticated S3 user, such as
 		// the IAM role associated with the instance
 		"s3-auth": "kola-authenticated",
+		// object created by configuration accessible by any authenticated
+		// S3 user, such as the IAM role associated with the instance
+		"s3-config": "kola-config",
 	})
 
-	// verify that the object is inaccessible anonymously
-	_, _, err := m.SSH("curl -sf https://s3-us-west-2.amazonaws.com/kola-fixtures/resources/authenticated")
-	if err == nil {
-		c.Fatal("anonymously fetching authenticated resource should have failed, but did not")
+	// verify that the objects are inaccessible anonymously
+	for _, objectName := range []string{"authenticated", "authenticated.ign"} {
+		_, _, err := m.SSH("curl -sf https://s3-us-west-2.amazonaws.com/kola-fixtures/resources/" + objectName)
+		if err == nil {
+			c.Fatal("anonymously fetching authenticated resource should have failed, but did not")
+		}
 	}
 
 	// ...but that the anonymous object is accessible


kola/tests/ignition/resource.go
```diff
@@ -128,6 +128,22 @@ func init() {
...
+		Name:        "coreos.ignition.v2_1.resource.s3_config",
+		Run:         remoteS3Config,
+		ClusterSize: 1,
+		Platforms:   []string{"aws"},
+		UserData: conf.Ignition(`{
+			"ignition": {
+				"version": "2.1.0",
+				"config": {
+					"append": [{
+						"source": "s3://kola-fixtures/resources/config.ign"
```
bgilbert: Let's rename the object to `authenticated.ign` to clarify that the ACL is relevant.

kola/tests/ignition/resource.go
```diff
@@ -128,6 +128,22 @@ func init() {
 		  }
 	      }`),
 	})
+	register.Register(&register.Test{
```
bgilbert: Would it make sense to combine this with `coreos.ignition.v2_1.resource.s3`?  It doesn't seem as though a separate test is necessary.

arithx: SGTM

arithx:
Update pushed

kola/tests/ignition/resource.go
```diff
@@ -112,21 +112,15 @@ func init() {
...
-		  "ignition": {
-		      "version": "2.1.0"
-		  },
-		  "storage": {
-		      "files": [
-			  {
-			      "filesystem": "root",
-			      "path": "/resource/s3-auth",
-			      "contents": {
-				  "source": "s3://kola-fixtures/resources/authenticated"
```
bgilbert: As discussed OOB, please leave this part of the test in place.

Also, the commit message needs updated.

arithx: Re-added this portion of the test (along with the validation) and updated the commit message. It looks like my reformatting of the text spacing made github think it was entirely new.

kola/tests/ignition/resource.go
```diff
@@ -112,21 +112,26 @@ func init() {
...
-			      }
-			  }
-		      ]
-		  }
-	      }`),
+			"ignition": {
```
bgilbert: Ideally this wouldn't be reindented.

kola/tests/ignition/resource.go
```diff
@@ -161,16 +166,21 @@ func resourceS3(c cluster.TestCluster) {
 	m := c.Machines()[0]
 
 	checkResources(c, m, map[string]string{
-		// object accessible by any authenticated S3 user, such as
-		// the IAM role associated with the instance
-		"s3-auth": "kola-authenticated",
+		// object created by configuration accessible by any authenticated
```
bgilbert: This comment is not correct for `s3-auth`.

kola/tests/ignition/resource.go
```diff
@@ -161,16 +166,21 @@ func resourceS3(c cluster.TestCluster) {
...
+		"s3-config": "kola-config",
 	})
 
-	// verify that the object is inaccessible anonymously
+	// verify that the objects are inaccessible anonymously
 	_, _, err := m.SSH("curl -sf https://s3-us-west-2.amazonaws.com/kola-fixtures/resources/authenticated")
 	if err == nil {
 		c.Fatal("anonymously fetching authenticated resource should have failed, but did not")
 	}
+	_, _, err = m.SSH("curl -sf https://s3-us-west-2.amazonaws.com/kola-fixtures/resources/authenticated.ign")
```
bgilbert: Would be slightly cleaner looping over a slice of URLs.

arithx:
Reverted to original spacing & updated to loop over slice of object names and construct urls for the curls

arithx:
Good to merge

PR #770, opened 2017-11-28
journal: also log raw journal output

ajeddeloh:
Save the raw `journalctl -o export` output as well as the output already saved. This enables binary journals to be built using `systemd-journal-remote` and then queried with `journalctl`. If we're concerned about the filesize (expect ~1Mb/machine instance), we could gzip it as well pretty trivially, but since disk is cheap and jenkins ends up xz'ing `_kola_temp` in the end, I'm not too worried.

### DIFF
--- a/platform/journal.go
+++ b/platform/journal.go
@@ -15,10 +15,14 @@
 package platform
 
 import (
+	"compress/gzip"
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
+
+	"github.com/coreos/pkg/multierror"
 
 	"github.com/coreos/mantle/network/journal"
 	"github.com/coreos/mantle/util"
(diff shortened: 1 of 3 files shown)


platform/journal.go
```diff
@@ -26,23 +26,31 @@ import (
...
-// inside the given output directory.
+// and "journal_raw.txt" inside the given output directory.
 func NewJournal(dir string) (*Journal, error) {
 	p := filepath.Join(dir, "journal.txt")
 	j, err := os.OpenFile(p, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0666)
 	if err != nil {
 		return nil, err
 	}
 
+	p = filepath.Join(dir, "journal_raw.txt")
```
bgilbert: I prefer bikesheds to be painted hyphen-colored.

ajeddeloh:
Rebased, hypenated, gzip'd, and cleaned up some error handling too. Added gz/cleanup as a seperate commit, since it seemed distinct enough.

platform/journal.go
```diff
@@ -83,16 +92,18 @@ func (j *Journal) Start(ctx context.Context, m Machine) error {
...
+		}
 	}
-	if err2 := j.journal.Close(); err == nil && err2 != nil {
-		err = err2
+	if e := j.journal.Close(); e != nil {
+		err = append(err, e)
 	}
-	if err2 := j.journalRaw.Close(); err == nil && err2 != nil {
-		err = err2
+	if e := j.journalRaw.Close(); e != nil {
```
bgilbert: [Leaks the underlying File](https://golang.org/pkg/compress/gzip/#Writer.Close).

PR #771, opened 2017-11-29
platform/azure: add kola/ore commands for Azure

arithx:
Add support for Azure to kola and add additional ore commands utilizing the ARM API.


arithx:
Marking WIP until I can perform additional testing (have been hitting an error `HTTP_1_1_REQUIRED` when creating subnets the past few days)

arithx:
All tests that aren't excluded are now passing. Unmarking as WIP. Good to review.

arithx:
@dm0- that works on Azure, testing on other platforms currently and if everything is passing I'll push the change

arithx:
Updates are pushed and this is ready for review.

I've tested all of the new ARM based functionality but do not have access to the ASM creds to test those components (the old ore commands & plume).

arithx:
@dm0- in the updated locksmith TLS tests I've changed them to generate the cert into `/etc/ssl/etcd.cnf` and then pass that profile/extfile to the openssl commands.

auth/azure.go
```diff
@@ -16,17 +16,34 @@ package auth
...
+	"unicode/utf16"
 
-	"github.com/coreos/mantle/platform/api/azure"
+	"github.com/coreos/mantle/platform"
+	"github.com/dimchansky/utfbom"
 )
 
 const AzureProfilePath = ".azure/azureProfile.json"
 
+type Options struct {
```
ajeddeloh: Should have a comment saying this is duplicating the one in platform

auth/azure.go
```diff
@@ -148,3 +168,25 @@ func ReadAzureProfile(path string) (*AzureProfile, error) {
 
 	return &ap, nil
 }
+
+func DecodeUtfbomFile(contents []byte) ([]byte, error) {
```
ajeddeloh: Why not use https://godoc.org/golang.org/x/text/encoding/unicode

platform/api/azure/api.go
```diff
@@ -55,5 +90,42 @@ func New(opts *Options) (*API, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve image: %v", err)
+	}
+
 	return api, nil
 }
+
+func randomName(prefix string) string {
+	b := make([]byte, 5)
+	rand.Read(b)
```
ajeddeloh: Should be `ReadFull()` (see https://github.com/satori/go.uuid/issues/73 as an example why)

platform/api/azure/image.go
```diff
@@ -66,3 +70,53 @@ func IsConflictError(err error) bool {
...
+	if a.opts.DiskURI != "" || a.opts.Version != "" {
+		return nil
+	}
+
+	resp, err := http.DefaultClient.Get(fmt.Sprintf("https://%s.release.core-os.net/amd64-usr/current/version.txt", a.opts.Sku))
+	if err != nil {
+		return fmt.Errorf("unable to fetch release bucket %v version: %v", a.opts.Sku, err)
+	}
+
+	scanner := bufio.NewScanner(resp.Body)
```
ajeddeloh: I think I remember seeing parsing this format somewhere else. Should break that out into a util package or something after this PR.

platform/api/azure/instance.go
```diff
@@ -0,0 +1,228 @@
...
+				BootDiagnostics: &compute.BootDiagnostics{
+					Enabled:    util.BoolToPtr(true),
+					StorageURI: &storageAccountURI,
+				},
+			},
+		},
+	}
+}
+
+func (a *API) CreateInstance(name, userdata, sshkey, resourceGroup, storageAccount string) (*Machine, error) {
```
ajeddeloh: Should this do cleanup for partial failures or is that handled by the caller?

platform/api/azure/instance.go
```diff
@@ -0,0 +1,228 @@
...
+	kr, err := a.GetStorageServiceKeysARM(storageAccount, resourceGroup)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving storage service keys: %v", err)
+	}
+
+	if kr.Keys == nil {
+		return nil, fmt.Errorf("no storage service keys found")
+	}
+	var key string
+	for _, k := range *kr.Keys {
```
ajeddeloh: could this just be
```
key := *kr.Keys[0]
```
with a check above that `len(*kr.Keys) != 0`?

platform/api/azure/instance.go
```diff
@@ -0,0 +1,228 @@
...
+	consoleURI := vm.VirtualMachineProperties.InstanceView.BootDiagnostics.SerialConsoleLogBlobURI
+	if consoleURI == nil {
+		return nil, fmt.Errorf("serial console URI is nil")
+	}
+
+	// Only the full URI to the logs are present in the virtual machine
+	// properties. Parse out the container & file name to use the GetBlob
+	// API call directly.
+	uri := []byte(*consoleURI)
+	containerre := regexp.MustCompile(`bootdiagnostics-kola[a-z0-9\-]+`)
```
ajeddeloh: nit: camelCase all these

platform/api/azure/instance.go
```diff
@@ -0,0 +1,228 @@
...
+	var data io.ReadCloser
+	err = util.Retry(6, 10*time.Second, func() error {
+		data, err = a.GetBlob(storageAccount, key, container, blobname)
+		return err
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	buf := new(bytes.Buffer)
```
ajeddeloh: would `ioutil.ReadAll` work?

platform/api/azure/network.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return a.createSubnet(resourceGroup)
+}
+
+func (a *API) createVirtualNetwork(resourceGroup string) error {
+	_, err := a.netClient.CreateOrUpdate(resourceGroup, "kola-vn", network.VirtualNetwork{
+		Location: &a.opts.Location,
+		VirtualNetworkPropertiesFormat: &network.VirtualNetworkPropertiesFormat{
+			AddressSpace: &network.AddressSpace{
+				AddressPrefixes: &[]string{"10.0.0.0/16"},
```
ajeddeloh: move `10.0.0.0/16` to a const

platform/api/azure/network.go
```diff
@@ -0,0 +1,154 @@
...
+		},
+	}, nil)
+
+	return err
+}
+
+func (a *API) createSubnet(resourceGroup string) (network.Subnet, error) {
+	_, err := a.subClient.CreateOrUpdate(resourceGroup, "kola-vn", "kola-subnet", network.Subnet{
+		SubnetPropertiesFormat: &network.SubnetPropertiesFormat{
+			AddressPrefix: util.StrToPtr("10.0.0.0/24"),
```
ajeddeloh: move `10.0.0.0/24` to a const

platform/api/azure/network.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	if ip.PublicIPAddressPropertiesFormat.IPAddress == nil {
+		return "", fmt.Errorf("IP Address is nil")
+	}
+
+	return *ip.PublicIPAddressPropertiesFormat.IPAddress, nil
+}
+
+func (a *API) GetIPAddresses(name, publicIPName, resourceGroup string) (string, string, error) {
```
ajeddeloh: comment would be good so it's apparent which returned string is which

platform/machine/azure/cluster.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if !rconf.NoSSHKeyInMetadata {
+		keys, err := ac.Keys()
+		if err != nil {
+			return nil, err
+		}
+		ac.sshKey = keys[0].String()
```
ajeddeloh: is `keys` always going to have at least 1 element?

platform/machine/azure/cluster.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		ac.sshKey = key
+	}
+
+	return ac, nil
+}
+
+func (ac *cluster) vmname() string {
+	b := make([]byte, 5)
+	rand.Read(b)
```
ajeddeloh: ReadFull

kola/tests/ignition/sethostname.go
```diff
@@ -59,17 +59,23 @@ func init() {
 		              ]
 		          }
 		      }`)
+
+	// These tests are disabled on Azure because the hostname
```
ajeddeloh: Maybe list the disabled tests in the commit message

README.md
```diff
@@ -214,7 +214,40 @@ sudo emerge --ask awscli
 ```
 
 ### azure
-TBD (FIXME)
+`azure` uses `~/.azure/azureProfile.json`. This can be created using the `az` [command](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli):
+```
+$ az login`
+```
+It also requires that the environment variable `AZURE_AUTH_LOCATION` points to a JSON file. The JSON file will require a service provider active directory account to be created.
```
ajeddeloh: Can mantle set this if it's unset so users do not need to?

README.md
```diff
@@ -214,7 +214,40 @@ sudo emerge --ask awscli
...
+
+You can find your subscriptionId & tenantId in the `~/.azure/azureProfile.json` via:
+```
+cat ~/.azure/azureProfile.json | jq '{subscriptionId: .subscriptions[].id, tenantId: .subscriptions[].tenantId}'
+```
+
+The JSON file exported to the variable `AZURE_AUTH_LOCATION` should be generated by hand and have the following contents:
+```
+{
+  "clientId": "<service provider id>",
```
ajeddeloh: should note this is from the `az ad sp create-for-rbac`

platform/api/azure/instance.go
arithx: At the end of every cluster's lifecycle the entire resource group is deleted which will remove any resources used by the test.

platform/machine/azure/cluster.go
arithx: There are other platforms that also utilize the same equivalent function, do we want to update it there too? ([Packet](https://github.com/coreos/mantle/blob/master/platform/machine/packet/cluster.go#L158-L162), [DO](https://github.com/coreos/mantle/blob/master/platform/machine/do/cluster.go#L141-L145), [ESX](https://github.com/coreos/mantle/blob/master/platform/machine/esx/cluster.go#L64-L68))

README.md
ajeddeloh: I'd say just pick one and go with it, needing to set an environment variable to run kola is a pain.

platform/machine/azure/cluster.go
ajeddeloh: Yeah probably.

auth/azure.go
arithx: We already have `dimchansky/utfbom` as a vendored dependency because of Azure, switching to `x/text/encoding/unicode` will also require `x/text/transform` neither of which are currently vendored.

Do you have a preference between the two?

platform/machine/azure/cluster.go
arithx: I believe so, when a new SSH agent (which is what the `Keys` function is listing) is created a key is always generated (see `NewSSHAgent` in `network/ssh.go`). We're also doing this same direct access of `keys[0]` in DigitalOcean.

README.md
arithx: @bgilbert: do you have a preference?

platform/machine/azure/cluster.go
arithx: Looking into this [`rand.Read`](https://golang.org/pkg/crypto/rand/#Read) uses `io.ReadFull` on the reader.

README.md
arithx: One thing I'd add about setting this environment variable is that there isn't going to be a good way to Unset the environment variable. We tend to do platform specific configuration inside of the Cluster or API but those do not provide sufficient points for deferred cleanup of an environment variable.

ajeddeloh: IIRC environment vars should die with the process.

arithx:
Update pushed

cmd/ore/azure/azure.go
```diff
@@ -42,6 +43,7 @@ func init() {
 
 	sv := Azure.PersistentFlags().StringVar
 	sv(&azureProfile, "azure-profile", "", "Azure Profile json file")
+	sv(&azureAuth, "azure-auth", "", "Azure auth location")
```
ajeddeloh: can we default this to something like `.azure/credentials.json`. I don't really care what, but we should have a canonical location so kola runs with azure don't need special args/envs

arithx:
Updates pushed

- Automatically importing the credentials file (defaults to `~/.azure/credentials.json`)
- Fixed a bug where the API would try to resolve the image when running `ore`
- Enabled `AcceleratedNetworking` on the NIC (required bumping the default size to `Standard_D2_v2`).

arithx:
Made a small change to the `create-image-arm` command to output a JSON struct with the ID & Location for future jenkins work.

ajeddeloh:
More usability nits from testing this again:
 - s/img/image in ore commands
 - I'd prefer named arguments instead of positional arguments for `ore azure upload-image-blob`. Additionally we should set defaults for storage account and container
 - default the upload-image-blob to default to the latest built `coreos_production_azure_image.vhd` similar to how the ore aws code works

I'd be fine merging as is without those changes, but they'd be nice to have

README.md
```diff
@@ -214,7 +214,40 @@ sudo emerge --ask awscli
...
 
 ### azure
-TBD (FIXME)
+`azure` uses `~/.azure/azureProfile.json`. This can be created using the `az` [command](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli):
+```
+$ az login`
+```
+It also requires that the environment variable `AZURE_AUTH_LOCATION` points to a JSON file (this can also be set via the `--azure-auth` parameter). The JSON file will require a service provider active directory account to be created.
+
+Service provider accounts can be created via the `az` command (the output will contain an ID which is used as the `clientId` variable in the `AZURE_AUTH_LOCATION` JSON):
```
csssuf: Would be good to note that you want the `appId` field of the output here, since there are a few other ID-looking fields.

arithx:
Pushed addressing nits

cmd/ore/azure/upload-blob-arm.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/Microsoft/azure-vhd-utils-for-go/vhdcore/validator"
+	"github.com/spf13/cobra"
+
+	"github.com/coreos/mantle/sdk"
+)
+
+var (
+	cmdUploadBlobARM = &cobra.Command{
+		Use:   "upload-blob-arm storage-account container blob-name file",
```
ajeddeloh: Needs to be updated

arithx: Done.

arithx:
rebased & updated

platform/machine/azure/flight.go
```diff
@@ -0,0 +1,119 @@
...
+		flight:      af,
+	}
+
+	if !rconf.NoSSHKeyInMetadata {
+		ac.sshKey = af.SSHKey
+	} else {
+		ac.sshKey = af.FakeSSHKey
+	}
+
+	ac.ResourceGroup, err = af.api.CreateResourceGroup("kola-cluster")
```
arithx: I'm going to open up an issue once this PR merges for moving the creation of the Resource Group & Storage Account into the Flight. It would have required a non-significant amount of refactoring to include in this change set, as such to reduce review complexity (@ajeddeloh has reviewed this code previously) I've left it as is.

auth/azure.go
```diff
@@ -18,14 +18,39 @@ import (
...
+)
+
+const (
+	AzureAuthPath    = ".azure/credentials.json"
+	AzureProfilePath = ".azure/azureProfile.json"
 )
 
-const AzureProfilePath = ".azure/azureProfile.json"
+// A version of the Options struct from platform/api/azure that only
+// contains the ASM values. Otherwise there's a cyclical depdendency
```
jlebon: dependence

cmd/ore/azure/create-image-arm.go
```diff
@@ -0,0 +1,74 @@
...
+	sv := cmdCreateImageARM.Flags().StringVar
+
+	sv(&imageName, "image-name", "", "image name")
+	sv(&blobUrl, "image-blob", "", "source blob url")
+	sv(&resourceGroup, "resource-group", "kola", "resource group name")
+
+	Azure.AddCommand(cmdCreateImageARM)
+}
+
+func runCreateImageARM(cmd *cobra.Command, args []string) error {
```
jlebon: I was really confused for a sec and thought Azure supported ARM.

arithx: Yeah ARM (Azure Resource Manager) can be a confusing name when looking from our usual scope.

imcleod:
I tested the ARM functionality here end to end with @arithx earlier today.  It LGTM and, in fact, is absolutely essential for us to push images to Azure using the type of credentials available to us these days.  Put more bluntly, without this, the existing azure implementation is not useful for COSA/RHCOS/FCOS.

Any effort to get this merged quickly would be greatly appreciated.

(cc: @yuqi-zhang )

platform/machine/azure/flight.go
```diff
@@ -0,0 +1,119 @@
+// Copyright 2016 CoreOS, Inc.
```
ajeddeloh: Should probably only have RH

platform/machine/azure/flight.go
```diff
@@ -0,0 +1,119 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package azure
+
+import (
+	"fmt"
```
ajeddeloh: \n between go standard library and other imports

arithx:
Comments addressed

PR #772, opened 2017-11-30
platform/qemu: log when qemu process exits

ajeddeloh:
Currently the qemu process is spawned and not touched until we kill it.
Add a goroutine to wait for it to exit and then log an error if it was
not killed.

This should at least serve as a starting point for hunting down the reboot test flake.

### DIFF
--- a/platform/machine/qemu/cluster.go
+++ b/platform/machine/qemu/cluster.go
@@ -18,10 +18,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	osexec "os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
+	"syscall"
 
 	"github.com/coreos/pkg/capnslog"
 	"github.com/satori/go.uuid"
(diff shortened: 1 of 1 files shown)


platform/machine/qemu/cluster.go
```diff
@@ -256,6 +258,22 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options Machin
...
+	// wait in the background for qemu to die so we can tell if it was killed (as it should be) or ran in to
+	// some other problem and died.
+	go func() {
+		err := qm.qemu.Wait()
+		if exitErr, ok := err.(*osexec.ExitError); ok && err != nil {
+			plog.Debugf("qemu exited: %v", exitErr)
+			// safe type assertion - we only support Linux
+			if waitStatus := exitErr.ProcessState.Sys().(syscall.WaitStatus); !waitStatus.Signaled() {
+				// Under normal operation, we kill qemu. It should not exit; if it does, log it.
+				plog.Errorf("qemu exited unexpectantly: %v", exitErr)
```
bgilbert: `unexpectedly`

platform/machine/qemu/cluster.go
```diff
@@ -256,6 +258,22 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options Machin
 		return nil, err
 	}
 
+	// wait in the background for qemu to die so we can tell if it was killed (as it should be) or ran in to
+	// some other problem and died.
+	go func() {
+		err := qm.qemu.Wait()
+		if exitErr, ok := err.(*osexec.ExitError); ok && err != nil {
+			plog.Debugf("qemu exited: %v", exitErr)
```
bgilbert: Maybe should be in an `else` clause below.

platform/machine/qemu/cluster.go
```diff
@@ -256,6 +258,22 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options Machin
 		return nil, err
 	}
 
+	// wait in the background for qemu to die so we can tell if it was killed (as it should be) or ran in to
+	// some other problem and died.
+	go func() {
```
bgilbert: It seems like this should be synchronous in `qm.Destroy()`.  It'll probably still need to use `plog`, since no one checks the return value from `Destroy`, but at least it avoids the goroutine.

ajeddeloh:
Huh, looks like this should already be logged. Turns out our exec.Cmd wrapper does a Wait() in its Kill() method after killing; see: https://github.com/coreos/mantle/blob/master/platform/machine/qemu/machine.go#L65. Closing this in favor of that.

PR #774, opened 2017-11-30
cmd/cork: add oci to download-image

arithx:
Adds the oci platform to the download-image command of cork.


arithx:
Good to merge

Issue #768, opened 2017-11-21
Kola tests flake because reboots fail randomly


ajeddeloh:
kola tests will frequently cause the nightly to fail because the machine goes down for a reboot and doesn't come back up for unknown reasons. `console.txt` shows the machine going down for a reboot then ends.

Example `$ tail console.txt` from a failed `coreos.locksmith.reboot`:

```
[   14.296903] EXT4-fs (vda9): re-mounted. Opts: data=ordered
[   14.297540] systemd-shutdown[1]: Remounting '/usr' read-only with options 'seclabel,block_validity,delalloc,barrier,user_xattr,acl'.
[   14.298362] EXT4-fs (dm-0): re-mounted. Opts: block_validity,delalloc,barrier,user_xattr,acl
[   14.299057] systemd-shutdown[1]: Unmounting /usr.
[   14.299377] systemd-shutdown[1]: Could not unmount /usr: Device or resource busy
[   14.299857] systemd-shutdown[1]: Remounting '/' read-only with options 'seclabel,data=ordered'.
[   14.300472] EXT4-fs (vda9): re-mounted. Opts: data=ordered
[   14.306517] Unregister pv shared memory for cpu 0
[   14.307076] reboot: Restarting system
[   14.307332] reboot: machine restart
```

I've only observed this with the nightly builds. I'm currently trying to reproduce locally and will update this bug once I ensure I can reproduce locally.

My guess is that qemu itself is dying for some reason.

bgilbert:
Is this a `qemu`-specific problem?

ajeddeloh:
It looks like it - I can't find any failures of this kind on other platforms. `qemu_uefi` is hard to tell since those time out half the time (woo, another bug).

I'm fairly sure this is the reason we get the a lot of weird test flakes, especially the `coreos.verity.*` ones.

ajeddeloh:
Looking at console output from failed nightly builds now that https://github.com/coreos/mantle/pull/775 has merged, it appears qemu is either exiting with status 0, is being killed in error, or is getting OOM-killed. Logs from the jenkins worker do not show it being OOM-killed.

ajeddeloh:
Confirmed kola does log stderr from qemu to stderr as well.

PR #775, opened 2017-12-01
platform/qemu: clean up error handling

ajeddeloh:
Clean up the error handling and actually log the errors when they occur.

I chose to log the error inside `Destroy()` instead of at the caller because `Destroy()` is frequently called in as `defer m.Destroy()` or in cleanup when creating a cluster fails. Furthermore there's usually not any action we'd take when `Destroy()` fails other than logging it.

I'll see what feedback there is for this, then port it to the other platforms once this is hammered out.

This will hopefully provide at least a start to figuring out what's going on with https://github.com/coreos/mantle/issues/768

### DIFF
--- a/platform/machine/qemu/machine.go
+++ b/platform/machine/qemu/machine.go
@@ -17,6 +17,7 @@ package qemu
 import (
 	"io/ioutil"
 
+	"github.com/coreos/pkg/multierror"
 	"golang.org/x/crypto/ssh"
 
 	"github.com/coreos/mantle/platform"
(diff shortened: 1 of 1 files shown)


platform/machine/qemu/machine.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/platform/local"
 	"github.com/coreos/mantle/system/exec"
+
+	"github.com/coreos/pkg/multierror"
```
arithx: Instead of making another block let's move this up with the import of `golang.org/x/crypto/ssh` similar to https://github.com/coreos/mantle/blob/6b13ddf1ce09e21f6dd0708e571a3bd02418d369/platform/base.go#L25

ajeddeloh:
fixed

platform/machine/qemu/machine.go
```diff
@@ -63,20 +64,28 @@ func (m *machine) Reboot() error {
...
-		err = err2
+	} else {
+		ret = append(ret, err)
 	}
 
 	m.qc.DelMach(m)
 
+	err := ret.AsError()
+	if err != nil {
+		plog.Errorf("Error destoying instance: %v", err)
```
bgilbert: `destroying`

Should probably interpolate `m.ID()` also.

PR #776, opened 2017-12-01
WIP: check for files not present/changed after wiping the disk

ajeddeloh:
Do not merge. Posting this so we have a tool to see what needs to be created by tmpfiles or stripped from the image we ship.

### DIFF
--- a/kola/tests/misc/files.go
+++ b/kola/tests/misc/files.go
@@ -16,10 +16,13 @@ package misc
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform"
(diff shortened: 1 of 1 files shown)


kola/tests/misc/files.go
```diff
@@ -208,3 +212,148 @@ func Blacklist(c cluster.TestCluster) {
 		c.Fatalf("Blacklisted files or directories found:\n%s", output)
 	}
 }
+
+func getAllFiles(c cluster.TestCluster, m platform.Machine) map[string]string {
+	out := string(c.MustSSH(m, "sudo find / -ignore_readdir_race -mount -path /var -prune -o -type f -print0 | xargs -0 sudo md5sum"))
+	ret := map[string]string{}
```
glevand: I think this should also get the permissions and extended attributes for comparison, as those will effect system behavior.  In the very least compare `ls -lZ`.

PR #778, opened 2017-12-02
Log machine destruction and make Destroy() void

ajeddeloh:
Use multierror to keep trying to clean things up when destroying machines, even if part of it errors out. We should clean up as much as possible. Also log the errors in `Destroy()` itself, since the callers rarely look at the return value.

### DIFF
--- a/platform/machine/aws/machine.go
+++ b/platform/machine/aws/machine.go
@@ -60,25 +60,21 @@ func (am *machine) Reboot() error {
 	return platform.RebootMachine(am, am.journal, am.cluster.RuntimeConf())
 }
 
-func (am *machine) Destroy() error {
+func (am *machine) Destroy() {
 	if err := am.cluster.api.TerminateInstances([]string{am.ID()}); err != nil {
-		return err
+		plog.Errorf("Error terminating instance %v: %v", am.ID(), err)
 	}
 
 	if am.journal != nil {
-		if err := am.journal.Destroy(); err != nil {
-			return err
-		}
+		am.journal.Destroy()
 	}
 
 	// faster when run after termination
 	if err := am.saveConsole(); err != nil {
-		return err
+		plog.Errorf("Error saving console for instance %v: %v", am.ID(), err)
 	}
 
 	am.cluster.DelMach(am)
-
-	return nil
 }
 
 func (am *machine) ConsoleOutput() string {
--- a/platform/machine/esx/machine.go
+++ b/platform/machine/esx/machine.go
@@ -61,28 +61,24 @@ func (em *machine) Reboot() error {
 	return platform.RebootMachine(em, em.journal, em.cluster.RuntimeConf())
 }
 
-func (em *machine) Destroy() error {
+func (em *machine) Destroy() {
 	if err := em.cluster.api.TerminateDevice(em.ID()); err != nil {
-		return err
+		plog.Errorf("Error terminating device %v: %v", em.ID(), err)
 	}
 
 	if em.journal != nil {
-		if err := em.journal.Destroy(); err != nil {
-			return err
-		}
+		em.journal.Destroy()
 	}
 
 	if err := em.saveConsole(); err != nil {
-		return err
+		plog.Errorf("Error saving console for device %v: %v", em.ID(), err)
 	}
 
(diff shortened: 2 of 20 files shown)


ajeddeloh:
Per OOB discussion: Updated to make Destory() void in all cases and instead responsible for it's own logging. We rarely check Destoy()'s return value and even when it does return an error all we do is log it. This also makes doing `defer x.Destroy()` not silently omit errors from the logs.

bgilbert:
Please structure the PR in such a way that every commit will compile and run.

platform/platform.go
```diff
@@ -76,8 +77,9 @@ type Cluster interface {
 	GetDiscoveryURL(size int) (string, error)
 
 	// Destroy terminates each machine in the cluster and frees any other
-	// associated resources.
-	Destroy() error
+	// associated resources. It should log and failures; since they are not
```
bgilbert: `any`

platform/machine/oci/machine.go
```diff
@@ -61,25 +62,21 @@ func (om *machine) Reboot() error {
 	return platform.RebootMachine(om, om.journal, om.cluster.RuntimeConf())
 }
 
-func (om *machine) Destroy() error {
-	if err := om.saveConsole(); err != nil {
-		// log error, but do not fail to terminate instance
-		plog.Error(err)
+func (om *machine) Destroy() {
+	var err multierror.Error
```
bgilbert: Unused.

platform/machine/aws/machine.go
```diff
@@ -60,25 +60,21 @@ func (am *machine) Reboot() error {
 	return platform.RebootMachine(am, am.journal, am.cluster.RuntimeConf())
 }
 
-func (am *machine) Destroy() error {
+func (am *machine) Destroy() {
 	if err := am.cluster.api.TerminateInstances([]string{am.ID()}); err != nil {
-		return err
+		plog.Errorf("Error terminate instance %v: %v", am.ID(), err)
```
bgilbert: `terminating`

ajeddeloh:
> Please structure the PR in such a way that every commit will compile and run.

I can, but that will squash the entire PR into one or two very large commits. I figured having more logical seperation was better, but am not too opinionated.

bgilbert:
> I can, but that will squash the entire PR into one or two very large commits.

I think that's okay here.  IMO it's better to have one large commit that makes a simple change in many places, rather than a series of repetitive commits that are untestable.

platform/machine/esx/machine.go
```diff
@@ -61,28 +61,24 @@ func (em *machine) Reboot() error {
 	return platform.RebootMachine(em, em.journal, em.cluster.RuntimeConf())
 }
 
-func (em *machine) Destroy() error {
+func (em *machine) Destroy() {
 	if err := em.cluster.api.TerminateDevice(em.ID()); err != nil {
-		return err
+		plog.Errorf("Error terminating instance %v: %v", em.ID(), err)
```
bgilbert: `terminating device`

platform/machine/esx/machine.go
```diff
@@ -61,28 +61,24 @@ func (em *machine) Reboot() error {
...
 	if em.journal != nil {
-		if err := em.journal.Destroy(); err != nil {
-			return err
-		}
+		em.journal.Destroy()
 	}
 
 	if err := em.saveConsole(); err != nil {
-		return err
+		plog.Errorf("Error saving console for instance %v: %v", em.ID(), err)
```
bgilbert: `device`; same below

platform/machine/packet/machine.go
```diff
@@ -60,19 +60,16 @@ func (pm *machine) Reboot() error {
 	return platform.RebootMachine(pm, pm.journal, pm.cluster.RuntimeConf())
 }
 
-func (pm *machine) Destroy() error {
+func (pm *machine) Destroy() {
 	if err := pm.cluster.api.DeleteDevice(pm.ID()); err != nil {
-		return err
+		plog.Errorf("Error terminating instance %v: %v", pm.ID(), err)
```
bgilbert: `device`

PR #781, opened 2017-12-07
kola/tests/locksmith: Set the DNS name in the X.509 SANs

dm0-:
[Go 1.9 ignores the certificate subject's CN when any SANs are defined](https://golang.org/doc/go1.9#crypto/x509), so when locksmith is built with 1.9, it requires both the IP address and DNS name as SANs.

This is required for coreos/coreos-overlay#2941.

### DIFF
--- a/kola/tests/locksmith/locksmith.go
+++ b/kola/tests/locksmith/locksmith.go
@@ -112,7 +112,7 @@ func init() {
       {
         "filesystem": "root",
         "path": "/etc/ssl/openssl.cnf",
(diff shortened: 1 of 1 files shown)


dm0-:
It is a potential issue, but we're going to have to go through this change at some point if we want to build our packages with a supported Go version.  Currently, an alternative is to default to Go 1.8, which will receive security updates until 1.10 is released.

bgilbert:
We should be sure to add a release note.

PR #782, opened 2017-12-11
kola/tests: disable coreos.etcd-member.v2-backup-restore on OCI

arithx:
The test requires ct rendering which is not available for the OCI
platform.


arithx:
Good to merge

PR #783, opened 2017-12-12
cmd/ore/packet: add fields required to instantiate API

arithx:
To instantiate a Packet API you're required to provide a working
gcloud JSON key & bucket storage URL. To be able to run ore for
Packet on things like Jenkins where we don't have these in the
pre-configured locations we need to add these options and plumb
them through.


arithx:
@dm0- Currently we're only cleaning up instances, this is more of a byproduct of the way that the API is instantiated requiring the Google JSON Key & bucket URL.

In the process of updating this PR to use the same variables as are in `bin/ore packet create-device` and move them up to the top-level.

arithx:
Per the comment about potentially making the parameters optional, I discussed this with @bgilbert OOB and while it's definitely possible we could make them optional the workflow for creating devices/clusters gets a bit more cluttered as they'd need to add an additional call to instantiate the gcloud API & bucket in between the API creation & device creation steps.

arithx:
Good to merge

PR #790, opened 2018-01-19
kola/test/kubernetes/kubelet_wrapper: fix flake

ajeddeloh:
Fix flake due to race preventing the unit name being applied to logs
that we grep through. Additionally update the test to not use deprecated
environment variables and use the current version of kubernetes.

See https://github.com/rkt/rkt/pull/2645 for more details about the
race.

### DIFF
--- a/kola/tests/kubernetes/kubelet_wrapper.go
+++ b/kola/tests/kubernetes/kubelet_wrapper.go
@@ -20,8 +20,8 @@ import (
 	"github.com/coreos/mantle/platform/conf"
 )
 
-const hyperkubeTag = "v1.5.7_coreos.0"
-const versionOutput = "Kubernetes v1.5.7+coreos.0" // --version for /hyperkube
+const hyperkubeTag = "v1.9.2_coreos.0"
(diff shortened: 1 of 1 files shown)


kola/tests/kubernetes/kubelet_wrapper.go
```diff
@@ -63,10 +63,10 @@ func testKubeletWrapperVarLog(c cluster.TestCluster) {
 	c.MustSSH(m, `
 	for i in {1..120}; do 
 		sleep 5
-		if journalctl -u kubelet -o cat | grep '`+versionOutput+`' &>/dev/null; then
+		if journalctl -t kubelet-wrapper -o cat | grep '`+versionOutput+`' &>/dev/null; then
```
dm0-: While we're modifying this, I think it would be nice to use `grep -F` since the pattern contains regex characters.  It shouldn't matter with this specific case, but it will make things look a bit saner.

ajeddeloh: good idea. Fixed.

PR #794, opened 2018-01-24
platform/machine/do: retry creating droplets

ajeddeloh:
Droplet creation frequently fails because DO returns a error 422 and
says try again. Take their advice and try again if droplet creation
fails.

### DIFF
--- a/platform/machine/do/cluster.go
+++ b/platform/machine/do/cluster.go
@@ -20,13 +20,16 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/coreos/pkg/capnslog"
+	"github.com/digitalocean/godo"
 
 	ctplatform "github.com/coreos/container-linux-config-transpiler/config/platform"
 	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/platform/api/do"
 	"github.com/coreos/mantle/platform/conf"
+	"github.com/coreos/mantle/util"
 )
 
 var (
@@ -87,8 +90,15 @@ func (dc *cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error)
 		return nil, err
 	}
 
-	droplet, err := dc.api.CreateDroplet(context.TODO(), dc.vmname(), dc.sshKeyID, conf.String())
-	if err != nil {
+	var droplet *godo.Droplet
+	// DO frequently gives us 422 errors saying "Please try again"
+	if err := util.RetryConditional(6, 10*time.Second, shouldRetry, func() error {
+		droplet, err = dc.api.CreateDroplet(context.TODO(), dc.vmname(), dc.sshKeyID, conf.String())
+		if err != nil {
+			plog.Errorf("Error creating droplet: %v, retrying...", err)
+		}
+		return err
+	}); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


platform/machine/do/cluster.go
```diff
@@ -87,8 +90,12 @@ func (dc *cluster) NewMachine(userdata *conf.UserData) (platform.Machine, error)
...
 	}
 
-	droplet, err := dc.api.CreateDroplet(context.TODO(), dc.vmname(), dc.sshKeyID, conf.String())
-	if err != nil {
+	var droplet *godo.Droplet
+	// DO frequently gives us 422 errors saying "Please try again"
+	if err := util.Retry(6, 10*time.Second, func() error {
+		droplet, err = dc.api.CreateDroplet(context.TODO(), dc.vmname(), dc.sshKeyID, conf.String())
+		return err
+	}); err != nil {
```
arithx: We should only retry if it's the 422 error.

ajeddeloh: I don't know. What if they return an error thats not 422 but still says retry? I don't see a ton of harm in retrying even if it'll just fail again. This would also retry in the case of 500 series errors, which imo we should do.

What about checking if its a `godo.ErrorResponse` (should cover any error returned from the api but not anything from the godo library itself)?

I'll also add logging, so we'll have a chance of seeing any errors.

arithx: At a minimum we definitely need to log the errors we're getting before doing a retry. When discussing retries in the past with @bgilbert our thought was to only add a specific set of errors in the case of platform/api level retries that we know is a platform issue (availability, known flakes with their deployment) so that we don't accidentally suppress real failures.

ajeddeloh: In the case of "not out fault" errors I think the correct action is to retry a couple times, then skip the test if it still fails. Those two are unrelated AFAICT. For the former (what is being addressed here) I think I'll add a `func shouldRetry(err error) bool` and maybe a `util.RetryConditional()` function.

For the latter (would be a seperate PR), it might be good to add a `IsCloudError(err error) bool` (not married to that name)  to the cluster interface that the harness can use to determine if the test should be skipped?

arithx: I agree, as long as we scope down the errors that trigger `shouldRetry` to be true and add logging denoting the error and that a retry is occuring.

@bgilbert any thoughts on whether 500's are acceptable for retries? My initial thought is that they probably are okay to be retried initially and we can potentially scope it back further if we find out that legitimate errors are being suppressed.

ajeddeloh:
Should log and only retry on 422's now, not that I can really test that.

platform/machine/do/cluster.go
```diff
@@ -147,3 +157,13 @@ func (dc *cluster) Destroy() {
...
 }
+
+// shouldRetry returns if the error is from DigitalOcean and we should
+// retry the request which generated it
+func shouldRetry(err error) bool {
+	errResp, ok := err.(*godo.ErrorResponse)
+	if !ok {
+		return false
+	}
+	return errResp.Response.StatusCode == 422
```
euank: You typically want to retry on 5xx errors; want to add those in too while we're at it?

ajeddeloh: Sure, done.

platform/machine/do/cluster.go
bgilbert: Absent platform documentation listing retryable errors, I'd generally prefer retrying only on errors we've specifically encountered and evaluated in the past, but I'm not tied to it.

The problem with something like `IsCloudError` is that we stringify and wrap errors all over the place.  It'd be easy to accidentally smash an error type during a refactor.

PR #795, opened 2018-01-26
kola/tests: Drop fleet; switch etcd2 dependencies to etcd-member

bgilbert:
https://coreos.com/blog/toward-etcd-v3-in-container-linux.html
https://coreos.com/blog/migrating-from-fleet-to-kubernetes.html

### DIFF
--- a/cmd/plume/specs.go
+++ b/cmd/plume/specs.go
@@ -279,7 +279,7 @@ var (
 				Container:         "publish",
 				Environments:      azureEnvironments,
 				Label:             "CoreOS Alpha",
-				Description:       "The Alpha channel closely tracks current development work and is released frequently. The newest versions of docker, etcd and fleet will be available for testing.",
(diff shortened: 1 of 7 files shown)


arithx:
I'm a bit hestitant to drop the testing for beta & stable.

For fleet I feel like we should set the `MaxVersion` for now and drop them once stable progresses past that.

The `etcd2` -> `etcd-member` change is a bit different, while I'm not particularly a fan of duplicating them it might be worthwhile in the interim to set `MaxVersion` on the current tests, create new tests which use `etcd-member` and a `MinVersion`, and then merge them down once stable progresses past the `MaxVersion` for the `etcd2` tests.

bgilbert:
@arithx I agree.  I completely forgot about beta and stable when making these changes.  Will fix.

bgilbert:
Set EndVersion on etcd2 and fleet tests. Forked all etcd2-using tests except `google.kubernetes.*`, which would have required more complexity than it seems to be worth.

PR #796, opened 2018-01-27
cmd/kola: teach spawn to put local SSH keys into userdata

bgilbert:
To use, run `kola spawn -k`.

Also add `kola spawn {-t,--detach}`, which is sugar for starting a machine, injecting local SSH keys, printing its ID and IP, and exiting.

### DIFF
--- a/cmd/kola/spawn.go
+++ b/cmd/kola/spawn.go
@@ -43,6 +43,7 @@ var (
 
 	spawnNodeCount      int
 	spawnUserData       string
+	spawnDetach         bool
 	spawnShell          bool
 	spawnRemove         bool
 	spawnVerbose        bool
@@ -54,6 +55,7 @@ var (
 func init() {
 	cmdSpawn.Flags().IntVarP(&spawnNodeCount, "nodecount", "c", 1, "number of nodes to spawn")
 	cmdSpawn.Flags().StringVarP(&spawnUserData, "userdata", "u", "", "file containing userdata to pass to the instances")
+	cmdSpawn.Flags().BoolVarP(&spawnDetach, "detach", "t", false, "-kv --shell=false --remove=false")
 	cmdSpawn.Flags().BoolVarP(&spawnShell, "shell", "s", true, "spawn a shell in an instance before exiting")
 	cmdSpawn.Flags().BoolVarP(&spawnRemove, "remove", "r", true, "remove instances after shell exits")
 	cmdSpawn.Flags().BoolVarP(&spawnVerbose, "verbose", "v", false, "output information about spawned instances")
@@ -72,6 +74,13 @@ func runSpawn(cmd *cobra.Command, args []string) {
 
 func doSpawn(cmd *cobra.Command, args []string) error {
 	var err error
+
+	if spawnDetach {
+		spawnSetSSHKeys = true
+		spawnVerbose = true
+		spawnShell = false
+		spawnRemove = false
+	}
 
 	if spawnNodeCount <= 0 {
 		return fmt.Errorf("Cluster Failed: nodecount must be one or more")
--- a/cmd/kola/spawn.go
+++ b/cmd/kola/spawn.go
@@ -1,4 +1,4 @@
-// Copyright 2015 CoreOS, Inc.
+// Copyright 2015-2018 CoreOS, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


ajeddeloh:
Yes yes yes!

I haven't looked into what the go library support for this would be, but it would be better to connect to the ssh-agent or maybe do that and read the files in .ssh? Just reading the files wont support users (like me) with gpg-derived ssh keys.

bgilbert:
You could copy your public key to `~/.ssh/id_rsa.pub` (I do) but it's a good point.  Let me think about that.

euank:
I think using the ssh-agent would be nice; if you don't want to implement it, I will as a followup.

I just checked whether it's easy to talk to the agent in golang, and it seems pretty simple; here's the [sample](https://github.com/euank/kitchen-sink/blob/master/snippets/go-ssh-add-l/main.go#L30) program I wrote which prints public keys.

platform/conf/conf.go
```diff
@@ -405,6 +421,13 @@ func (c *Conf) copyKeysIgnitionV21(keys []*agent.Key) {
 	for _, key := range keys {
 		keyObjs = append(keyObjs, v21types.SSHAuthorizedKey(key.String()))
 	}
+	for i := range c.ignitionV21.Passwd.Users {
```
euank: note, in all these loops we can use `for _, user :=` ... The usual iter-value-reused thing doesn't apply to us here because of the early return.

I think it is a bit more readable to do that in all these loops

cmd/kola/spawn.go
```diff
@@ -38,19 +42,30 @@ var (
...
 
 func init() {
+	defaultKeys, err := defaultSSHKeyPaths()
+	if err != nil {
+		plog.Warning(err)
+	}
+
 	cmdSpawn.Flags().IntVarP(&spawnNodeCount, "nodecount", "c", 1, "number of nodes to spawn")
 	cmdSpawn.Flags().StringVarP(&spawnUserData, "userdata", "u", "", "file containing userdata to pass to the instances")
+	cmdSpawn.Flags().BoolVarP(&spawnDetach, "detach", "t", false, "-v --shell=false --remove=false, plus -k if -u not set")
```
euank: I can see how this flag is useful, but the `-k` conditionally defaulting is kinda weird. Maybe `-k` should always default true, and have the restriction of conflicting with `-u` removed?

I'm not going to argue particularly strongly since it only seems a little odd

cmd/kola/spawn.go
```diff
@@ -64,12 +79,28 @@ func runSpawn(cmd *cobra.Command, args []string) {
...
+	}
+
 	if spawnNodeCount <= 0 {
 		return fmt.Errorf("Cluster Failed: nodecount must be one or more")
 	}
 
 	var userdata *conf.UserData
-	if spawnUserData != "" {
+	if spawnSetSSHKeys && spawnUserData != "" {
+		return fmt.Errorf("Cannot set SSH keys if userdata is specified")
```
euank: It seems like we could support `-k -u...` by merging in the user's ssh keys. Is there a reason not to do that? Just simplicity?

It is true that if `-u` is used it generally should already have ssh keys, but it doesn't seem like they *have* to conflict

platform/conf/conf.go
bgilbert: That's what I had first and it was broken. Those are slices of structs, not slices of references to structs, and we're modifying one of them in place.

euank: Oh, whoops. Missed that '&' there, you're right.

bgilbert:
Updated.

cmd/kola/spawn.go
```diff
@@ -133,3 +163,58 @@ func doSpawn(cmd *cobra.Command, args []string) error {
...
+		}
+
+		// populate list of key files
+		userInfo, err := user.Current()
+		if err != nil {
+			return nil, err
+		}
+		for _, name := range []string{"id_rsa.pub", "id_dsa.pub", "id_ecdsa.pub", "id_ed25519.pub"} {
+			path := filepath.Join(userInfo.HomeDir, ".ssh", name)
+			if _, err := os.Stat(path); !os.IsNotExist(err) {
```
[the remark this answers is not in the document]
bgilbert: In that case, the current code causes an error an when we try to read the file later on, which I think is the right thing?  The alternative is silently ignoring key files that are broken in some way.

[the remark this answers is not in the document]
euank: Since this is in "guess the keys" code anyways, it seems to me like silently ignoring fs errors fits, but my opinion isn't all that strong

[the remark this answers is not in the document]
bgilbert: Okay, updated.

PR #797, opened 2018-01-29
Fix coreos.cloudinit.basic and coreos.ignition.v2_1.resource.local on Packet

bgilbert:


platform/base.go
```diff
@@ -45,14 +45,15 @@ type BaseCluster struct {
 
 	name       string
```
arithx: Could also switch this from `string` to `platform.Name`

bgilbert: That's the cluster name, not the platform name.  The intent of `platform.Name` is that it should be treated as an enumerated type.

PR #801, opened 2018-01-29
kola/tests/misc/update: mask update-engine

ajeddeloh:
This fixes a race where update-engine sets the boot partition back to USR-A after the test sets it to USR-B.

At least we think that's what going on. This seems to have fixed it locally, but the flakyness of the tests makes it hard to know.


ajeddeloh:
Tested this by launching a bunch of kola runs in parallel, all 54 that spawned correctly (hit some rate limiting) passed.

PR #804, opened 2018-02-02
platform/do: move retry logic to platform

ajeddeloh:
The api package wraps the error from DO so the shouldRetry logic doesn't
detect the DO errors correctly. Move the logic to the api package so it
can correctly detect the error type.


ajeddeloh:
I mean I ran it and I can still `kola spawn`, but I can't test DO giving me a 422 in any real sense.

arithx:
Being able to spawn is what I meant.

PR #805, opened 2018-02-02
kola/tests: add test for MachineID

### LINKED ISSUE #568: kola: Check /etc/machine-id
Check that it is not the COREOS_BLANK_MACHINE_ID or empty.

arithx:
Fixes #568

### DIFF
--- a/kola/tests/coretest/core.go
+++ b/kola/tests/coretest/core.go
@@ -41,6 +41,7 @@ func init() {
 			"ReadOnly":         TestReadOnlyFs,
 			"RandomUUID":       TestFsRandomUUID,
 			"Useradd":          TestUseradd,
+			"MachineID":        TestMachineID,
 		},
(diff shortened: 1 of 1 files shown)


kola/tests/coretest/core.go
```diff
@@ -337,3 +338,12 @@ func TestUseradd() error {
 
 	return nil
 }
+
+// Test that /etc/machine-id isn't empty or COREOS_BLANK_MACHINE_ID
+func TestMachineID() error {
+	id := MachineID()
+	if id == "" || id == "COREOS_BLANK_MACHINE_ID" {
+		return fmt.Errorf("machine-id is %s", id)
```
ajeddeloh: I'd split this into two if's becuase if it is an empty string this will be a very confusing error.

arithx:
Updated

PR #807, opened 2018-02-05
kola/tests: disable Ignition remote resource fetches on DO

bgilbert:
DigitalOcean doesn't have proper networking in the initramfs. This is a known issue and there's no point cluttering our test logs.

cc @arithx @ajeddeloh


arithx:
Do we have a bug for networking in DO's initramfs open?

bgilbert:
Yup, coreos/bugs#2205.

PR #810, opened 2018-02-06
Don't create PV AMIs in regions that don't support them

bgilbert:
Add `eu-west-3` and `cn-northwest-1`, which don't, to plume.  Add PV AKI ID for `cn-north-1`, which does.

### DIFF
--- a/cmd/plume/prerelease.go
+++ b/cmd/plume/prerelease.go
@@ -425,7 +425,7 @@ func awsUploadToPartition(spec *channelSpec, part *awsPartitionSpec, imageName, 
 		return nil, nil, fmt.Errorf("couldn't tag images: %v", err)
 	}
 
-	postprocess := func(imageID string) (map[string]string, error) {
+	postprocess := func(imageID string, pv bool) (map[string]string, error) {
 		if len(part.LaunchPermissions) > 0 {
 			if err := api.GrantLaunchPermission(imageID, part.LaunchPermissions); err != nil {
 				return nil, err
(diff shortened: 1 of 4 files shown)


platform/api/aws/images.go
```diff
@@ -439,6 +454,14 @@ func (a *API) CopyImage(sourceImageID string, regions []string) (map[string]stri
 		return nil, err
 	}
 
+	if *image.VirtualizationType == "paravirtual" {
+		for _, region := range regions {
+			if _, ok := akis[region]; !ok {
```
euank: Might as well reuse the `RegionSupportsPV` helper for readability here.

platform/api/aws/images.go
```diff
@@ -439,6 +454,14 @@ func (a *API) CopyImage(sourceImageID string, regions []string) (map[string]stri
 		return nil, err
 	}
 
+	if *image.VirtualizationType == "paravirtual" {
```
euank: `ec2.VirtualizationTypeParavirtual` looks like the correct const to use here.

cmd/plume/prerelease.go
```diff
@@ -436,7 +436,9 @@ func awsUploadToPartition(spec *channelSpec, part *awsPartitionSpec, imageName,
 		foundBucketRegion := false
 		for _, region := range part.Regions {
 			if region != part.BucketRegion {
-				destRegions = append(destRegions, region)
+				if !pv || aws.RegionSupportsPV(region) {
+					destRegions = append(destRegions, region)
+				}
```
euank: Perhaps we should have an "else if pv" that logs to indicate the region pv's being skipped for?

bgilbert:
Updated.

PR #811, opened 2018-02-06
readme: improve oci docs

ajeddeloh:


README.md
```diff
@@ -263,9 +263,10 @@ tenancy=tenancy.ocid.goes.here
 region=region-goes-here
 ```
 
-The `oci setup config` command will generate the key file automatically.
+The `oci setup config` command will generate the key file automatically. You must add your public key
+(`~/.oci/oci_api_key_public.pem`) to your oci user in the oci web console.
 
-The `~/.oraclebmc/config.mantle` file is only used by Mantle and must be populated manually:
+The `~/.oci/config.mantle` file is only used by Mantle and must be populated manually:
```
arithx: Should we note that this file is optional, and is meant mostly as to allow custom fields (such as `compartment`)

ajeddeloh: I didn't know it was optional. Do you not need to specify the compartment?

arithx: `compartment` is required by `kola` but is not required by the `oci` tool which uses the base config. You can however also just put the config key inside of the `~/.oci/config` file and it will still parse.

README.md
```diff
@@ -263,9 +263,10 @@ tenancy=tenancy.ocid.goes.here
 region=region-goes-here
 ```
 
-The `oci setup config` command will generate the key file automatically.
+The `oci setup config` command will generate the key file automatically. You must add your public key
+(`~/.oci/oci_api_key_public.pem`) to your oci user in the oci web console.
```
bgilbert: Does `oci setup` not do that for you?  I don't remember ever doing it by hand.

bgilbert: OK, I may be misremembering or the tool may have changed.

Issue #622, opened 2017-05-31
cork enter --experimental doesn't properly expose gpg agent socket


bgilbert:
It tries to bind the socket into `/run/user/UID/agent-NNNNNN/` and point `GPG_AGENT_INFO` at it.  `GPG_AGENT_INFO` is ignored by gpg &ge; 2.1, and gpg &ge; 2.1.13 expects to find the agent socket in `/run/user/UID/gnupg`.

ajeddeloh:
Closed via https://github.com/coreos/mantle/pull/812

PR #812, opened 2018-02-09
cork: pass through gpg sockets

ajeddeloh:
bind mount ~/.gnupg to the same directory in the sdk. Mounting it other
places causes gpg to ignore the /run/user/$uid/gnupg directory and
create temp directories instead, although it is unclear why.

Always mount the /run/user/$uid/gnupg directory to the same location in
the sdk becuase configuring gpg to use a different directory is no
longer as simple as using the GPG_AGENT_INFO environment variable.

Addresses https://github.com/coreos/mantle/issues/622

### DIFF
--- a/sdk/enter.go
+++ b/sdk/enter.go
@@ -153,61 +153,65 @@ func (e *enter) MountAPI() error {
 	return nil
 }
 
-// MountAgent bind mounts a SSH or GnuPG agent socket into the chroot
-func (e *enter) MountAgent(env string) error {
-	origPath := os.Getenv(env)
+// MountAgent bind mounts a SSH agent socket into the chroot
+func (e *enter) mountSSHAgent() error {
+	origPath := os.Getenv("SSH_AUTH_SOCK")
 	if origPath == "" {
 		return nil
 	}
 
-	origDir, origFile := filepath.Split(origPath)
+	origDir := filepath.Dir(origPath)
 	if _, err := os.Stat(origDir); err != nil {
 		// Just skip if the agent has gone missing.
 		return nil
 	}
 
-	newDir, err := ioutil.TempDir(e.UserRunDir, "agent-")
-	if err != nil {
+	newDir := filepath.Join(e.Chroot, origDir)
+	if err := os.Mkdir(newDir, 0700); err != nil && !os.IsExist(err) {
 		return err
 	}
 
-	if err := system.Bind(origDir, newDir); err != nil {
-		return err
-	}
-
-	newPath := filepath.Join(newDir, origFile)
-	chrootPath := strings.TrimPrefix(newPath, e.Chroot)
-	return os.Setenv(env, chrootPath)
+	return system.Bind(origDir, newDir)
 }
 
 // MountGnupg bind mounts $GNUPGHOME or ~/.gnupg and the agent socket
 // if available. The agent is ignored if the home dir isn't available.
-func (e *enter) MountGnupg() error {
+func (e *enter) mountGnupg() error {
 	origHome := os.Getenv("GNUPGHOME")
 	if origHome == "" {
 		origHome = filepath.Join(e.User.HomeDir, ".gnupg")
 	}
 
 	if _, err := os.Stat(origHome); err != nil {
-		// Skip but do not pass along $GNUPGHOME
-		return os.Unsetenv("GNUPGHOME")
+		// Skip but do not bind mount anything
+		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


sdk/enter.go
```diff
@@ -188,26 +188,30 @@ func (e *enter) MountGnupg() error {
 		origHome = filepath.Join(e.User.HomeDir, ".gnupg")
 	}
 
-	if _, err := os.Stat(origHome); err != nil {
-		// Skip but do not pass along $GNUPGHOME
-		return os.Unsetenv("GNUPGHOME")
+	// gpg misbehaves in the sdk with GNUPGHOME set to anything but ~/.gnupg
+	// so always unset it
+	if err := os.Unsetenv("GNUPGHOME"); err != nil {
```
arithx: Does this remain unset after cork exits?

ajeddeloh: Tested and nope. Only for this process (and things it launches?)

ajeddeloh:
Updated to also not be clever with SSH_AUTH_SOCK

sdk/enter.go
```diff
@@ -153,61 +153,59 @@ func (e *enter) MountAPI() error {
...
-func (e *enter) MountAgent(env string) error {
+// MountAgent bind mounts a SSH agent socket into the chroot
+func (e *enter) mountAgent(env string) error {
 	origPath := os.Getenv(env)
 	if origPath == "" {
 		return nil
 	}
 
-	origDir, origFile := filepath.Split(origPath)
+	origDir, _ := filepath.Split(origPath)
```
euank: Also known as `filepath.Dir`

sdk/enter.go
```diff
@@ -153,61 +153,59 @@ func (e *enter) MountAPI() error {
...
 
+	newHome := filepath.Join(e.Chroot, e.User.HomeDir, ".gnupg")
 	if err := system.Bind(origHome, newHome); err != nil {
 		return err
 	}
 
-	chrootHome := strings.TrimPrefix(newHome, e.Chroot)
-	if err := os.Setenv("GNUPGHOME", chrootHome); err != nil {
+	// gpg expects the socket at /run/user/$uid/gnupg
+	origAgentDir := filepath.Join("/run", "user", e.User.Uid, "gnupg")
```
euank: Should we check for the existence of this directory? If the user isn't running the gpg-agent, I'd expect it to not exist and this to error out.

sdk/enter.go
```diff
@@ -365,11 +363,11 @@ func enterChrootHelper(args []string) (err error) {
 		return err
 	}
 
-	if err := e.MountAgent("SSH_AUTH_SOCK"); err != nil {
+	if err := e.mountAgent("SSH_AUTH_SOCK"); err != nil {
```
euank: It seems weird to pass in an env var name here, but to assume GNUPG home in mountGnupg a little below.

I'd rather just not pass it in and let the function assume for consistency; it's not really buying us anything to abstract it.

ajeddeloh:
Updated with @euank's suggestions.

sdk/enter.go
```diff
@@ -153,61 +153,65 @@ func (e *enter) MountAPI() error {
...
-func (e *enter) MountGnupg() error {
+func (e *enter) mountGnupg() error {
 	origHome := os.Getenv("GNUPGHOME")
 	if origHome == "" {
 		origHome = filepath.Join(e.User.HomeDir, ".gnupg")
 	}
 
 	if _, err := os.Stat(origHome); err != nil {
-		// Skip but do not pass along $GNUPGHOME
-		return os.Unsetenv("GNUPGHOME")
```
bgilbert: Why are we passing through`GNUPGHOME` in this case?

ajeddeloh: We're not. It's unconditionally unset below.

bgilbert: Not after returning it isn't.

ajeddeloh: Oh, hmm. I guess it does make sense to unconditionally unset it before any returns could happen. It's not doing anything with gpg in that case, but it would be set to a path from the real root which might make gpg confused if the user has their own gpg setup inside the sdk.

Looking at this more I'm somewhat confused what is controlling which environment variables are passed through to the chroot and which aren't. If I set `FOOBAR=BARFOO` then `cork enter` and `FOOBAR` is unset. `GNUPGHOME` does stay set though. 😕

bgilbert: https://github.com/coreos/mantle/blob/78e6c559eddbcd1653c67078a90d90af72f3e432/sdk/create.go#L67-L75

PR #814, opened 2018-02-12
cmd/cork: switch enter default to new enter

ajeddeloh:
Change `cork enter`'s default to be what used to be the --experimental
flag. Rename the flag to --legacy and make that use the old enter
method.

Keep the experimental flag, but make it no-op and warn that it will go
away soon.

Doesn't strictly depend on https://github.com/coreos/mantle/pull/812 but that should probably be merged first.

### DIFF
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -48,6 +48,8 @@ var (
 	allowReplace bool
 
 	// only for `enter` command
+	legacy bool
+	// only for `enter` command, deprecated.
 	experimental bool
 
 	// only for `update` command
@@ -123,8 +125,11 @@ func init() {
 	root.AddCommand(createCmd)
 
 	enterCmd.Flags().AddFlagSet(chrootFlags)
+	enterCmd.Flags().BoolVar(&legacy,
+		"legacy", false, "Use old enter implementation")
 	enterCmd.Flags().BoolVar(&experimental,
-		"experimental", false, "Use new enter implementation")
+		"experimental", true, "Do not use. Kept for compatibility")
+	enterCmd.Flags().MarkDeprecated("experimental", "--experimental is now default. This flag will be removed soon. Use --legacy for the old enter method.")
 	root.AddCommand(enterCmd)
 
 	deleteCmd.Flags().AddFlagSet(chrootFlags)
@@ -262,9 +267,9 @@ func updateRepo() {
 }
 
 func runEnter(cmd *cobra.Command, args []string) {
-	enter := sdk.OldEnter
-	if experimental {
-		enter = sdk.Enter
+	enter := sdk.Enter
+	if legacy || !experimental {
+		enter = sdk.OldEnter
 	}
 
 	if err := enter(chrootName, args...); err != nil && len(args) != 0 {


cmd/cork/create.go
```diff
@@ -123,8 +125,10 @@ func init() {
 	root.AddCommand(createCmd)
 
 	enterCmd.Flags().AddFlagSet(chrootFlags)
+	enterCmd.Flags().BoolVar(&legacy,
+		"legacy", false, "Use old enter implementation")
 	enterCmd.Flags().BoolVar(&experimental,
-		"experimental", false, "Use new enter implementation")
+		"experimental", false, "No-op. do not use. Kept for compatibilty")
```
bgilbert: Maybe use `MarkHidden()`?

cmd/cork/create.go
```diff
@@ -262,9 +266,12 @@ func updateRepo() {
 }
 
 func runEnter(cmd *cobra.Command, args []string) {
-	enter := sdk.OldEnter
 	if experimental {
-		enter = sdk.Enter
+		plog.Warningf("--experimental is deprecated and does nothing. Fix your scripts.")
```
bgilbert: The option is deprecated but the mode is now the default.  Also, the second sentence is redundant -- and there's no point in yelling at users who have been helpfully testing the new mode.

cmd/cork/create.go
ajeddeloh: Huh, didn't know that existed. Maybe `MarkDeprecated()` instead. Come to think of it, I think it will techincally break anyone using `cork enter --experimental=false`, but I'm not sure if we care about that?

cmd/cork/create.go
ajeddeloh: I didn't mean to yell =(. I meant to just add some urgency since this doesn't exactly have a defined deprecation window. Maybe "--experimental is deprecated and does nothing. It will be removed next release" or "Update your scripts." (Since any scripts aren't _broken_ really, but do need to be updated)

ajeddeloh:
Updated to remove "fix your scripts", make --experimental still work.

cmd/cork/create.go
```diff
@@ -48,6 +48,8 @@ var (
 	allowReplace bool
 
 	// only for `enter` command
+	legacy bool
+	// kept for backwards compatibility, does nothing
```
bgilbert: It does now...

cmd/cork/create.go
```diff
@@ -123,8 +125,11 @@ func init() {
 	root.AddCommand(createCmd)
 
 	enterCmd.Flags().AddFlagSet(chrootFlags)
+	enterCmd.Flags().BoolVar(&legacy,
+		"legacy", false, "Use old enter implementation")
 	enterCmd.Flags().BoolVar(&experimental,
-		"experimental", false, "Use new enter implementation")
+		"experimental", true, "Do not use. Kept for compatibilty")
```
bgilbert: `compatibility`, though this doesn't seem to be shown in the help anyway.

cmd/cork/create.go
bgilbert: We can remove legacy mode without removing `--experimental` (at least for one value of `--experimental`, namely `true`) so I don't think there's a lot of urgency anyway.

PR #815, opened 2018-02-13
sdk/enter: unconditionally unset GNUGHOME

ajeddeloh:
Always unset GNUPGHOME since if there is a ~/.gnupg we're bind mounting
then we're always binding it to the default for GNUPGHOME and if there's
not, we don't want to touch GPG at all, so we should not carry over the
environment variable.


euank:
Alternately it could be removed from `env_keep` I think, which would be cleaner (based on @bgilbert's comment https://github.com/coreos/mantle/pull/812#discussion_r167724666)

ajeddeloh:
That won't handle existing SDKs. It should probably be removed from there as well, though.

ajeddeloh:
Updated to drop it from `env_keep` as well. Added a comment about it in the entry code.

sdk/enter.go
```diff
@@ -182,17 +182,19 @@ func (e *enter) mountGnupg() error {
...
 
-	if _, err := os.Stat(origHome); err != nil {
-		// Skip but do not bind mount anything
-		return nil
-	}
-
 	// gpg misbehaves in the sdk with GNUPGHOME set to anything but ~/.gnupg
 	// so always unset it so the default ~/.gnupg is used.
+	// This is no longer in the env_keep setup for sudo when creating the
+	// sdk, but that won't work with existing sdk's.
```
bgilbert: `that won't work with` doesn't parse cleanly for me.  What about:

> gpg misbehaves in the SDK with GNUPGHOME set to anything but ~/.gnupg.  New SDKs don't encounter this problem because GNUPGHOME is no longer included in the sudo env_keep list.  For old SDKs, unset GNUPGHOME so the default ~/.gnupg is used.

ajeddeloh:
Updated comment, and added a commit because it looks like entering new SDKs broke (since ~/.gnupg may not exist).

PR #816, opened 2018-02-13
Revert "sdk/create: drop gnupg env vars from env_keep"

ajeddeloh:
This reverts commit 174af1561700bfeb63d90bb60c3bd4e73e31d6bf.

This breaks Jenkins signing.

cc @dm0-

PR #817, opened 2018-02-13
Revert gpg

ajeddeloh:
This reverts mantle to the state of 78e6c559eddbcd1653c67078a90d90af72f3e432.

Passing the gpg agent is causing Jenkins failures, so revert this for releases until we can track down what's going on.


bgilbert:
Reverts #814, #815, and #816.

PR #818, opened 2018-02-13
Revert "cork: pass through gpg sockets"

ajeddeloh:
This reverts commit 02e1f8c19eb703151086448e5bb8f9ccf91d1cde.

This should finish reverting all of the gpg bind mounting related changes.


bgilbert:
It does not bring mantle to the state of 5ceb160, because that would also exclude #813, #810, #811, #809, #808, #807, #804, #783, #805, #796, #801, and #797.

PR #821, opened 2018-02-14
platform/aws: automatically resolve dupe images

euank:
This is intended to automatically resolve duplicate image name errors
which frequently crop up during pre-release.

This change does not stop 'plume release' from failing if there
are multiple images. This is an error scenario we may wish to handle as
well, in which case `FindImage` can be removed entirely.
For now, I felt it safer to limit the changes to the pre-release
operations.

### DIFF
--- a/platform/api/aws/images.go
+++ b/platform/api/aws/images.go
@@ -631,6 +631,10 @@ func (a *API) findAndDedupeImage(name string) (string, error) {
 			ImageId: aws.String(ids[i]),
 		})
 		if err != nil {
+			if awserr, ok := err.(awserr.Error); ok && awserr.Code() == "InvalidAMIID.Unavailable" {
+				// This error happens if the image is already deregistered
+				continue
+			}
 			return "", fmt.Errorf("error trying to cleanup dupe images named %q (%v): %v", name, ids, err)
 		}
 	}
--- a/platform/api/aws/images.go
+++ b/platform/api/aws/images.go
@@ -382,7 +382,7 @@ func (a *API) createImage(params *ec2.RegisterImageInput) (string, error) {
 		// The AMI already exists. Get its ID. Due to races, this
 		// may take several attempts.
 		for {
-			imageID, err := a.FindDedupeImage(*params.Name)
+			imageID, err := a.findAndDedupeImage(*params.Name)
 			if err != nil {
 				return "", err
 			}
@@ -515,7 +515,7 @@ func (a *API) CopyImage(sourceImageID string, regions []string) (map[string]stri
 }
 
 func (a *API) copyImageIn(sourceRegion, sourceImageID, name, description string, imageTags, snapshotTags []*ec2.Tag, launchPermissions []*ec2.LaunchPermission) (string, error) {
-	imageID, err := a.FindDedupeImage(name)
+	imageID, err := a.findAndDedupeImage(name)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


euank:
Pending:

- [ ] A minimal amonut of manual testing against dev account

platform/api/aws/images.go
```diff
@@ -585,19 +586,68 @@ func (a *API) copyImageIn(sourceRegion, sourceImageID, name, description string,
...
-// Find an image we own with the specified name. Return ID or "".
+// FindDedupeImage finds an image given its name, but also attempts to
+// automatically handle duplicate private images. If multiple images, all
+// private, are found, it deregisters all but the first.
+// This is necessary because, although names are meant to be unique, use of the
+// "CopyImage" API semi-frequently results in duplicates.
+func (a *API) FindDedupeImage(name string) (string, error) {
+	images, err := a.findImages(name)
+	if err != nil || len(images) == 0 {
+		return "", err
```
arithx: Is it intentional that this returns `"", nil` if `len(images) == 0` && `err == nil`?

platform/api/aws/images.go
```diff
@@ -585,19 +586,68 @@ func (a *API) copyImageIn(sourceRegion, sourceImageID, name, description string,
...
 	}
 
-	return imageID, nil
+	return dedupedId, nil
 }
 
-// Find an image we own with the specified name. Return ID or "".
+// FindDedupeImage finds an image given its name, but also attempts to
+// automatically handle duplicate private images. If multiple images, all
+// private, are found, it deregisters all but the first.
```
arithx: Do we want to clarify that by first we mean first alphabetical rather than first uploaded?

platform/api/aws/images.go
euank: Yes, intentional.
I think this matches the previous `a.FindImage` behavior, so I assume all callers expect that as well.

platform/api/aws/images.go
```diff
@@ -585,19 +586,75 @@ func (a *API) copyImageIn(sourceRegion, sourceImageID, name, description string,
...
+	// Deterministically pick an id to keep
+	sort.Strings(ids)
+
+	plog.Infof("found multiple images named %q (%v): keeping %q", name, ids, ids[0])
+
+	for i := 1; i < len(ids); i++ {
+		_, err := a.ec2.DeregisterImage(&ec2.DeregisterImageInput{
+			ImageId: aws.String(ids[i]),
+		})
+		if err != nil {
```
euank: In practice, this does happen. From my attempt to test this in our dev account:

```
error trying to cleanup dupe images named "Container-Linux-developer-1702.1.0_jenkins2-build-1702_local-1246-hvm" ([ami-5a7de525 ami-e67be399]): InvalidAMIID.Unavailable: The image ID 'ami-e67be399' is no longer available
```

It looks like this error will have to be ignored since that happens pretty easily.

euank:
Minimal manual testing done in the form of parallel `ore aws copy-image` calls in our developer account.

Everything seems to work as I'd expect with the addition of an extra error check.

euank:
cc @bgilbert for review; I admit deregistering images seems a little scary, but it shouldn't be too bad...

I've only manually tested the `ore aws copy-image` code-path, but that matches the `plume pre-release` code closely enough it should be fine.

arithx (approved):
LGTM

euank:
Container Linux is deprecated. Closing this PR as stale.

PR #823, opened 2018-02-20
platform/machine/aws: always write console output

arithx:


arithx:
Still need investigation on the timeout issue. Still happening on my local machine even with the 10 minute bump.

arithx:
Updated

platform/machine/aws/machine.go
```diff
@@ -113,6 +124,20 @@ func (am *machine) saveConsole(origConsole string) error {
 		return true, nil
 	})
 	if err != nil {
+		if origConsole != "" {
+			if writeErr := am.writeConsole(origConsole); writeErr != nil {
```
bgilbert: This would be cleaner as:

```go
err := fmt.Errorf("retrieving console output of %v: %v", am.ID(), err)
if origConsole != "" {
    plog.Warning(err)
    // fall through, to save what we can
} else {
    return err
}
```

Then the writing code can remain unchanged, and you don't need `writeConsole`.  I don't know that it's necessary to distinguish partial output due to timeout from partial output due to other kinds of failures.

arithx: Agreed, updated.

platform/machine/aws/machine.go
```diff
@@ -113,7 +113,12 @@ func (am *machine) saveConsole(origConsole string) error {
 		return true, nil
 	})
 	if err != nil {
-		return fmt.Errorf("retrieving console output of %v: %v", am.ID(), err)
+		err = fmt.Errorf("retrieving console output of %v: %v", am.ID(), err)
+		if origConsole != "" {
+			plog.Warning(err)
```
bgilbert: I just realized: at this point `am.console` may equal `origConsole` or may be empty, and we're relying on the matching heuristic below to ensure that they're correctly merged.  It might be a good idea to add special cases below to ensure this is handled correctly.

arithx: https://play.golang.org/p/La1BAPAVT2O

It looks like it still works with these cases, was there any others that you had in mind?

ajeddeloh: Actually, I'd drop this to debug level. It's not really a useful thing to see most of the time unless debugging

bgilbert: It is an actual (non-fatal) error, and under other circumstances (tests that run in less than ~5 minutes) will be logged that way by the caller.

bgilbert: The second case was a thinko on my part; `LastIndex` will return the index of the end of the string, so we won't lose any of `origConsole`.  As to the first case, the heuristic may be confused if the first 100 chars of `am.console` occur more than once in `origConsole` (e.g. if there's log spam).  That's true anyway, but in this special case we have additional information we can use to avoid that confusion, namely that the two strings are identical.  I don't think it's a dealbreaker though.

PR #824, opened 2018-02-21
Add cork enter --legacy and --bind-gpg-agent

ajeddeloh:
Add cork enter `--legacy` and `--bind-gpg-agent`. Additionally refactor the enter code to make adding new similar options easier and more clear.

The changes in `sdk/repo` are untested at the moment and should be tested with Jenkins before we attempt a release with these changes, although I don't _think_ they should cause any problems.

### DIFF
--- a/sdk/enter.go
+++ b/sdk/enter.go
@@ -154,8 +154,8 @@ func (e *enter) MountAPI() error {
 }
 
 // MountAgent bind mounts a SSH or GnuPG agent socket into the chroot
-func (e *enter) MountAgent(env string) error {
-	origPath := os.Getenv(env)
+func (e *enter) MountSSHAgent() error {
+	origPath := os.Getenv("SSH_AUTH_SOCK")
 	if origPath == "" {
 		return nil
 	}
@@ -177,12 +177,12 @@ func (e *enter) MountAgent(env string) error {
 
 	newPath := filepath.Join(newDir, origFile)
 	chrootPath := strings.TrimPrefix(newPath, e.Chroot)
-	return os.Setenv(env, chrootPath)
+	return os.Setenv("SSH_AUTH_SOCK", chrootPath)
 }
 
 // MountGnupg bind mounts $GNUPGHOME or ~/.gnupg and the agent socket
 // if available. The agent is ignored if the home dir isn't available.
-func (e *enter) MountGnupg() error {
+func (e *enter) MountGnupgHome() error {
 	origHome := os.Getenv("GNUPGHOME")
 	if origHome == "" {
 		origHome = filepath.Join(e.User.HomeDir, ".gnupg")
@@ -193,8 +193,11 @@ func (e *enter) MountGnupg() error {
 		return os.Unsetenv("GNUPGHOME")
 	}
 
-	newHome, err := ioutil.TempDir(e.UserRunDir, "gnupg-")
-	if err != nil {
+	// gpg gets confused when GNUPGHOME isn't ~/.gnupg, so mount it there.
+	// Additionally, set the GNUPGHOME variable so commands run with sudo
+	// can also use it.
+	newHome := filepath.Join(e.Chroot, e.User.HomeDir, ".gnupg")
+	if err := os.Mkdir(newHome, 0700); err != nil && !os.IsExist(err) {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


ajeddeloh:
Oops, yeah the commit message is wrong, just fixed.

sdk/enter.go
```diff
@@ -197,29 +198,32 @@ func (e *enter) MountGnupgHome() error {
 	// Additionally, set the GNUPGHOME variable so commands run with sudo
 	// can also use it.
 	newHome := filepath.Join(e.Chroot, e.User.HomeDir, ".gnupg")
-	if err := os.Mkdir(newHome, 0700); err != nil {
+	if err := os.Mkdir(newHome, 0700); err != nil && !os.IsExist(err) {
```
arithx: Should the changes in this commit to `MountGnupugHome` be in the previous commit?

sdk/enter.go
```diff
@@ -197,29 +198,32 @@ func (e *enter) MountGnupgHome() error {
...
+	origAgentDir := filepath.Join("/run", "user", e.User.Uid, "gnupg")
 	if _, err := os.Stat(origAgentDir); err != nil {
 		// Skip
 		return nil
 	}
 
 	// gpg acts weird if this is elsewhere, so use /run/user/$uid/gnupg
 	newAgentDir := filepath.Join(e.Chroot, origAgentDir)
-	if err := os.Mkdir(newAgentDir, 0700); err != nil {
+	if err := os.Mkdir(newAgentDir, 0700); err != nil && !os.IsExist(err) {
```
arithx: Should this check be in the previous commit?

ajeddeloh:
Updated to move things to the correct commits.

ajeddeloh:
AFAIK we don't have a way of testing it on jenkins without merging, unless @dm0 has some tricks up his sleeve.

dm0-:
You can replay the last mantle job and insert a `sh "git fetch refs/pull/824/head ; git checkout FETCH_HEAD"` line after the `checkout scm` so it builds this PR.  After it starts building, you could also disable the job's polling trigger in case something else gets merged before the OS build finishes.  When done testing, build the mantle job normally with default parameters, and it will reset everything.

dm0-:
The test build looked fine, so I'm resetting the mantle build to normal for the nightlies.

sdk/repo.go
```diff
@@ -153,5 +157,10 @@ func RepoSync(chroot string, force bool) error {
 	if force {
 		args = append(args, "--force-sync")
 	}
-	return enterChroot(chroot, chrootRepoRoot, args...)
+	return enterChroot(enter{
+		RepoRoot: RepoRoot(),
+		Chroot:   filepath.Join(RepoRoot(), chroot),
+		CmdDir:   chrootRepoRoot,
```
bgilbert: `enterChroot` should default these fields if not set (or maybe override them entirely if they don't need to be customizable?), rather than open-coding their values everywhere.

ajeddeloh:
updated

sdk/enter.go
```diff
@@ -209,6 +210,9 @@ func (e *enter) MountGnupgHome() error {
 }
 
 func (e *enter) MountGnupgAgent() error {
+	if !e.BindGpgAgent {
```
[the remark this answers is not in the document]
bgilbert: This is now here _and_ in the caller, and also still public.

sdk/enter.go
```diff
@@ -318,16 +322,19 @@ func (e *enter) CopyGoogleCreds() error {
...
+		return fmt.Errorf("got %d args, need exactly 1", len(args))
 	}
 
 	e := enter{
-		RepoRoot: args[0],
-		Chroot:   args[1],
-		CmdDir:   args[2],
-		Cmd:      args[3:],
+		RepoRoot: RepoRoot(),
+		Chroot:   filepath.Join(RepoRoot(), "chroot"),
```
bgilbert: For the record, this moves these calculations from the parent process to the child.  Seems okay, but is a behavior change.

sdk/repo.go
```diff
@@ -153,5 +156,9 @@ func RepoSync(chroot string, force bool) error {
 	if force {
 		args = append(args, "--force-sync")
 	}
-	return enterChroot(chroot, chrootRepoRoot, args...)
+	return enterChroot(enter{
+		Chroot: filepath.Join(RepoRoot(), chroot),
```
bgilbert: This still has the `Join` in all of the callers.  My point was that `enterChroot` could do it before marshalling.

sdk/enter.go
```diff
@@ -365,14 +381,20 @@ func enterChrootHelper(args []string) (err error) {
...
+	if err := e.MountSSHAgent(); err != nil {
 		return err
 	}
 
-	if err := e.MountGnupg(); err != nil {
+	if err := e.MountGnupgHome(); err != nil {
 		return err
 	}
 
+	if e.BindGpgAgent {
```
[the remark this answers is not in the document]
ajeddeloh: Fixed now

ajeddeloh:
Moved calculations with RepoRoot back to the parent process, fixed bindAgent check being in the wrong commit.

ajeddeloh:
Fixed duplicate bindAgent checking, fixed gofmt

PR #826, opened 2018-02-26
platform/api/do: rety for up to 5 min instead of 1

ajeddeloh:


platform/api/do/api.go
```diff
@@ -133,8 +133,9 @@ func (a *API) PreflightCheck(ctx context.Context) error {
 func (a *API) CreateDroplet(ctx context.Context, name string, sshKeyID int, userdata string) (*godo.Droplet, error) {
 	var droplet *godo.Droplet
 	var err error
-	// DO frequently gives us 422 errors saying "Please try again"
-	err = util.RetryConditional(6, 10*time.Second, shouldRetry, func() error {
+	// DO frequently gives us 422 errors saying "Please try again". Retry every 10 seconds
+	// for up to 5 min
```
arithx: Technically this isn't correct, it will do up to 30 retries with a 10 second delay in between. The `WaitUntilReady` function is the one with a hard limit on the amount of time.

ajeddeloh: I just wanted the general amount of time, make it clear why the those numbers in particular were being multiplied.

arithx: Gotcha, it's fine either way. Just want to make sure that we weren't looking for the hard limit of time.

PR #827, opened 2018-02-28
kola: add support for debugging systemd units

ajeddeloh:
Add support for enabling debug logging on `systemd-*.unittype` units. It
creates drop in units that sets the environment variable
`SYSTEMD_LOG_LEVEL=debug` for the listed units. It applies to both spawn
and run commands.

I'm not 100% this is the best interface for this. We could make it more generic and let users specify whatever dropins they want, but that makes it more verbose. Hell we could even let them supply a whole Ignition config to append. I think this is a good start though and we can add those things if we decide we want them. I'd like to enable this on the nightly with `systemd-networkd.service` and `systemd-networkd-wait-online.service` to hunt down that flake.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/coreos/mantle/auth"
 	"github.com/coreos/mantle/kola"
+	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/sdk"
 )
 
@@ -43,6 +44,7 @@ var (
 func init() {
 	sv := root.PersistentFlags().StringVar
 	bv := root.PersistentFlags().BoolVar
+	ss := root.PersistentFlags().StringSlice
 
 	// general options
 	sv(&outputDir, "output-dir", "", "Temporary output directory for test data and logs")
@@ -51,6 +53,7 @@ func init() {
 	root.PersistentFlags().IntVarP(&kola.TestParallelism, "parallel", "j", 1, "number of tests to run in parallel")
 	sv(&kola.TAPFile, "tapfile", "", "file to write TAP results to")
 	sv(&kola.Options.BaseName, "basename", "kola", "Cluster name prefix")
+	ss("debug-systemd-unit", []string{}, "full-unit-name.service to enable SYSTEMD_LOG_LEVEL=debug on. Specify multiple times for multiple units.")
 
 	// aws-specific options
 	defaultRegion := os.Getenv("AWS_REGION")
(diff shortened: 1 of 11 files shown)


kola/harness.go
```diff
@@ -371,6 +371,7 @@ func runTest(h *harness.H, t *register.Test, pltfrm string) {
 		NoSSHKeyInUserData: t.HasFlag(register.NoSSHKeyInUserData),
 		NoSSHKeyInMetadata: t.HasFlag(register.NoSSHKeyInMetadata),
 		NoEnableSelinux:    t.HasFlag(register.NoEnableSelinux),
+		DebugSystemdUnits:  strings.Split(Options.DebugSystemdUnits, " "),
```
bgilbert: If the intent is to make this a user-provided global setting (as the commit message indicates), this doesn't belong in `RuntimeConfig`, which is a per-cluster config driven by test declarations.  Note that the option is currently kola-wide but only actually applies to subcommands calling through `runTest`, i.e. not `spawn`.  A better approach would be to have `NewBaseCluster` take `Options` rather than `Options.BaseName`, as it should have in the first place, and then let `RenderUserData` read it directly from there.

platform/base.go
```diff
@@ -47,13 +47,14 @@ type BaseCluster struct {
 	rconf      *RuntimeConfig
 	platform   Name
 	ctPlatform string
+	baseopts   *Options
 }
 
-func NewBaseCluster(basename string, rconf *RuntimeConfig, platform Name, ctPlatform string) (*BaseCluster, error) {
-	return NewBaseClusterWithDialer(basename, rconf, platform, ctPlatform, network.NewRetryDialer())
+func NewBaseCluster(opts *Options, rconf *RuntimeConfig, platform Name, ctPlatform string) (*BaseCluster, error) {
```
bgilbert: I'm afraid this time I'm going to insist: this refactor should be in a separate commit.

cmd/kola/options.go
```diff
@@ -51,6 +53,7 @@ func init() {
 	root.PersistentFlags().IntVarP(&kola.TestParallelism, "parallel", "j", 1, "number of tests to run in parallel")
 	sv(&kola.TAPFile, "tapfile", "", "file to write TAP results to")
 	sv(&kola.Options.BaseName, "basename", "kola", "Cluster name prefix")
+	sv(&debugSystemdUnits, "debug-systemd-units", "", "space delimited list of full-unit-names.service to enabled SYSTEMD_LOG_LEVEL=debug on")
```
bgilbert: `to enable`

This should probably also use a string slice variable (e.g. `--debug-systemd-unit systemd-networkd.service --debug-systemd-unit systemd-resolved.service`) rather than using internal delimiters in the argument.

platform/base.go
ajeddeloh: Fixed now

ajeddeloh:
Broke into two commit and used StringSlice instead of a space delimited list.

cmd/kola/options.go
```diff
@@ -51,6 +53,7 @@ func init() {
 	root.PersistentFlags().IntVarP(&kola.TestParallelism, "parallel", "j", 1, "number of tests to run in parallel")
 	sv(&kola.TAPFile, "tapfile", "", "file to write TAP results to")
 	sv(&kola.Options.BaseName, "basename", "kola", "Cluster name prefix")
+	ss("debug-systemd-units", []string{}, "full-unit-names.service to enable SYSTEMD_LOG_LEVEL=debug on. Specify multiple times for multiple units.")
```
bgilbert: Each instance of the option will address a single unit, so the option name shouldn't be plural.

PR #829, opened 2018-03-06
kola/harness: update segfault regex

arithx:
The current regex to check for segfaults looks for `SEGV` which
incorrectly triggered when an SSH key on the machine had SEGV in it.
This updates the regex to `SIGSEGV`.

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -117,7 +117,7 @@ var (
 		},
 		{
 			desc:  "segfault",
-			match: regexp.MustCompile("SEGV"),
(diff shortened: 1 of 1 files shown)


arithx:
@bgilbert updating to `SIGSEGV|11/SEGV`

This seems to cover both cases while also avoiding the SEGV in hash string: https://play.golang.org/p/FOv6HZgCw8j

dm0-:
It won't match that specific host key, but it can still match others.  The fingerprints are base64-encoded, so `/` is still a valid character.  If you also prefix the `=` character, it should be safe in general.

arithx:
Updated with the `=` prefix

PR #830, opened 2018-03-06
Bump timeouts for packet, ec2

ajeddeloh:
Let's have fewer test failures due to clouds being slow.


platform/api/aws/ec2.go
```diff
@@ -108,7 +108,7 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
 	var insts []*ec2.Instance
 
 	// 5 minutes is a pretty reasonable timeframe for AWS instances to work.
```
euank: stale comment as well

ajeddeloh:
Fixed the stale comments

PR #832, opened 2018-03-12
platform/oci: update to oracle/oci-go-sdk

arithx:
The new Oracle OCI SDK (`oci-go-sdk`) has been released which replaces `bmcs-go-sdk`.

This drops the previous vendor'd library (no more non-glide vendor :tada:) which was originally added in 4abdebc, adds the new library, and then updates `platform/api/oci` & `cmd/ore/oci`.


arithx:
Wasn't sure if splitting the commits for each section of the API would make more sense.

platform/api/oci/api.go
```diff
@@ -91,25 +99,40 @@ func New(opts *Options) (*API, error) {
 		opts.CompartmentID = profile.CompartmentID
 	}
 
-	extraOpts := []baremetal.NewClientOptionsFunc{}
-	extraOpts = append(extraOpts, baremetal.PrivateKeyFilePath(opts.KeyFile))
+	privateKey, err := ioutil.ReadFile(opts.KeyFile)
+	if err != nil {
+		return nil, fmt.Errorf("parsing RSA private key: %v", err)
```
ajeddeloh: s/parsing RSA private key/reading RSA private key file

platform/api/oci/api.go
```diff
@@ -120,3 +143,11 @@ func (a *API) GC(gracePeriod time.Duration) error {
 func boolToPtr(b bool) *bool {
 	return &b
 }
+
```
ajeddeloh: do we not have these elsewhere to just import?

platform/api/oci/image.go
```diff
@@ -15,30 +15,40 @@
...
+func (a *API) UploadImage(bucketName, name, filePath string) (objectstorage.PutObjectResponse, error) {
+	namespace, err := a.os.GetNamespace(context.Background(), objectstorage.GetNamespaceRequest{})
 	if err != nil {
-		return nil, err
+		return objectstorage.PutObjectResponse{}, err
 	}
-	if namespace == nil {
-		return nil, fmt.Errorf("namespace was nil")
+	if namespace.Value == nil {
+		return objectstorage.PutObjectResponse{}, fmt.Errorf("namespace was nil")
```
ajeddeloh: s/namespace was nil/recieved nil namespace

platform/api/oci/instance.go
```diff
@@ -30,7 +31,7 @@ func (a *API) CreateInstance(name, userdata, sshKey string) (*Machine, error) {
 		return nil, err
 	}
 
-	subnet, err := a.getSubnetOnVCN(vcn.ID)
+	subnet, err := a.getSubnetOnVCN(*vcn.Id)
```
ajeddeloh: check for nil for vcn.Id, or if nil is OK in this context, drop a comment saying so

platform/api/oci/instance.go
```diff
@@ -45,98 +46,113 @@ func (a *API) CreateInstance(name, userdata, sshKey string) (*Machine, error) {
...
-		if inst.State != "RUNNING" {
+		if inst.LifecycleState != "RUNNING" {
 			return false, nil
 		}
 
 		return true, nil
 	})
 	if err != nil {
-		a.TerminateInstance(id)
+		a.TerminateInstance(*id)
```
ajeddeloh: check for nil

platform/api/oci/instance.go
```diff
@@ -45,98 +46,113 @@ func (a *API) CreateInstance(name, userdata, sshKey string) (*Machine, error) {
...
-		},
-	}
-	vnicAttachments, err := a.client.ListVnicAttachments(a.opts.CompartmentID, &vnicOpts)
+	vnicAttachments, err := a.compute.ListVnicAttachments(context.Background(), core.ListVnicAttachmentsRequest{
+		CompartmentId: &a.opts.CompartmentID,
+		InstanceId:    inst.Id,
+	})
 	if err != nil {
-		a.TerminateInstance(inst.ID)
+		a.TerminateInstance(*inst.Id)
```
ajeddeloh: check for nil

platform/api/oci/instance.go
```diff
@@ -45,98 +46,113 @@ func (a *API) CreateInstance(name, userdata, sshKey string) (*Machine, error) {
...
 		return nil, fmt.Errorf("listing vnic attachments: %v", err)
 	}
 
-	if len(vnicAttachments.Attachments) < 1 {
-		a.TerminateInstance(inst.ID)
+	if len(vnicAttachments.Items) < 1 {
+		a.TerminateInstance(*inst.Id)
```
ajeddeloh: check for nil

platform/api/oci/instance.go
```diff
@@ -45,98 +46,113 @@ func (a *API) CreateInstance(name, userdata, sshKey string) (*Machine, error) {
...
 	if err != nil {
 		return nil, fmt.Errorf("getting vnic: %v", err)
 	}
 
 	return &Machine{
-		Name:             inst.DisplayName,
-		ID:               inst.ID,
-		PublicIPAddress:  vnic.PublicIPAddress,
-		PrivateIPAddress: vnic.PrivateIPAddress,
+		Name:             *inst.DisplayName,
```
ajeddeloh: check these for nil.

platform/api/oci/instance.go
```diff
@@ -147,25 +163,28 @@ func (a *API) GetConsoleOutput(instanceID string) (string, error) {
...
+		InstanceConsoleHistoryId: consoleHistoryStatus.ConsoleHistory.Id,
+		Length: intToPtr(2 * 1024 * 1024),
+		Offset: intToPtr(0),
 	})
 	if err != nil {
 		return "", fmt.Errorf("getting console history data: %v", err)
 	}
 
-	return content.Data, nil
+	return *content.Value, nil
```
ajeddeloh: check for nil

platform/api/oci/instance.go
```diff
@@ -174,13 +193,13 @@ func (a *API) gcInstances(gracePeriod time.Duration) error {
...
 
-		switch instance.State {
+		switch instance.LifecycleState {
 		case "TERMINATING", "TERMINATED":
 			continue
 		}
 
-		if err := a.TerminateInstance(instance.ID); err != nil {
-			return fmt.Errorf("couldn't terminate instance %v: %v", instance.ID, err)
+		if err := a.TerminateInstance(*instance.Id); err != nil {
```
ajeddeloh: check for nil

platform/api/oci/networking.go
```diff
@@ -15,194 +15,253 @@
...
 	if err != nil {
-		return nil, err
+		return core.Vcn{}, err
 	}
 
-	for _, v := range vcns.VirtualNetworks {
-		if name == v.DisplayName {
-			return &v, nil
+	for _, v := range vcns.Items {
+		if v.DisplayName != nil && name == *v.DisplayName {
```
ajeddeloh: check for nil

platform/api/oci/networking.go
```diff
@@ -15,194 +15,253 @@
...
-			Protocol: "17",
-			Source:   "10.0.0.0/16",
-			UDPOptions: &baremetal.UDPOptions{
-				DestinationPortRange: baremetal.PortRange{
-					Min: 1,
-					Max: 65535,
+			IngressSecurityRules: []core.IngressSecurityRule{
+				{
+					// Allow all TCP on private network
+					Protocol: strToPtr("6"),
```
ajeddeloh: make all the magic protocol numbers consts

cmd/ore/oci/create.go
```diff
@@ -46,25 +46,26 @@ func runCreateVCN(cmd *cobra.Command, args []string) {
...
-	vcn, err := API.CreateVCN("kola", "10.0.0.0/16")
+	vcnResp, err := API.CreateVCN("kola", "10.0.0.0/16")
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Creating VCN: %v\n", err)
 		os.Exit(1)
 	}
+	vcn := vcnResp.Vcn
 
-	secList, err := API.CreateDefaultSecurityList(vcn.ID)
+	secList, err := API.CreateDefaultSecurityList(*vcn.Id)
```
ajeddeloh: check for nil throughout this commit

platform/api/oci/api.go
arithx: We do not, I can certainly add a commit making these more general though.

ajeddeloh: If we don't, I'd say leave it as is for now. We can refactor laterif we want; keep this PR on topic.

platform/api/oci/instance.go
arithx: Whoops, forgot to add these. Forgot that mandatory fields in JSON do not mean that the fields will not be nil during unmarshaling.

arithx:
Updated with missing nil checks & moved magic strings to `const`

Running tests again locally once the jenkins jobs finish.

cmd/ore/oci/create.go
```diff
@@ -46,30 +46,51 @@ func runCreateVCN(cmd *cobra.Command, args []string) {
...
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Creating VCN: %v\n", err)
 		os.Exit(1)
 	}
+	vcn := vcnResp.Vcn
 
-	secList, err := API.CreateDefaultSecurityList(vcn.ID)
+	if vcn.Id == nil {
+		fmt.Fprintf(os.Stderr, "received virtual cloud network id nil\n")
+		os.Exit(1)
```
ajeddeloh: I'm not generally a fan of `os.Exit()`ing all over the place. Maybe wrap this in another function call that returns an error and handle all the os.Exit() in the caller?

Either that or maybe have a `exitIfNil(variable interface{}, msg string)` to help with the verbosity

arithx:
Updated per OOB discussion to reduce the amount of `os.Exit` calls.

Looking into the `delete-kola-vcn` functionality (it seems that in the change between `bmcs-go-sdk` & `oci-go-sdk` something about the delete / ordering changed and this has been flaky. Currently I'm having to switch the ordering around back and forth to get successful deletes.

arithx:
Updated the `cmd/ore/oci` commit to disable the `delete-kola-vcn` command. The code remains present for future investigation / re-instatement when the API becomes more stable.

platform/api/oci/instance.go
```diff
@@ -45,98 +50,133 @@ func (a *API) CreateInstance(name, userdata, sshKey string) (*Machine, error) {
...
-		a.TerminateInstance(id)
+		a.TerminateInstance(*id)
 		return nil, fmt.Errorf("waiting for machine to become active: %v", err)
 	}
 
-	vnicOpts := baremetal.ListVnicAttachmentsOptions{
-		InstanceIDListOptions: baremetal.InstanceIDListOptions{
-			InstanceID: inst.ID,
-		},
+	if inst.Id == nil {
```
ajeddeloh: move this check about the TerminateInstance command.

platform/api/oci/instance.go
```diff
@@ -45,98 +50,133 @@ func (a *API) CreateInstance(name, userdata, sshKey string) (*Machine, error) {
...
-		consoleHistoryStatus, err = a.client.GetConsoleHistory(metadata.ID)
+		consoleHistoryStatus, err := a.compute.GetConsoleHistory(context.Background(), core.GetConsoleHistoryRequest{
+			InstanceConsoleHistoryId: metadata.ConsoleHistory.Id,
+		})
 		if err != nil {
 			return false, fmt.Errorf("getting console history status: %v", err)
 		}
 
-		if consoleHistoryStatus.State != "SUCCEEDED" {
+		if consoleHistoryStatus.ConsoleHistory.LifecycleState != "SUCCEEDED" {
```
ajeddeloh: Mabye move these strings out to consts as well, to protect against typos. I know I personally can't spell succeed to save my life.

platform/api/oci/networking.go
```diff
@@ -15,194 +15,259 @@
...
 
-	"github.com/oracle/bmcs-go-sdk"
+	"github.com/oracle/oci-go-sdk/core"
+	"github.com/oracle/oci-go-sdk/identity"
 )
 
-func (a *API) GetVCN(name string) (*baremetal.VirtualNetwork, error) {
-	vcns, err := a.client.ListVirtualNetworks(a.opts.CompartmentID, nil)
+const (
+	TCP  = "6"
```
ajeddeloh: Do these need to be public? (Normally I'd agree they should be ALL_CAPS, but golang has other plans...(

arithx: They dont, seems old habits die hard. I'll switch to private.

platform/api/oci/instance.go
arithx: The earlier `TerminateInstance` calls uses the `id` variable which is checked before hand. Inside of the `WaitUntilReady` we are updating the `inst` variable with the output of `GetInstance`, this check is to make sure that the new output also doesn't have a nil `Id`.

Are you wanting me to move this check to after each update of the `inst` variable and swap the `TerminateInstance` call to use `inst.Id`?

ajeddeloh: Oh, woops, got `id` and `inst.Id` mixed up. This is good.

arithx:
updated

PR #835, opened 2018-03-28
coreos.locksmith.cluster: switch to using util.Timeout

arithx:
Adds a new utility function `Timeout` and uses it in `coreos.locksmith.cluster` to prevent the test from spinning for extended durations.

### DIFF
--- a/kola/tests/locksmith/locksmith.go
+++ b/kola/tests/locksmith/locksmith.go
@@ -206,6 +206,9 @@ func locksmithReboot(c cluster.TestCluster) {
 	// The machine should be able to reboot without etcd in the default mode
 	m := c.Machines()[0]
 
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	defer cancel()
+
 	output, err := c.SSH(m, "sudo systemctl stop sshd.socket && locksmithctl send-need-reboot")
 	if _, ok := err.(*ssh.ExitMissingError); ok {
 		err = nil // A terminated session is perfectly normal during reboot.
@@ -216,7 +219,7 @@ func locksmithReboot(c cluster.TestCluster) {
 		c.Fatalf("failed to run \"locksmithctl send-need-reboot\": output: %q status: %q", output, err)
 	}
 
-	err = platform.CheckMachine(m)
+	err = platform.CheckMachine(ctx, m)
 	if err != nil {
 		c.Fatalf("failed to check rebooted machine: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


arithx:
Investigating ways to get the goroutine to stop if either of you have suggestions.

kola/tests/locksmith/locksmith.go
```diff
@@ -233,32 +233,40 @@ func locksmithCluster(c cluster.TestCluster) {
 
 	c.MustSSH(machs[0], "locksmithctl status")
 
-	ctx := context.Background()
-	wg := worker.NewWorkerGroup(ctx, len(machs))
```
euank: Can the diff just be 

```
ctx, cleanup := context.WithTimeout(context.TODO(), 10 * time.Minute)
defer cleanup()
```

and not make any other changes?

I don't think the util timeout function is useful if we already are using contexts since they have `WithTimeout` already

arithx: Just using `context.WithTimeout` doesn't cause the test to fail when a timeout occurs. I'll look into ways of making it properly bubble out the failure.

euank: I think the failure to bubble up the failure is still in the locksmith code after reading more.

Basically, I think it's expected the function passed to `wg.Start` should bubble up context errors if they're relevant.

In this case, they are, so I think the right fix is to have the worker func defined a few lines below return `ctx.Err()` after the machine check.

It does seem a little awkward that the timeout will fail the waitgroup if it happens before all work is scheduled, but requires the work to bubble it up if it happens later, but I think it's overall reasonable and in-line with how context is meant to be used.

euank: Actually, after looking a bit more, it seems like `platform.CheckMachine` should probably take a `ctx` as an argument and respect its timeout since that's where the time's really being spent: retrying there.

arithx: Pushed a change that adds a `context.Context` variable to `platform.CheckMachine`.

platform/platform.go
```diff
@@ -260,9 +261,12 @@ func NewMachines(c Cluster, userdata *conf.UserData, n int) ([]Machine, error) {
 // It also ensures the remote system is running Container Linux by CoreOS.
 //
 // TODO(mischief): better error messages.
-func CheckMachine(m Machine) error {
+func CheckMachine(ctx context.Context, m Machine) error {
 	// ensure ssh works and the system is ready
 	sshChecker := func() error {
+		if err := ctx.Err(); err != nil {
```
euank: This still feels slightly hacky, but I think it's okay.

Ideally, I think `m.SSH` would also take a ctx, and `util.Retry` would understand if the error returned was a context timeout it's not retryable... but plumbing that through seems like it's overkill for the immediate problem being solved here.

PR #838, opened 2018-03-29
cmd/ore/do: add list-keys and delete-keys commands

bgilbert:
Copypasted from `packet`.


cmd/ore/do/list-keys.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2017 CoreOS, Inc.
```
arithx: 2018

cmd/ore/do/delete-keys.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2017 CoreOS, Inc.
```
arithx: 2018

PR #840, opened 2018-04-06
kola/tests: add updatepayload test

arithx:
Creates a new test `coreos.update.updatepayload` that mirrors the
functionality of `kola updatepayload` when an update payload is provided
via the `--update-payload` flag.

### DIFF
--- a/kola/tests/update/update.go
+++ b/kola/tests/update/update.go
@@ -1,0 +1,183 @@
+// Copyright 2018 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package update
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/coreos/go-omaha/omaha"
+
+	"github.com/coreos/mantle/kola"
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	tutil "github.com/coreos/mantle/kola/tests/util"
+	"github.com/coreos/mantle/platform"
+	"github.com/coreos/mantle/platform/local"
+	"github.com/coreos/mantle/util"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Name:        "coreos.update.payload",
+		Run:         payload,
+		ClusterSize: 1,
+		NativeFuncs: map[string]func() error{
+			"Omaha": Serve,
+		},
+	})
+}
+
+func Serve() error {
+	omahaserver, err := omaha.NewTrivialServer(":34567")
+	if err != nil {
+		return fmt.Errorf("creating trivial omaha server: %v\n", err)
+	}
+
+	omahawrapper := local.OmahaWrapper{TrivialServer: omahaserver}
+
+	if err = omahawrapper.AddPackage("/updates/update.gz", "update.gz"); err != nil {
+		return fmt.Errorf("bad payload: %v", err)
+	}
+
+	return omahawrapper.Serve()
+}
+
+func payload(c cluster.TestCluster) {
+	addr := configureOmahaServer(c, c.Machines()[0])
+
+	// create the actual test machine, the machine
+	// that is created by the test registration is
+	// used to host the omaha server
+	m, err := c.NewMachine(nil)
+	if err != nil {
+		c.Fatalf("creating test machine: %v", err)
+	}
+
+	// Machines are intentionally configured post-boot
+	// via SSH to allow for testing versions which predate
+	// Ignition
+	configureMachineForUpdate(c, m, addr)
+
+	tutil.AssertBootedUsr(c, m, "USR-A")
+
+	updateMachine(c, m)
+
+	tutil.AssertBootedUsr(c, m, "USR-B")
+
+	tutil.InvalidateUsrPartition(c, m, "USR-A")
+
+	updateMachine(c, m)
+
+	tutil.AssertBootedUsr(c, m, "USR-A")
+}
+
+func configureOmahaServer(c cluster.TestCluster, srv platform.Machine) string {
+	if kola.UpdatePayloadFile == "" {
+		c.Skip("no update payload provided")
+	}
+
+	in, err := os.Open(kola.UpdatePayloadFile)
+	if err != nil {
+		c.Fatalf("opening update payload: %v", err)
+	}
+	defer in.Close()
+	if err := platform.InstallFile(in, srv, "/updates/update.gz"); err != nil {
+		c.Fatalf("copying update payload to omaha server: %v", err)
+	}
+
+	c.MustSSH(srv, fmt.Sprintf("sudo systemd-run --quiet ./kolet run %s Omaha", c.Name()))
+
+	err = util.WaitUntilReady(60*time.Second, 5*time.Second, func() (bool, error) {
+		_, _, err := srv.SSH(fmt.Sprintf("curl %s:34567", srv.PrivateIP()))
+		return err == nil, nil
+	})
+	if err != nil {
+		c.Fatal("timed out waiting for omaha server to become active")
+	}
+
+	return fmt.Sprintf("%s:34567", srv.PrivateIP())
+}
+
+func configureMachineForUpdate(c cluster.TestCluster, m platform.Machine, addr string) {
+	// update atomicly so nothing reading update.conf fails
+	c.MustSSH(m, fmt.Sprintf(`sudo bash -c "cat >/etc/coreos/update.conf.new <<EOF
+GROUP=developer
+SERVER=http://%s/v1/update
+EOF"`, addr))
+	c.MustSSH(m, "sudo mv /etc/coreos/update.conf{.new,}")
+
+	// inject dev key
+	c.MustSSH(m, `sudo bash -c "cat >/etc/coreos/update-payload-key.pub.pem <<EOF
+-----BEGIN PUBLIC KEY-----
+MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAzFS5uVJ+pgibcFLD3kbY
+k02Edj0HXq31ZT/Bva1sLp3Ysv+QTv/ezjf0gGFfASdgpz6G+zTipS9AIrQr0yFR
++tdp1ZsHLGxVwvUoXFftdapqlyj8uQcWjjbN7qJsZu0Ett/qo93hQ5nHW7Sv5dRm
+/ZsDFqk2Uvyaoef4bF9r03wYpZq7K3oALZ2smETv+A5600mj1Xg5M52QFU67UHls
+EFkZphrGjiqiCdp9AAbAvE7a5rFcJf86YR73QX08K8BX7OMzkn3DsqdnWvLB3l3W
+6kvIuP+75SrMNeYAcU8PI1+bzLcAG3VN3jA78zeKALgynUNH50mxuiiU3DO4DZ+p
+5QIDAQAB
+-----END PUBLIC KEY-----
+EOF"`)
+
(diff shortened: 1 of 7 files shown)


kola/tests/update/update.go
```diff
@@ -0,0 +1,250 @@
...
+func updatepayload(c cluster.TestCluster) {
+	if kola.UpdatePayload == "" {
+		c.Skip("no update payload provided")
+	}
+
+	// create the actual test machine, the machine
+	// that is created by the test registration is
+	// used to host the omaha server if running on
+	// a non-qemu platform
+	m, err := c.NewMachine(conf.Ignition(`{"ignition":{"version": "2.1.0"}}`))
```
bgilbert: `c.NewMachine(nil)` is sufficient (though that will currently pass a 2.0.0 config, if that matters).

kola/tests/update/update.go
```diff
@@ -0,0 +1,250 @@
...
+		c.Fatalf("invalidating USR-A failed: %s: %v: %s", out, err, stderr)
+	}
+
+	updateMachine(m, c)
+
+	checkUsrA(m, c)
+}
+
+func configureOmahaServer(c cluster.TestCluster) string {
+	if qc, ok := c.Cluster.(*qemu.Cluster); ok {
```
bgilbert: Is there value to either
1. simplifying the code by always using the dedicated omaha machine, or
2. not launching the omaha machine on qemu (which would require manually copying kolet)?

kola/tests/update/update.go
```diff
@@ -0,0 +1,250 @@
...
+			c.Fatalf("copying update payload to omaha server: %v", err)
+		}
+
+		c.MustSSH(srv, fmt.Sprintf("sudo systemd-run --quiet ./kolet run %s Omaha", c.Name()))
+
+		return fmt.Sprintf("%s:34567", srv.PrivateIP())
+	}
+}
+
+func configureMachine(m platform.Machine, c cluster.TestCluster, addr string) {
```
bgilbert: Conventionally `c` before `m`.

kola/tests/update/update.go
```diff
@@ -0,0 +1,250 @@
...
+6kvIuP+75SrMNeYAcU8PI1+bzLcAG3VN3jA78zeKALgynUNH50mxuiiU3DO4DZ+p
+5QIDAQAB
+-----END PUBLIC KEY-----
+EOF"`)
+
+	c.MustSSH(m, "sudo mount --bind /etc/coreos/update-payload-key.pub.pem /usr/share/update_engine/update-payload-key.pub.pem")
+
+	// disable reboot so the test has explicit control
+	c.MustSSH(m, "sudo systemctl mask locksmithd.service")
+	c.MustSSH(m, "sudo systemctl stop locksmithd.service")
```
bgilbert: The above can be combined into `mask --now`.

kola/tests/update/update.go
```diff
@@ -0,0 +1,250 @@
...
+	c.MustSSH(m, "sudo systemctl reset-failed locksmithd.service")
+
+	c.MustSSH(m, "sudo systemctl restart update-engine.service")
+
+	c.MustSSH(m, "update_engine --v=10")
```
bgilbert: Left over from debugging?

kola/tests/update/update.go
```diff
@@ -0,0 +1,250 @@
...
+	for _, p := range pairs {
+		spl := strings.SplitN(p, "=", 2)
+		if len(spl) == 2 {
+			m[spl[0]] = spl[1]
+		}
+	}
+	return m
+}
+
+func checkUsrA(m platform.Machine, c cluster.TestCluster) {
```
bgilbert: `misc` already has `assertBootedUsr()`.  Probably better to export that (maybe in a `util` package?) and reuse.

arithx:
My initial vision was to produce a payload during manifest builds in jenkins and run at least the currently released version -> version being built. I do think that identifying specific versions for each channel would have some value as well.

kola/tests/update/update.go
```diff
@@ -0,0 +1,250 @@
...
+			fmt.Printf("creating trivial omaha server: %v\n", err)
+		}
+
+		omahawrapper := local.OmahaWrapper{TrivialServer: omahaserver}
+
+		if err = omahawrapper.AddPackage("/updates/update.gz", "update.gz"); err != nil {
+			fmt.Printf("bad payload: %v", err)
+		}
+
+		go omahawrapper.Serve()
```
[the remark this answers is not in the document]
euank: Agreed; I'm not sure either of the goroutines are necessary.

kola/tests/update/update.go
arithx: I'll play around with using a dedicated omaha machine for the QEMU cluster rather than the one provided via the `LocalCluster`.

kola/tests/update/update.go
arithx: Thoughts on location for the containing `util` package? As this requires both `cluster.TestCluster` & `platform.Machine` it cannot live in `util/` or in `platform/util.go` because of circular imports. I'm not particularly partial to creating a `util` package inside of `kola/tests` either. Maybe inside of `kola/cluster`?

euank:
I think it would be preferable if the test could dictate the version it's updating from.

e.g., it would be useful if we could have a test for `coreos.update.free-magic-grub` which updates a specific version of CL that has grub free-magic.

Right now, that test can't be written since the test would have to define a base CL version with a bad grub. As implemented, that base version is implied by the environment, so to run that test would be a matter of downloading that old version and running a regular update test.
Correct me if I'm misunderstanding how this would work.

I recognize that pragmatically it would be difficult to teach kola how to boot old versions on each platform and this implementation might be the best reasonable compromise, I just wanted to make sure that if we're choosing not to do the above after a conscious decision.

arithx:
@euank you're understanding of the current implementation is correct; my thinking for now was that we would define an additional job inside of the jenkins-os that would take care of running `coreos.update.*` tests with a set of known verisons more manually e.g.:
```
for id in <ami_id1> <ami_id2>; do
  kola run -p aws --aws-ami <ami_id> coreos.update.*
done
```

I do agree that it would be nice to be able to pin versions to certain tests declaratively but the implementation on the runner side might not be something we'd want to add directly to `kola`. Maybe a seperate binary in mantle that's goal is directly targeting update testing from a set of known versions defined by the tests?

kola/tests/update/update.go
bgilbert: I do think it makes sense to have a `kola.tests.util`, containing purely test code shared among multiple tests.  I'm open to other ideas though.

arithx:
Update pushed

kola/tests/update/update.go
```diff
@@ -0,0 +1,180 @@
...
+func ConfigureOmahaServer(c cluster.TestCluster, srv platform.Machine) string {
+	if kola.UpdatePayloadFile == "" {
+		c.Skip("no update payload provided")
+	}
+
+	if qc, ok := c.Cluster.(*qemu.Cluster); ok {
+		// An Omaha Server is built for all qemu clusters via the LocalCluster;
+		// Destroy it before setting up the omaha server on the machine created
+		// for the test.
+		qc.OmahaServer.Destroy()
```
bgilbert: Why is this necessary?  Also, is `OmahaServer` safe against multiple destructions?

arithx: I'll test without destroying this, I left the disabling code because previously it was overriding the `OmahaServer` I was setting up but that shouldn't be an issue anymore.

arithx: You were correct that this is no longer necessary. Now that the `OmahaServer` is being hosted on the original machine created by `kola` it is on a different IP than the host (`10.0.0.2` for the `kola` machine vs `10.0.0.1` for the host).

Pushed an update dropping this.

kola/tests/update/update.go
```diff
@@ -0,0 +1,172 @@
...
+	in, err := os.Open(kola.UpdatePayloadFile)
+	if err != nil {
+		c.Fatalf("opening update payload: %v", err)
+	}
+	defer in.Close()
+	if err := platform.InstallFile(in, srv, "/updates/update.gz"); err != nil {
+		c.Fatalf("copying update payload to omaha server: %v", err)
+	}
+
+	c.MustSSH(srv, fmt.Sprintf("sudo systemd-run --quiet ./kolet run %s Omaha", c.Name()))
```
euank: There should also be a verification that the server is running before the address is returned, otherwise this is racing with the launching of the new machine.

kola/tests/update/update.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// create the actual test machine, the machine
+	// that is created by the test registration is
+	// used to host the omaha server
+	m, err := c.NewMachine(nil)
+	if err != nil {
+		c.Fatalf("creating test machine: %v", err)
+	}
+
+	ConfigureMachineForUpdate(c, m, addr)
```
euank: Is there a reason the machine isn't being configured with Ignition?

That'll make it easier to not race with update-engine for writing files (since the systemd unit that bindmounts over the pubkey can have a `Before=update-engine.service`).

I also think a CLC is more readable than the procedural ssh bits in this function.

kola/tests/update/update.go
```diff
@@ -0,0 +1,172 @@
...
+	c.MustSSH(m, "sudo mount --bind /etc/coreos/update-payload-key.pub.pem /usr/share/update_engine/update-payload-key.pub.pem")
+
+	// disable reboot so the test has explicit control
+	c.MustSSH(m, "sudo systemctl mask --now locksmithd.service")
+	c.MustSSH(m, "sudo systemctl reset-failed locksmithd.service")
+
+	c.MustSSH(m, "sudo systemctl restart update-engine.service")
+}
+
+func UpdateMachine(c cluster.TestCluster, m platform.Machine) {
```
euank: doesn't need to be an exported function.

kola/tests/update/update.go
```diff
@@ -0,0 +1,172 @@
...
+	err = util.WaitUntilReady(120*time.Second, 10*time.Second, func() (bool, error) {
+		envs, stderr, err := m.SSH("update_engine_client -status 2>/dev/null")
+		if err != nil {
+			return false, fmt.Errorf("checking status failed: %v: %s", err, stderr)
+		}
+
+		return splitNewlineEnv(string(envs))["CURRENT_OP"] == "UPDATE_STATUS_UPDATED_NEED_REBOOT", nil
+	})
+	if err != nil {
+		c.Fatalf("Updating machine: %v", err)
```
euank: A clearer message might be nice, e.g. `error waiting for STATUS_NEED_REBOOT: %v`

kola/tests/update/update.go
bgilbert: In practice, is it likely for the kolet launch to lose the race with the instance launch?

euank: No. In reality we can probably leave this be, but I also think it would be nice to verify it's serving for the chance that kolet fails to launch correctly. `systemd-run` can quite easily eat errors and just spit out "yeah, I ran a unit" even if the unit instantly failed.

arithx: `update_engine` retries even in cases of no connectivity to the omaha server so this would only really validate that the omaha server eventually starts correctly. Is the thought to detect that failure case and fail the test earlier rather than failing when the machine fails to update and deciphering the logs?

kola/tests/update/update.go
arithx: My initial thought for this was to leave the UserData completely up to the user so different tests could configure machines however they wanted and then these functions would do post-configuration.

Now that we're shipping an updated Ignition vendor I'll look into adding some new methods into `platform/conf` for combining configs and provide a function in this module that takes a custom userdata and adds in the requisite pieces for the update configuration.

kola/tests/update/update.go
arithx: I saw this as more of a library function that other tests could utilize that was specific to updates. I can make this non-exported if we want to keep all update tests to this module.

kola/tests/update/update.go
euank: After thinking about it again, it actually makes sense to not use Ignition since we may want to test updates on versions which predate ignition.

I think a comment to that effect on the function would be worth having.

kola/tests/update/update.go
euank: That's a benefit, yeah. It's mostly about trying to have the function have a clear contract that it abides by.
Returning the address implies that it's already running unless there's documentation to the contrary.

kola/tests/update/update.go
arithx: If we're going to make this function not exported should we also not export the `ConfigureOmahaServer` function?

euank: If it's a library function I wouldn't want it to live in the module. If it ends up being useful for another test, I'd rather it be moved and exported than have that other test import from another test module.

Each test module should be a leaf in the import graph; tests importing directly from other tests will eventually lead to cycles.

arithx:
Update pushed

PR #841, opened 2018-04-07
Update Ignition to 0.24.0

bgilbert:


bgilbert:
Experimental configs would be easy to add if we needed them, at the cost of slightly more maintenance overhead:

1. Experimental configs are not API-stable.
2. When the experimental version became stable, we'd need to be careful of test min/end versions.

I'm okay adding if the need arises.

Deferring merge until after the alpha.

PR #843, opened 2018-04-25
kola/tests/ignition: add a security tls test

arithx:
Adds a test for custom CA certs via Ignition.

### DIFF
--- a/kola/tests/ignition/security.go
+++ b/kola/tests/ignition/security.go
@@ -1,0 +1,149 @@
+// Copyright 2018 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ignition
+
+import (
+	"crypto/tls"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+
+	"github.com/vincent-petithory/dataurl"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform/conf"
+	"github.com/coreos/mantle/platform/machine/packet"
+)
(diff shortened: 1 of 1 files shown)


kola/tests/ignition/security.go
```diff
@@ -0,0 +1,148 @@
...
+
+	c.MustSSH(server, "openssl ecparam -genkey -name secp384r1 -out server.key")
+	c.MustSSH(server, strings.Replace(`openssl req -new -x509 -sha256 -key server.key -out server.crt -days 3650 -subj "/CN=$IP" -config <(cat <<-EOF
+[req]
+default_bits = 2048
+default_md = sha256
+distinguished_name = dn
+
+[ dn ]
+CN = 192.168.0.20
```
dm0-: Is hard-coding this intentional?

kola/tests/ignition/security.go
```diff
@@ -0,0 +1,148 @@
...
+		c.Fatalf("starting client: %v", err)
+	}
+
+	checkResources(c, client, map[string]string{
+		"data": "kola-data",
+	})
+}
+
+func TLSServe() error {
+	publicKey, err := ioutil.ReadFile("/home/core/server.crt")
```
dm0-: Are we still concerned about running kola tests on RHEL-based images?  If so, are we still assuming this will use the `core` user?  Otherwise, maybe these generated files should be written to an absolute path.

kola/tests/ignition/security.go
```diff
@@ -0,0 +1,148 @@
...
+default_md = sha256
+distinguished_name = dn
+
+[ dn ]
+CN = 192.168.0.20
+
+[ SAN ]
+subjectAltName = IP:$IP
+EOF
+) -extensions SAN`, "$IP", ip, -1))
```
dm0-: The current method looks okay, but if you'd like to do something like write the OpenSSL config file with Ignition, you could try using `${ENV::IP}` in the file to substitute the value with an environment variable.

kola/tests/ignition/security.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/coreos/mantle/platform/machine/packet"
+)
+
+var (
+	localSecurityClient = conf.Ignition(`{
+         "ignition": {
+             "version": "2.2.0",
+             "config": {
+               "append": [{
+				   "source": "https://$IP"
```
dm0-: This seems to be switching back and forth between tabs and spaces.

kola/tests/ignition/security.go
arithx: Should be `$IP`, mis-copy from what I was testing locally

kola/tests/ignition/security.go
arithx: Agreed, switching to an absolute path.

PR #845, opened 2018-04-30
kola/spawn: add omaha-package support

lucab:
This introduces a new `--omaha-package` flag to `kola spawn`, in order
to add an update payload to the omaha server and point spawned node to
it.

### DIFF
--- a/cmd/kola/spawn.go
+++ b/cmd/kola/spawn.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -32,6 +33,8 @@ import (
 	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/platform/conf"
 	"github.com/coreos/mantle/platform/machine/qemu"
+	"github.com/coreos/mantle/sdk"
+	"github.com/coreos/mantle/sdk/omaha"
 )
 
 var (
@@ -45,6 +48,7 @@ var (
 	spawnNodeCount      int
 	spawnUserData       string
 	spawnDetach         bool
+	spawnOmahaPackage   string
 	spawnShell          bool
 	spawnRemove         bool
 	spawnVerbose        bool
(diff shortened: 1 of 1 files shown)


lucab:
I'm using this for manually fiddling with postinst, I guess it may be useful for others too.

/cc @arithx for review

cmd/kola/spawn.go
```diff
@@ -125,10 +130,31 @@ func doSpawn(cmd *cobra.Command, args []string) error {
 		defer cluster.Destroy()
 	}
 
+	var updateConf *strings.Reader
+	if spawnOmahaPackage != "" {
+		qc, ok := cluster.(*qemu.Cluster)
+		if !ok {
+			//TODO(lucab): expand platform support
+			return errors.New("--omaha-package is currently only supported on qemu")
```
arithx: Why use `errors.New` vs `fmt.Errorf` like the `spawnNodeCount` check?

cmd/kola/spawn.go
```diff
@@ -57,6 +61,7 @@ func init() {
 	cmdSpawn.Flags().IntVarP(&spawnNodeCount, "nodecount", "c", 1, "number of nodes to spawn")
 	cmdSpawn.Flags().StringVarP(&spawnUserData, "userdata", "u", "", "file containing userdata to pass to the instances")
 	cmdSpawn.Flags().BoolVarP(&spawnDetach, "detach", "t", false, "-kv --shell=false --remove=false")
+	cmdSpawn.Flags().StringVar(&spawnOmahaPackage, "omaha-package", "", "add an update payload to the Omaha server, referenced by image version (e.g. 'latest')")
```
arithx: Would it make sense to mirror the `updatepayload` functionality where you can either specify a direct link to the payload you want to test and only generate a payload from the `latest` image version if one is not provided?

arithx:
@bgilbert: would this require `--detach=true`?

arithx:
Also, do we want to disable & mask `locksmithd` to allow the user to control the reboots explicitly?

cmd/kola/spawn.go
lucab: Because there is no string formatting involved, error message is constant.

cmd/kola/spawn.go
lucab: The default empty string here has a different semantics, it is used to signal that the user does not want to opt-in the whole omaha-package machinery. If the user specifies a version, then this will either re-use an existing payload or try to generate it first (if it doesn't exist).

lucab:
I would try not to overload too much smartness into this. Both disabling locksmith and detaching makes sense in most cases, but not always (e.g. I'm currently using this with an attached shell).

lucab:
@arithx @bgilbert are you ok with this (which is just a minimal helper on top of existing infra) or would you definitely prefer something broader? Otherwise, I'll stick to the green-light above and merge this.

PR #846, opened 2018-04-30
vendor: update ct to 0.8.0

bgilbert:


platform/conf/conf.go
```diff
@@ -232,7 +232,7 @@ func (u *UserData) Render(ctPlatform string) (*Conf, error) {
 			plog.Warningf("rendering Container Linux config: %s", report)
 		}
 
-		c.ignitionV21 = &ignc
```
ajeddeloh: I'd break this into a seperate commit, otherwise LGTM

bgilbert: Yeah, that'd be my preference, but I didn't want to break the build for one commit.

ajeddeloh: At least call it out in the commit message?

bgilbert: Done.

PR #850, opened 2018-05-07
sdk/enter: make bind mount recursive.

ajeddeloh:
This is a break is compatibility between the legacy and new sdk enter
methods. Make them both recursive bind mounts.

Wait until after 1772.0.0 is released before merging.


bgilbert:
We should tag 0.9.1 after this merges.

PR #858, opened 2018-05-29
DNM: platform/api/aws/ec2: use VPC launches

arithx:
Currently we're doing non-VPC launches which requires a legacy account.
This change supports running on newer accounts by allowing the
specification of security group IDs directly and fetching the subnet ID
to be specified directly.


arithx:
Marking as DNM while investigating a further refactoring of the usage of VPC/SG/Subnets in the AWS platform.

arithx:
Superceded by #862

PR #860, opened 2018-05-30
platform/machine/do: return journal output in ConsoleOutput

arithx:
DigitalOcean doesn't have a way in the SDK to get the console output so
we're currently returning an empty string. This avoids all console
checks some of which are present in the journal. Changes the function to
return the journal output to enable basic checking.

### DIFF
--- a/platform/journal.go
+++ b/platform/journal.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
@@ -30,10 +31,11 @@ import (
 
 // Journal manages recording the journal of a Machine.
 type Journal struct {
-	journal    io.WriteCloser
-	journalRaw io.WriteCloser
-	recorder   *journal.Recorder
-	cancel     context.CancelFunc
+	journal     io.WriteCloser
+	journalRaw  io.WriteCloser
+	journalPath string
+	recorder    *journal.Recorder
+	cancel      context.CancelFunc
 }
 
 // wrapper that also closes the underlying file
@@ -62,8 +64,8 @@ func NewJournal(dir string) (*Journal, error) {
 		return nil, err
 	}
 
-	p = filepath.Join(dir, "journal-raw.txt.gz")
-	jr, err := os.OpenFile(p, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0666)
+	pr := filepath.Join(dir, "journal-raw.txt.gz")
+	jr, err := os.OpenFile(pr, os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0666)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


platform/journal.go
```diff
@@ -78,9 +82,10 @@ func NewJournal(dir string) (*Journal, error) {
 	}
 
 	return &Journal{
-		journal:    j,
-		journalRaw: jrzc,
-		recorder:   journal.NewRecorder(journal.ShortWriter(j), jrzc),
+		journal:     j,
+		journalRaw:  jrzc,
+		recorder:    journal.NewRecorder(journal.ShortWriter(j), jrzc),
+		journalPath: filepath.Join(dir, "journal.txt"),
```
bgilbert: Probably better not to reuse `p` above so you can DRY here.

platform/journal.go
```diff
@@ -34,6 +35,9 @@ type Journal struct {
 	journalRaw io.WriteCloser
 	recorder   *journal.Recorder
 	cancel     context.CancelFunc
+
+	// store the path to the journal so we can read it for DO
```
bgilbert: `for ReadJournal`, really.  This module need not know about DO.

platform/journal.go
```diff
@@ -112,6 +117,14 @@ func (j *Journal) Start(ctx context.Context, m Machine) error {
 	return nil
 }
 
+func (j *Journal) ReadJournal() ([]byte, error) {
+	f, err := os.Open(j.journalPath)
```
bgilbert: Leaked.

platform/machine/do/machine.go
```diff
@@ -77,6 +77,11 @@ func (dm *machine) Destroy() {
 }
 
 func (dm *machine) ConsoleOutput() string {
-	// DigitalOcean provides no API for this
-	return ""
+	// DigitalOcean provides no API for retrieving ConsoleOutput
+	// return the journal instead to allow for error checks to be run.
+	data, err := dm.journal.ReadJournal()
```
bgilbert: `Destroy()` guards against `dm.journal == nil`, and `ConsoleOutput` is called from `dm.cluster.DelMach()`, so this will need the guard as well.

platform/journal.go
```diff
@@ -112,6 +117,14 @@ func (j *Journal) Start(ctx context.Context, m Machine) error {
 	return nil
 }
 
+func (j *Journal) ReadJournal() ([]byte, error) {
```
bgilbert: Probably needs a comment that there are no guarantees about the returned bytes if called before `Destroy()`.

Maybe just call this `Read()`?  `journal.ReadJournal` seems redundant.

arithx: My initial thought for calling it `ReadJournal` was to allow for distinctions if we ever added a `ReadRawJournal` but `Read` & `ReadRaw` would work as well.

arithx:
Change pushed addressing comments.

PR #862, opened 2018-06-04
platform/api/aws/ec2: create networking resources if missing

arithx:
Currently kola will only create a SecurityGroup if it is not found. With
this patch it will also create a VPC attached to the SecurityGroup (as
well as the route table, internet gateway, and subnets). This allows
kola to be ran against a fresh account/region and work out of the box.

This patch also adds the ability to detect if searching security groups
by name via filters is disabled and manually filters the list.

### DIFF
--- a/platform/api/aws/ec2.go
+++ b/platform/api/aws/ec2.go
@@ -61,6 +61,17 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
 	if err != nil {
 		return nil, fmt.Errorf("error resolving security group: %v", err)
 	}
+
+	vpcId, err := a.getVPCID(sgId)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving vpc: %v", err)
+	}
+
+	subnetId, err := a.getSubnetID(vpcId)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving subnet: %v", err)
+	}
+
 	key := &keyname
 	if keyname == "" {
 		key = nil
@@ -72,6 +83,7 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
 		KeyName:          key,
 		InstanceType:     &a.opts.InstanceType,
 		SecurityGroupIds: []*string{&sgId},
+		SubnetId:         &subnetId,
 		UserData:         ud,
 		IamInstanceProfile: &ec2.IamInstanceProfileSpecification{
 			Name: &a.opts.IAMInstanceProfile,
--- a/platform/api/aws/ec2.go
+++ b/platform/api/aws/ec2.go
@@ -20,7 +20,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
 
 	"github.com/coreos/mantle/util"
@@ -238,90 +237,4 @@ func (a *API) GetConsoleOutput(instanceID string) (string, error) {
 	}
 
 	return string(decoded), nil
-}
-
-// getSecurityGroupID gets a security group matching the given name.
-// If the security group does not exist, it's created.
-func (a *API) getSecurityGroupID(name string) (string, error) {
-	sgIds, err := a.ec2.DescribeSecurityGroups(&ec2.DescribeSecurityGroupsInput{
-		GroupNames: []*string{&name},
-	})
-	if isSecurityGroupNotExist(err) {
-		return a.createSecurityGroup(name)
-	}
-
-	if err != nil {
-		return "", fmt.Errorf("unable to get security group named %v: %v", name, err)
-	}
-	if len(sgIds.SecurityGroups) == 0 {
-		return "", fmt.Errorf("zero security groups matched name %v", name)
-	}
-	return *sgIds.SecurityGroups[0].GroupId, nil
-}
-
-// createSecurityGroup creates a security group with tcp/22 access allowed from the
-// internet.
-func (a *API) createSecurityGroup(name string) (string, error) {
-	sg, err := a.ec2.CreateSecurityGroup(&ec2.CreateSecurityGroupInput{
-		GroupName:   aws.String(name),
-		Description: aws.String("mantle security group for testing"),
-	})
-	if err != nil {
-		return "", err
-	}
-	plog.Debugf("created security group %v", *sg.GroupId)
-
-	allowedIngresses := []ec2.AuthorizeSecurityGroupIngressInput{
-		{
-			// SSH access from the public internet
-			GroupId: sg.GroupId,
-			IpPermissions: []*ec2.IpPermission{
-				{
-					IpProtocol: aws.String("tcp"),
-					IpRanges: []*ec2.IpRange{
-						{
-							CidrIp: aws.String("0.0.0.0/0"),
-						},
-					},
-					FromPort: aws.Int64(22),
-					ToPort:   aws.Int64(22),
-				},
-			},
-		},
-		{
-			// Access from all things in this vpc with the same SG (e.g. other
-			// machines in our kola cluster)
-			GroupId:                 sg.GroupId,
-			SourceSecurityGroupName: aws.String(name),
-		},
-	}
-
-	for _, input := range allowedIngresses {
-		_, err := a.ec2.AuthorizeSecurityGroupIngress(&input)
-
-		if err != nil {
-			// We created the SG but can't add all the needed rules, let's try to
-			// bail gracefully
-			_, delErr := a.ec2.DeleteSecurityGroup(&ec2.DeleteSecurityGroupInput{
-				GroupId: sg.GroupId,
-			})
-			if delErr != nil {
-				return "", fmt.Errorf("created sg %v (%v) but couldn't authorize it. Manual deletion may be required: %v", *sg.GroupId, name, err)
-			}
-			return "", fmt.Errorf("created sg %v (%v), but couldn't authorize it and thus deleted it: %v", *sg.GroupId, name, err)
-		}
-	}
-	return *sg.GroupId, err
-}
-
-func isSecurityGroupNotExist(err error) bool {
-	if err == nil {
-		return false
-	}
-	if awsErr, ok := (err).(awserr.Error); ok {
-		if awsErr.Code() == "InvalidGroup.NotFound" {
-			return true
-		}
-	}
-	return false
 }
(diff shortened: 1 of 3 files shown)


platform/api/aws/ec2.go
```diff
@@ -255,12 +283,192 @@ func (a *API) getSecurityGroupID(name string) (string, error) {
...
+			{
+				Name:   aws.String("vpc-id"),
+				Values: []*string{&vpc},
+			},
+		},
+	})
+	if err != nil {
+		return "", fmt.Errorf("unable to get subnets for vpc %v: %v", vpc, err)
+	}
+	for _, id := range subIds.Subnets {
```
ashcrow: nit: Since this will get the first subnet id should the method be named ``getFirstSubnetID``?

arithx: This functions the same as the pre-existing `getSecurityGroupID` which also will return the first entry if one exists (otherwise one is created)

platform/api/aws/ec2.go
```diff
@@ -255,12 +283,192 @@ func (a *API) getSecurityGroupID(name string) (string, error) {
...
+	return "", fmt.Errorf("no subnets found for vpc %v: %v", vpc, err)
+}
+
+// creates an InternetGateway and attaches it to the given VPC
+func (a *API) createInternetGateway(vpcId *string) (string, error) {
+	igw, err := a.ec2.CreateInternetGateway(&ec2.CreateInternetGatewayInput{})
+	if err != nil {
+		return "", err
+	}
+	a.CreateTags([]string{*igw.InternetGateway.InternetGatewayId}, map[string]string{
```
ashcrow: nit: I notice the ``"CreatedBy": "mantle"`` is used multiple times. In the future it may be worth making a tags variable which gets reused for these calls. No biggie though. Just a thought.

platform/api/aws/ec2.go
ashcrow: OK, fair enough.

platform/api/aws/ec2.go
```diff
@@ -310,6 +532,12 @@ func (a *API) createSecurityGroup(name string) (string, error) {
 	return *sg.GroupId, err
 }
 
+func (a *API) tagCreatedByMantle(resources []string) error {
```
ashcrow: :+1:

platform/api/aws/ec2.go
```diff
@@ -284,12 +492,6 @@ func (a *API) createSecurityGroup(name string) (string, error) {
 				},
 			},
 		},
-		{
-			// Access from all things in this vpc with the same SG (e.g. other
-			// machines in our kola cluster)
-			GroupId:                 sg.GroupId,
-			SourceSecurityGroupName: aws.String(name),
```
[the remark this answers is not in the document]
arithx: As a follow up, I had to move from using `SourceSecurityGroup` to setting up `UserIdGroupPairs` for each `IpProtocol` type.

platform/api/aws/ec2.go
```diff
@@ -240,27 +252,243 @@ func (a *API) GetConsoleOutput(instanceID string) (string, error) {
 // If the security group does not exist, it's created.
 func (a *API) getSecurityGroupID(name string) (string, error) {
 	sgIds, err := a.ec2.DescribeSecurityGroups(&ec2.DescribeSecurityGroupsInput{
-		GroupNames: []*string{&name},
+		Filters: []*ec2.Filter{
```
arithx: Using the `group-name` Filter over the `GroupNames` parameter distances the check from default VPC's.

As a result this also allows us to drop the `isSecurityGroupNotExist` function.

arithx:
This is ready for rounds of review.

The only outstanding issue that remains is if a kola test which has multiple machines is ran directly (skipping the semver check) then multiple VPC's will be created. My idea for potentially working around that would be an additional `getSecurityGroup` call during early cluster creation.

ashcrow:
@arithx it may be worth adding the "outstanding issue" in the documentation until a fix/workaround is added.

platform/api/aws/ec2.go
```diff
@@ -240,27 +252,243 @@ func (a *API) GetConsoleOutput(instanceID string) (string, error) {
...
+	err = a.tagCreatedByMantle([]string{*vpc.Vpc.VpcId})
+	if err != nil {
+		return "", err
+	}
+
+	_, err = a.ec2.ModifyVpcAttribute(&ec2.ModifyVpcAttributeInput{
+		EnableDnsHostnames: &ec2.AttributeBooleanValue{
+			Value: aws.Bool(true),
+		},
+		VpcId: vpc.Vpc.VpcId,
```
ashcrow: did I stutter

platform/api/aws/ec2.go
```diff
@@ -240,27 +252,243 @@ func (a *API) GetConsoleOutput(instanceID string) (string, error) {
...
+			{
+				Name:   aws.String("group-name"),
+				Values: []*string{&name},
+			},
+		},
 	})
-	if isSecurityGroupNotExist(err) {
+
+	if len(sgIds.SecurityGroups) == 0 {
 		return a.createSecurityGroup(name)
```
ashcrow: I realize this wasn't change in this PR but it may be worth updating in the future if there is a possibility that creating a security group could error.

platform/api/aws/ec2.go
ashcrow: This would be a perfect comment to add to the code.

platform/api/aws/ec2.go
arithx: `createSecurityGroup` also returns `string, error` so if this function were to fail it would bubble out properly and should be detected in whatever function references this one.

platform/api/aws/ec2.go
```diff
@@ -239,28 +251,246 @@ func (a *API) GetConsoleOutput(instanceID string) (string, error) {
...
+	})
+	if err != nil {
+		return "", fmt.Errorf("attaching internet gateway to vpc: %v", err)
+	}
+	return *igw.InternetGateway.InternetGatewayId, nil
+}
+
+// createSubnets creates a subnet in each availability zone for the region
+// that is associated with the given VPC associated with the given RouteTable
+func (a *API) createSubnets(vpcId, routeTableId *string) error {
```
bgilbert: These functions seem to be in somewhat arbitrary order.

platform/api/aws/ec2.go
```diff
@@ -239,28 +251,246 @@ func (a *API) GetConsoleOutput(instanceID string) (string, error) {
...
+	azs, err := a.ec2.DescribeAvailabilityZones(&ec2.DescribeAvailabilityZonesInput{})
+	if err != nil {
+		return fmt.Errorf("retrieving availability zones: %v", err)
+	}
+
+	for i, az := range azs.AvailabilityZones {
+		// 15 is the maximum amount of zones possible when giving them a /20
```
bgilbert: Returning on `i > 15` implies 16 zones.

platform/api/aws/ec2.go
```diff
@@ -239,28 +251,246 @@ func (a *API) GetConsoleOutput(instanceID string) (string, error) {
...
+			AvailabilityZone: aws.String(name),
+			VpcId:            vpcId,
+			// Increment the CIDR block by 16 every time
+			CidrBlock: aws.String(fmt.Sprintf("172.31.%d.0/20", i*16)),
+		})
+		if err != nil {
+			// Some availability zones get returned but cannot have subnets
+			// created inside of them
+			if awsErr, ok := (err).(awserr.Error); ok {
+				if awsErr.Code() == "InvalidParameterValue" {
```
bgilbert: Nit: could use `ok && awsErr.Code() ==` to avoid the nesting.

platform/api/aws/ec2.go
arithx: Ah, yeah, was referring to the variable rather than the actual amount. Updating the comment to be 16 & switching the error check to `i < 16` for consistency.

platform/api/aws/ec2.go
```diff
@@ -239,28 +251,246 @@ func (a *API) GetConsoleOutput(instanceID string) (string, error) {
...
+	}
+
+	_, err = a.ec2.ModifyVpcAttribute(&ec2.ModifyVpcAttributeInput{
+		EnableDnsHostnames: &ec2.AttributeBooleanValue{
+			Value: aws.Bool(true),
+		},
+		VpcId: vpc.Vpc.VpcId,
+	})
+	if err != nil {
+		return "", fmt.Errorf("modifying VPC attributes: %v", err)
```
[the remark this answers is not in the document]
arithx: Changed each one to distinguish which attribute is being set (the fact you can't do both in one is a serious SDK failure).

platform/api/aws/ec2.go
```diff
@@ -239,28 +251,246 @@ func (a *API) GetConsoleOutput(instanceID string) (string, error) {
...
 	}
-	if len(sgIds.SecurityGroups) == 0 {
-		return "", fmt.Errorf("zero security groups matched name %v", name)
-	}
+
 	return *sgIds.SecurityGroups[0].GroupId, nil
 }
 
+// getVPCID gets a VPC for the given security group
+func (a *API) getVPCID(sgId string) (string, error) {
```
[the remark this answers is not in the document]
arithx: Dropped these from `ec2.go` & created `network.go`

platform/api/aws/ec2.go
arithx: Moved all the functions into alphabetical order in the new file.

arithx:
Pushed addressing comments.

platform/api/aws/ec2.go
bgilbert: In general I prefer ordering based on caller-callee relationships (and that's what we do elsewhere in mantle) but can live with alphabetical.

bgilbert:
Could you move the existing code in a separate commit from changing it?

arithx:
Will do.

arithx:
Pushed (the wrong order on github is because it's using the chronological dates rather than the order from `git log`)

platform/api/aws/network.go
```diff
@@ -96,14 +168,175 @@ func (a *API) createSecurityGroup(name string) (string, error) {
...
+	azs, err := a.ec2.DescribeAvailabilityZones(&ec2.DescribeAvailabilityZonesInput{})
+	if err != nil {
+		return fmt.Errorf("retrieving availability zones: %v", err)
+	}
+
+	for i, az := range azs.AvailabilityZones {
+		// 16 is the maximum amount of zones possible when giving them a /20
+		// CIDR range inside of a /16 network.
+		if i < 16 {
+			return nil
```
bgilbert: This will always create zero subnets.

arithx: Ah yeah, my brain failed me there.

arithx:
Pushed with the ordering change to caller-callee.

arithx:
Changes re-tested

PR #864, opened 2018-06-06
platform/aws/images: drop pv in unsupp. regions

ajeddeloh:
Do not merge until after the release is finished, it's having a hard enough time as it is.

Don't create PV ami's in regions that do not support it.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/virtualization_types.html has the PV supported regions


ajeddeloh:
I'm totally fine with waiting till alpha 4 to merge. Also agree we should release note it. I'll merge after the triple release.

PR #865, opened 2018-06-08
cork: drop deprecated enter method

ajeddeloh:
depends on https://github.com/coreos/scripts/pull/820

### DIFF
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -49,7 +49,6 @@ var (
 
 	// only for `enter` command
 	experimental bool
-	legacy       bool
 	bindGpgAgent bool
 
 	// only for `update` command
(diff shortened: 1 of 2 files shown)


ajeddeloh:
Passed jenkins

cmd/cork/create.go
```diff
@@ -127,9 +126,7 @@ func init() {
 	enterCmd.Flags().AddFlagSet(chrootFlags)
 	enterCmd.Flags().BoolVar(&experimental,
 		"experimental", false, "Use new enter implementation")
-	enterCmd.Flags().MarkDeprecated("experimental", "--experimental is now default and the flag will be removed soon. Implies --bind-gpg-agent=false")
-	enterCmd.Flags().BoolVar(&legacy,
-		"legacy", false, "Use the old enter implementation")
+	enterCmd.Flags().MarkDeprecated("experimental", "--experimental is deprecated and will be removed next release. Use --bind-gpg-agent=false")
```
bgilbert: `Use` is probably too strong; it's true for Jenkins but probably not for anyone else.

dm0-: Should the flag be changed in the scripts PR as part of this change set?  The change isn't required, but if we're removing it in the next release, we might as well update.

ajeddeloh: @bgilbert how about `This is equivilent to --bind-gpg-agent=false`?

bgilbert: Could just be `Equivalent to --bind-gpg-agent=false`, but sure.  I'm also okay with the wording of the existing second sentence.

PR #866, opened 2018-06-08
kola/tests: add test to fetch urls using TLS

### LINKED ISSUE #571: kola: Test for working TLS
Check that curl and wget can fetch HTTPS URLs without throwing certificate or protocol errors.

rfairley:
Uses curl and wget commands to query a list of
popular HTTPS URLs, to check that TLS works on
the CL image under test.

closes #571 

**Notes**
- This addition has been tested using AWS, after building the mantle tools on top of changes in #862 (along with the changes in this PR)
- Different URLs may be used - there is no particular reason for including the current URLs in `URLsToFetch`

### DIFF
--- a/kola/tests/misc/tls.go
+++ b/kola/tests/misc/tls.go
@@ -1,0 +1,34 @@
+package misc
+
+import (
+	"fmt"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+)
+
+var (
+	urlsToFetch = []string{
+		"https://www.example.com/",
+		"https://www.wikipedia.org/",
+		"https://start.fedoraproject.org/",
+	}
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:              TestTLSFetchURLs,
+		ClusterSize:      1,
+		Name:             "coreos.tls.fetch-urls",
+		ExcludePlatforms: []string{"qemu"}, // Networking outside cluster required
+	})
+}
+
+func TestTLSFetchURLs(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	for _, url := range urlsToFetch {
+		c.MustSSH(m, fmt.Sprintf("curl -s -S -m 30 --retry 2 %s", url))
+		c.MustSSH(m, fmt.Sprintf("wget -nv -T 30 -t 2 --delete-after %s", url))
+	}
+}


kola/tests/misc/tls.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"fmt"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+)
+
+var (
+	URLsToFetch = []string{
```
ajeddeloh: lowercase since it doesn't need to be public

rfairley:
@ajeddeloh, I amended the commit and changed `URLsToFetch` to `urlsToFetch` - how's this?

kola/tests/misc/tls.go
```diff
@@ -0,0 +1,34 @@
...
+		ExcludePlatforms: []string{"qemu"}, // Networking outside cluster required
+	})
+}
+
+func TestTLSFetchURLs(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	for _, url := range urlsToFetch {
+		c.MustSSH(m, fmt.Sprintf("curl %s", url))
+		c.MustSSH(m, fmt.Sprintf("wget --delete-after %s", url))
```
arithx: The only default timeout for `wget` is the read-timeout at 900 seconds, in the case of broken TLS this would most likely hang for inordinate amounts of time.

Let's maybe switch this to `wget -T 30 -t 2 --delete-after %s` so we have a maximum of 1 minute trying to fetch the URL (the default tries are also 10).

We could also consider adding `-q` to quiet the logs down (as they aren't particularly useful).

kola/tests/misc/tls.go
```diff
@@ -0,0 +1,34 @@
...
+		Name:             "coreos.tls.fetch-urls",
+		ExcludePlatforms: []string{"qemu"}, // Networking outside cluster required
+	})
+}
+
+func TestTLSFetchURLs(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	for _, url := range urlsToFetch {
+		c.MustSSH(m, fmt.Sprintf("curl %s", url))
```
arithx: We should add a retry amount, timeout and consider adding silent & show-error.

Both silent & show-error will still output in case of an error but will disable the progress bar.

`curl -s -S -m 30 --retry 2 %s`

arithx:
Also, can you update the description from `Implements #571` to `closes #571` or `fixes #571` so the [keyword automation](https://help.github.com/articles/closing-issues-using-keywords/) can trigger?

kola/tests/misc/tls.go
rfairley: Thanks - adding these in now.

Adding the `-q` to `wget` sounds good, as the test will still fail in the case of an error due to the non-zero exit code `wget` will give. Seems there is not a clean way to have `wget` log errors only - however `-nv` gives one line of output only for successes (when downloading only one web page at a time), and one line for failures if we err on the side of logging more. Could `-nv` be worth using instead of `-q`?

kola/tests/misc/tls.go
rfairley: Thank you, will add these in.

kola/tests/misc/tls.go
arithx: Yeah `-nv` works. Just trying to reduce the general amount of log spam, having a single line works.

rfairley:
@arithx Will change the description and will keep in mind the keywords for next time, thanks!

rfairley:
@arithx @ajeddeloh the updated changes are now pushed, with the additional options for `curl` and `wget`.

One concern is that `wget` will still send output to the console when a test passes, using `-nv`, for example:
```
=== RUN   coreos.tls.fetch-urls
--- PASS: coreos.tls.fetch-urls (92.97s)
        cluster.go:104: 2018-06-11 15:19:58 URL:https://www.example.com/ [1270/1270] -> "index.html.tmp" [1]
        cluster.go:104: 2018-06-11 15:19:59 URL:https://www.wikipedia.org/ [75576/75576] -> "index.html.tmp" [1]
        cluster.go:104: 2018-06-11 15:20:00 URL:https://start.fedoraproject.org/ [16610/16610] -> "index.html.tmp" [1]
PASS, output in _kola_temp/aws-2018-06-11-1118-10150
```
Most tests show `--- PASS: ...` and no other output, when a test passes, whereas with `wget -nv` the "success" output is still sent to `stderr`, and thus shows even when a test passes (as above). Ideally, additional output would show only after a `--- FAIL:`. I can't think of a way to get around this other than filtering the `stderr` output manually within the `MustSSH` function.

I'm thinking it is acceptable to allow the additional output (as above), on the basis that only one line is logged to `stderr` so it isn't too spammy. The other option is using `-q` to suppress the output completely - then no `stderr` output from `wget` will be logged (which might be OK since curl will get logged).

Let me know what you think - if having the extra output after `--- PASS:` is OK, then this PR can be merged. Otherwise, the `-nv` can be changed to `-q`.

bgilbert:
To avoid the extra output, you could do

```shell
set -o pipefail; wget -nv -T 30 -t 2 --delete-after %s 2>&1 | (grep -v -- '->' >&2 ||:)
```

but admittedly that's a little extreme.

bgilbert:
It's much more tolerable using process substitution:

```bash
wget -nv -T 30 -t 2 --delete-after %s 2> >(grep -v -- '->' >&2)
```

rfairley:
@bgilbert , I tested the process substitution command on AWS and `wget` shows output in console on fail (from `stderr` and `_kola_temp` logs) as expected. I think this change can go in. @arithx  @ajeddeloh WDYT?

PR #869, opened 2018-06-13
[WIP] ore: Add functionality to add custom tags to AMIs

sayanchowdhury:
Signed-off-by: Sayan Chowdhury <sayan.chowdhury2012@gmail.com>

### DIFF
--- a/cmd/ore/aws/upload.go
+++ b/cmd/ore/aws/upload.go
@@ -40,7 +40,8 @@ After a successful run, the final line of output will be a line of JSON describi
 		Example: `  ore aws upload --region=us-east-1 \
 	  --ami-name="CoreOS-stable-1234.5.6" \
 	  --ami-description="CoreOS stable 1234.5.6" \
-	  --file="/home/.../coreos_production_ami_vmdk_image.vmdk"`,
+	  --file="/home/.../coreos_production_ami_vmdk_image.vmdk" \
+	  --tags="machine=production"`,
 		RunE: runUpload,
 	}
 
@@ -57,6 +58,7 @@ After a successful run, the final line of output will be a line of JSON describi
 	uploadAMIDescription string
 	uploadGrantUsers     []string
 	uploadCreatePV       bool
+	uploadTags           []string
 )
 
 func init() {
@@ -76,6 +78,7 @@ func init() {
 	cmdUpload.Flags().StringVar(&uploadAMIDescription, "ami-description", "", "description of the AMI to create (default: empty)")
 	cmdUpload.Flags().StringSliceVar(&uploadGrantUsers, "grant-user", []string{}, "grant launch permission to this AWS user ID")
 	cmdUpload.Flags().BoolVar(&uploadCreatePV, "create-pv", false, "create a PV AMI in addition to the HVM AMI")
+	cmdUpload.Flags().StringSliceVar(&uploadTags, "tags", []string{}, "list to tags to attach to the AMI")
 }
 
 func defaultBucketNameForRegion(region string) string {
@@ -236,6 +239,21 @@ func runUpload(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	tagMap := make(map[string]string)
+	if len(uploadTags) > 0 {
+		for _, tag := range uploadTags {
+			splitTag := strings.SplitN(tag, "=", 2)
+			key, value := splitTag[0], splitTag[1]
+			tagMap[key] = value
+		}
+		err = API.CreateTags([]string{hvmID}, tagMap)
+
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "unable to add tags: %v\n", err)
+			os.Exit(1)
+		}
+	}
+
 	var pvID string
 	if uploadCreatePV {
 		pvImageID, err := API.CreatePVImage(sourceSnapshot, amiName, uploadAMIDescription)
@@ -249,6 +267,15 @@ func runUpload(cmd *cobra.Command, args []string) error {
 			err = API.GrantLaunchPermission(pvID, uploadGrantUsers)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "unable to grant launch permission: %v\n", err)
+				os.Exit(1)
+			}
+		}
+
+		if len(uploadTags) > 0 {
+			err = API.CreateTags([]string{pvID}, tagMap)
+
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "unable to add tags: %v\n", err)
 				os.Exit(1)
 			}
 		}


cmd/ore/aws/upload.go
```diff
@@ -236,6 +238,21 @@ func runUpload(cmd *cobra.Command, args []string) error {
...
+		for _, tag := range uploadTags {
+			splitTag := strings.Split(tag, "=")
+			key, value := splitTag[0], splitTag[1]
+			hvmTagMap[key] = value
+		}
+		err = API.CreateTags([]string{hvmID}, hvmTagMap)
+
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "unable to add tags: %v\n", err)
+			os.Exit(1)
```
ashcrow: This function expects error returns. Would it make more sense to return the error?

cmd/ore/aws/upload.go
```diff
@@ -252,6 +269,21 @@ func runUpload(cmd *cobra.Command, args []string) error {
 				os.Exit(1)
 			}
 		}
+
+		if len(uploadTags) > 0 {
```
ashcrow: The logic here makes sense to me **but** it looks identical to the previous block. Maybe this should become a function?

cmd/ore/aws/upload.go
```diff
@@ -236,6 +238,21 @@ func runUpload(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	if len(uploadTags) > 0 {
+		hvmTagMap := make(map[string]string)
```
arithx: Rather than duplicate this code you could build a single tagMap and then only the call to `CreateTags` and the error check would be present in both image blocks.

cmd/ore/aws/upload.go
```diff
@@ -236,6 +238,21 @@ func runUpload(cmd *cobra.Command, args []string) error {
...
+		for _, tag := range uploadTags {
+			splitTag := strings.Split(tag, "=")
```
jlebon: Hmm, we should probably use `SplitN(tag, "=", 1)` here instead to handle values that have `=`?

cmd/ore/aws/upload.go
arithx: Using `os.Exit`'s in `ore` is pretty normal (though most functions don't return `error`'s inside of `ore`). This is following the rest of the checks inside of this function.

ashcrow:
@sayanchowdhury good work so far though!!

cmd/ore/aws/upload.go
bgilbert: Returning an error from the run function causes the usage message to be displayed, which doesn't make sense for actual errors.

cmd/ore/aws/upload.go
```diff
@@ -76,6 +77,7 @@ func init() {
 	cmdUpload.Flags().StringVar(&uploadAMIDescription, "ami-description", "", "description of the AMI to create (default: empty)")
 	cmdUpload.Flags().StringSliceVar(&uploadGrantUsers, "grant-user", []string{}, "grant launch permission to this AWS user ID")
 	cmdUpload.Flags().BoolVar(&uploadCreatePV, "create-pv", false, "create a PV AMI in addition to the HVM AMI")
+	cmdUpload.Flags().StringSliceVar(&uploadTags, "tags", []string{}, "list to tags to attach to the AMI")
```
bgilbert: Should document that this takes `key=value` pairs.

cmd/ore/aws/upload.go
```diff
@@ -236,6 +238,21 @@ func runUpload(cmd *cobra.Command, args []string) error {
...
 	}
 
+	if len(uploadTags) > 0 {
+		hvmTagMap := make(map[string]string)
+		for _, tag := range uploadTags {
+			splitTag := strings.Split(tag, "=")
+			key, value := splitTag[0], splitTag[1]
+			hvmTagMap[key] = value
+		}
+		err = API.CreateTags([]string{hvmID}, hvmTagMap)
```
bgilbert: This does not tag the underlying snapshot.  Is that intentional?

cmd/ore/aws/upload.go
sayanchowdhury: Documented this in the example, but IMO we should have a detailed documentation for the commands. I could not find one, blame my poor if it already exists.

sayanchowdhury:
@arithx I am +1 on creating a `ore aws tag` which would give more freedom to tag resources.

cmd/ore/aws/upload.go
arithx: @bgilbert: Is there a particular reason we'd want to tag the snapshot?

cmd/ore/aws/upload.go
bgilbert: The help text is the primary detailed documentation right now.

cmd/ore/aws/upload.go
bgilbert: Deleting an AMI doesn't automatically delete the underlying snapshot, resulting in a stray snapshot without much context about why it exists.  So I'd be in favor of tagging the snapshot as well, unless there's a reason to do otherwise.

bgilbert:
@sayanchowdhury There are no tests for ore.

dustymabe:
anything left blocking this from being merged?

arithx:
@dustymabe only remaining thing is to tag the underlying snapshot (https://github.com/coreos/mantle/pull/869#discussion_r195149346).

cmd/ore/aws/upload.go
ashcrow: @sayanchowdhury Any progress?

arithx: @ashcrow @sayanchowdhury is on vacation. @mike-nguyen picked it up in https://github.com/coreos/mantle/pull/873

cmd/ore/aws/upload.go
```diff
@@ -236,6 +239,21 @@ func runUpload(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	tagMap := make(map[string]string)
+	if len(uploadTags) > 0 {
+		for _, tag := range uploadTags {
+			splitTag := strings.SplitN(tag, "=", 2)
+			key, value := splitTag[0], splitTag[1]
```
ajeddeloh: This will panic on bad input. Should either make this print and error and exit if `len(splitTag) != 2`

cmd/ore/aws/upload.go
```diff
@@ -236,6 +239,21 @@ func runUpload(cmd *cobra.Command, args []string) error {
...
+			tagMap[key] = value
+		}
+		err = API.CreateTags([]string{hvmID}, tagMap)
```
ajeddeloh: style nit: I'm typically a fan of the `if err := func(); err != nil {` style but this is also fine. We do typically keep the if next to the err assignment with no line break, however

cmd/ore/aws/upload.go
ajeddeloh: agreed.

cmd/ore/aws/upload.go
ashcrow: No biggie, but I agree.

dustymabe:
hey @ajeddeloh - since sayan is AFK this week could we have someone make those changes in a followup PR?

arithx:
Per offline discussions let's make the changes in #873 and we'll just merge that one (leaving attribution on the top commit to Sayan).

PR #870, opened 2018-06-13
kola/tests: tls test - direct wget failure output only to stderr

rfairley:
uses process subsitution and grep to filter stderr output,
so only failure messages are logged in stderr


rfairley:
Follow-up to #866 .

PR #871, opened 2018-06-15
WIP: network/journal: use sudo for journalctl recording

arithx:
RHCOS requires sudo for journalctl (and adding it doesn't effect CL).


arithx:
Thanks @dm0, meant to mark this as a WIP.

arithx:
Dropping in favor of a workaround to add the correct groups to the core user on other platforms.

PR #873, opened 2018-06-20
ore/aws/upload: add tags to snapshots

mike-nguyen:
Add tags to snapshots so there is still information about a snapshot
if an ami is deleted but the snapshot is not.

#869 must be merged first

### DIFF
--- a/cmd/ore/aws/upload.go
+++ b/cmd/ore/aws/upload.go
@@ -40,7 +40,8 @@ After a successful run, the final line of output will be a line of JSON describi
 		Example: `  ore aws upload --region=us-east-1 \
 	  --ami-name="CoreOS-stable-1234.5.6" \
 	  --ami-description="CoreOS stable 1234.5.6" \
-	  --file="/home/.../coreos_production_ami_vmdk_image.vmdk"`,
+	  --file="/home/.../coreos_production_ami_vmdk_image.vmdk" \
+	  --tags="machine=production"`,
 		RunE: runUpload,
 	}
 
@@ -57,6 +58,7 @@ After a successful run, the final line of output will be a line of JSON describi
 	uploadAMIDescription string
 	uploadGrantUsers     []string
 	uploadCreatePV       bool
+	uploadTags           []string
 )
 
 func init() {
@@ -76,6 +78,7 @@ func init() {
 	cmdUpload.Flags().StringVar(&uploadAMIDescription, "ami-description", "", "description of the AMI to create (default: empty)")
 	cmdUpload.Flags().StringSliceVar(&uploadGrantUsers, "grant-user", []string{}, "grant launch permission to this AWS user ID")
 	cmdUpload.Flags().BoolVar(&uploadCreatePV, "create-pv", false, "create a PV AMI in addition to the HVM AMI")
+	cmdUpload.Flags().StringSliceVar(&uploadTags, "tags", []string{}, "list to tags to attach to the AMI")
 }
 
 func defaultBucketNameForRegion(region string) string {
@@ -236,6 +239,21 @@ func runUpload(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	tagMap := make(map[string]string)
+	if len(uploadTags) > 0 {
+		for _, tag := range uploadTags {
+			splitTag := strings.SplitN(tag, "=", 2)
+			key, value := splitTag[0], splitTag[1]
+			tagMap[key] = value
+		}
+		err = API.CreateTags([]string{hvmID}, tagMap)
+
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "unable to add tags: %v\n", err)
+			os.Exit(1)
+		}
+	}
+
 	var pvID string
 	if uploadCreatePV {
 		pvImageID, err := API.CreatePVImage(sourceSnapshot, amiName, uploadAMIDescription)
@@ -249,6 +267,15 @@ func runUpload(cmd *cobra.Command, args []string) error {
 			err = API.GrantLaunchPermission(pvID, uploadGrantUsers)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "unable to grant launch permission: %v\n", err)
+				os.Exit(1)
+			}
+		}
+
+		if len(uploadTags) > 0 {
+			err = API.CreateTags([]string{pvID}, tagMap)
+
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "unable to add tags: %v\n", err)
 				os.Exit(1)
 			}
 		}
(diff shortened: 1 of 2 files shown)


cmd/ore/aws/upload.go
```diff
@@ -252,6 +270,15 @@ func runUpload(cmd *cobra.Command, args []string) error {
 				os.Exit(1)
 			}
 		}
+
+		if len(uploadTags) > 0 {
+			err = API.CreateTags([]string{pvID, sourceSnapshot}, tagMap)
```
arithx: You don't need to retag `sourceSnapshot` here.

mike-nguyen: Thanks!  Fixed.

dustymabe:
any blockers for merge of this (and #869, first) ?

arithx:
Just needs a second maintainer LGTM on each.

cmd/ore/aws/upload.go
```diff
@@ -122,6 +122,26 @@ func defaultBucketURL(urlPrefix, imageName, board, file, region string) (*url.UR
 	return s3URL, nil
 }
 
+func createTags(tags []string, res []string) {
+	if len(tags) > 0 {
+		tagMap := make(map[string]string)
```
arithx: In switching to a split function we've reverted to creating the same `tagMap` multiple times from input.

You can take a couple of different paths to not have to create the map multiple times -- create the tag map and pass it in to `createTags` or move the actual length check + tagging functionality into a lambda inside of `runUpload` (rather than making this `createTags` function) that accepts a list of ids to tag would both work off the top of my head.

@ajeddeloh do you have any preferences on this?

ajeddeloh: Yeah, we should ideally break that out so we're not duplicating work. Since you can loop over 0 length arrays and call length on maps, it should be fine to just move this out.

It may even be ok to not both checking if there are tags to create and blindly call API.CreateTags, but that just wraps the aws go sdk and I don't know if that will be ok with it. If it is, I'd say drop all the conditionals entirely. Hell, if it isn't we could also modify API.CreateTags to just return nil on empty tag lists.

mike-nguyen: I was trying to think of way to not create the `tagMap` multiple times without more code/duplication than in the original PR.

If I move the creation of `tagMap` back into `runUpload`, wouldn't I still need to check `len(uploadTags) > 0` or `len(tagMap) > 0` before calling my `createTags` or `API.CreateTags` for tagging the PV?  Let me know if I'm missing something.

ajeddeloh: You _might_ need to check that. Depends on if the aws go sdk allows you to call their CreateTags with and empty tag list. If it does, then you can just blindly call it. I'm personally not a fan of functions that die when you hand them empty lists/maps/etc, so if that is the case, I think we ought to fix [our wrapper](https://github.com/coreos/mantle/blob/master/platform/api/aws/ec2.go#L199) to just return early if there's nothing to do.

arithx: You could leave the original instantiation of `tagMap` from the previous PR and then do the following (inside of `runUpload`):

```golang
tagger := func(resources []string) {
	if len(uploadTags) > 0 {
		err = API.CreateTags(resources, tagMap)

		if err != nil {
			fmt.Fprintf(os.Stderr, "unable to add tags: %v\n", err)
			os.Exit(1)
		}
	}
}

...

tagger([]string{hvmID, sourceSnapshot})
```

mike-nguyen: I could wrap the API.CreateTags like this

```
func createTags(t map[string]string, resource []string) {
  if t != nil {
    API.CreateTags(resource, t)
  }
}

// back in runUpload...
var tagMap map[string]string
if len(uploadTags) > 0 {
  // create tagMap
  createTags(tagMap, []string{hvmID, sourceSnapshot})
}

if uploadCreatePV {
  ...
  createTags(tagMap, []string{pvID})
}
```

arithx: If you were to go with that approach personally I'd move the part checking if the `uploadTags` length & `tagMap` creation onto their own so you could have something like:
```golang
var tagMap map[string]string
if len(uploadTags) > 0 {
    // create tagMap
}

...

createTags(tagMap, []string{hvmID, sourceSnapshot})

...

if uploadCreatePV {
    ...
    createTags(tagMap, []string{pvID})
}
```

But that's a pretty minor stylistic thing on my part that I'd be fine either way with.

----

Overall though I'd lean more towards @ajeddeloh's approach to update the `CreateTags` function inside of the API to handle receiving a nil map (if `aws/aws-sdk-go` can't handle a nil map).

mike-nguyen: Yeah @ajeddeloh's approach seems the cleanest.  In going with that approach, should I just no-op and return nil in API.CreateTags when len(tags) == 0?

ajeddeloh: Yeah that sounds good. That probably also gets rid of the need for the wrapper function here. 

Also, golang has some weird edge cases with maps/slices and nil, so always check `len(things) == 0` not `thing == nil`.

arithx: If `aws/aws-sdk-go` doesn't properly handle `nil` tag maps then yes just add a check if the tags map is `nil` and if so immediately return `nil`.

ajeddeloh: The docs make it seem like it expects a non-zero-length list, so even if they allow it right now they may break that in the future. 

go map/slice nil weirdness:
https://play.golang.org/p/LRsYY9mHo3c

mike-nguyen: > The docs make it seem like it expects a non-zero-length list, so even if they allow it right now they may break that in the future.
go map/slice nil weirdness:
https://play.golang.org/p/LRsYY9mHo3c

Funny, the instructor for the golang course I just took spent some time on this.  I guess its the difference between declaration and initialization.  

https://blog.golang.org/go-maps-in-action

In the end it makes sense to always check length == 0 unless you are explicitly checking for a map that has been declared but not initialized.

mike-nguyen:
:arrow_up:  pushed changes based on the discussion

cmd/ore/aws/upload.go
```diff
@@ -236,6 +239,24 @@ func runUpload(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	tagMap := make(map[string]string)
+	if len(uploadTags) > 0 {
```
ajeddeloh: Don't need the if, looping over an empty/nil list is a no-op

mike-nguyen: :+1: Fixed it in latest push

ashcrow:
@mike-nguyen looks like this PR carries the commit from #869. Can we merge as is?

PR #874, opened 2018-06-20
*: add distro flag

arithx:
Adds `-l`/`--distribution` to `kola` which allows the distribution to be
specified. Two new test flags are also added: `Distros` &
`ExcludeDistros`.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -30,6 +30,7 @@ var (
 	kolaPlatform       string
 	defaultTargetBoard = sdk.DefaultBoard()
 	kolaPlatforms      = []string{"aws", "do", "esx", "gce", "packet", "qemu"}
+	kolaDistros        = []string{"cl", "rhcos"}
 	kolaDefaultImages  = map[string]string{
 		"amd64-usr": sdk.BuildRoot() + "/images/amd64-usr/latest/coreos_production_image.bin",
 		"arm64-usr": sdk.BuildRoot() + "/images/arm64-usr/latest/coreos_production_image.bin",
@@ -50,6 +51,7 @@ func init() {
 	sv(&outputDir, "output-dir", "", "Temporary output directory for test data and logs")
 	sv(&kola.TorcxManifestFile, "torcx-manifest", "", "Path to a torcx manifest that should be made available to tests")
 	root.PersistentFlags().StringVarP(&kolaPlatform, "platform", "p", "qemu", "VM platform: "+strings.Join(kolaPlatforms, ", "))
+	root.PersistentFlags().StringVarP(&kola.Options.Distribution, "distro", "b", "cl", "Distribution: "+strings.Join(kolaDistros, ", "))
 	root.PersistentFlags().IntVarP(&kola.TestParallelism, "parallel", "j", 1, "number of tests to run in parallel")
 	sv(&kola.TAPFile, "tapfile", "", "file to write TAP results to")
 	sv(&kola.Options.BaseName, "basename", "kola", "Cluster name prefix")
@@ -115,15 +117,21 @@ func syncOptions() error {
 	kola.PacketOptions.Board = kola.QEMUOptions.Board
 	kola.PacketOptions.GSOptions = &kola.GCEOptions
 
-	ok := false
-	for _, platform := range kolaPlatforms {
-		if platform == kolaPlatform {
-			ok = true
-			break
+	validateOption := func(name, item string, valid []string) error {
+		for _, v := range valid {
+			if v == item {
+				return nil
+			}
 		}
+		return fmt.Errorf("unsupported %v %q", name, item)
 	}
-	if !ok {
-		return fmt.Errorf("unsupport platform %q", kolaPlatform)
+
+	if err := validateOption("platform", kolaPlatform, kolaPlatforms); err != nil {
+		return err
+	}
+
+	if err := validateOption("distro", kola.Options.Distribution, kolaDistros); err != nil {
+		return err
 	}
 
 	image, ok := kolaDefaultImages[kola.QEMUOptions.Board]
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -183,34 +183,33 @@ func filterTests(tests map[string]*register.Test, pattern, platform string, vers
 			continue
 		}
 
-		allowed := true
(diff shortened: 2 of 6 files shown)


cmd/kola/options.go
```diff
@@ -50,6 +52,7 @@ func init() {
 	sv(&outputDir, "output-dir", "", "Temporary output directory for test data and logs")
 	sv(&kola.TorcxManifestFile, "torcx-manifest", "", "Path to a torcx manifest that should be made available to tests")
 	root.PersistentFlags().StringVarP(&kolaPlatform, "platform", "p", "qemu", "VM platform: "+strings.Join(kolaPlatforms, ", "))
+	root.PersistentFlags().StringVarP(&kolaDistro, "distro", "l", "cl", "Distribution: "+strings.Join(kolaDistros, ", "))
```
arithx: Looking for suggestions for a better short-name

arithx:
@ajeddeloh brought up a question offline about what this will mean for some of the automated things present in `kola` (like certain platforms automatically resolving CL channel names into current image IDs).

My current thought along that line is that we'd defer it for this PR; ultimately, I think it will end up where we create a new function in `BaseCluster` (this function would accept a string [the distribution] and would then handle calling functions like `resolveAMI` passing the distribution). With this we'd have one generic entrypoint defined but on platforms that do not need to perform distribution specific actions they can skip making this call or passing the distribution to an API that otherwise does not need it.

kola/harness.go
```diff
@@ -377,12 +377,17 @@ func getClusterSemver(pltfrm, distro, outputDir string) (*semver.Version, error)
 		return nil, fmt.Errorf("parsing /etc/os-release: %v: %s", err, stderr)
 	}
 
-	version, err := semver.NewVersion(strings.Split(string(out), "=")[1])
-	if err != nil {
-		return nil, fmt.Errorf("parsing os-release semver: %v", err)
+	// TODO: add distro specific version handling
+	if distro != "rhcos" {
```
ajeddeloh: I would prefer an
```
if distro == "rhcos" {
    return &semver.Version{}
}
```
to keep indentation minimal

cmd/kola/options.go
ajeddeloh: `-b` for `distroBution`?

cmd/kola/options.go
```diff
@@ -126,6 +129,17 @@ func syncOptions() error {
 		return fmt.Errorf("unsupport platform %q", kolaPlatform)
 	}
 
+	ok = false
```
ajeddeloh: I'd break this out into a function like `validateDistro` just to keep this function as clean as possible and hide the ugly golang away

kola/harness.go
```diff
@@ -201,6 +201,24 @@ func filterTests(tests map[string]*register.Test, pattern, platform string, vers
 			continue
 		}
 
+		allowed = true
```
ajeddeloh: since this is pretty much the same as the platform code, we could generalize and break out into a new function

kola/harness.go
arithx: Would you prefer if I made this:
```golang
if distro == "cl" {
    ...
}
```

I think in the end we'll end up with this kind of indentation once we start handling other version parsing.

ajeddeloh: Actually, what about breaking this into `getCLVersion` and `getRHCOSVersion` and use a switch. Would make adding FCOS later easier as well.

arithx:
Pushed updates (apologies the diff now looks like garbage)

kola/harness.go
```diff
@@ -358,7 +358,20 @@ func getClusterSemver(pltfrm, outputDir string) (*semver.Version, error) {
 		return nil, fmt.Errorf("parsing /etc/os-release: %v: %s", err, stderr)
 	}
 
-	version, err := semver.NewVersion(strings.Split(string(out), "=")[1])
+	// TODO: add distro specific version handling
+	switch distro {
+	case "cl":
+		return parseCLVersion(out)
+	case "rhcos":
```
ajeddeloh: I think these should be seperate cases. `default` should error and rhcos should return the empty semver

arithx: Dropping the `default` case and changing the end of function return to:

```golang
return nil, fmt.Errorf("no case to handle distribution %q", distro)
```

ashcrow:
Is this ready to be re-reviewed and possibly merged?

platform/platform.go
```diff
@@ -111,7 +111,8 @@ type Options struct {
 
 // RuntimeConfig contains cluster-specific configuration.
 type RuntimeConfig struct {
-	OutputDir string
+	OutputDir    string
+	Distribution string
```
bgilbert: By the existing schema, this should be in `Options` instead of `RuntimeConfig` because it doesn't vary between clusters in a run.  Is there a reason it should be here instead?

kola/harness.go
```diff
@@ -359,7 +359,19 @@ func getClusterSemver(pltfrm, outputDir string) (*semver.Version, error) {
...
+		return &semver.Version{}, nil
+	}
+
+	return nil, fmt.Errorf("no case to handle version parsing for distribution %q", distro)
+}
+
+func parseCLVersion(input []byte) (*semver.Version, error) {
+	version, err := semver.NewVersion(strings.Split(string(input), "=")[1])
```
bgilbert: It probably makes more sense to split out the value in the caller.

platform/platform.go
arithx: Ah yeah, `Options` makes sense. I'll switch it.

arithx:
Pushed with changes.

bgilbert:
The flag name in the commit message is stale.

kola/harness.go
```diff
@@ -358,8 +357,21 @@ func getClusterSemver(pltfrm, outputDir string) (*semver.Version, error) {
 	if err != nil {
 		return nil, fmt.Errorf("parsing /etc/os-release: %v: %s", err, stderr)
 	}
+	ver := strings.Split(string(out), "+")[1]
```
bgilbert: This changed from `=` to `+`.  Was that intentional?

arithx: It was not, typo'd.

arithx:
Typo & commit message fixed in latest push.

PR #876, opened 2018-06-28
kola/tests: check for AWS friendly device name

lucab:
Test covering friendly names on AWS, recently fixed on some specific instance types at https://github.com/coreos/bugs/issues/2399. This invariant should not be instance-specific though.

### DIFF
--- a/kola/tests/misc/aws.go
+++ b/kola/tests/misc/aws.go
@@ -1,0 +1,42 @@
+// Copyright 2018 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


ajeddeloh:
Perhaps check if any nvme devices exist first and skip if they don't? Otherwise this test gives false positives on instance types without nvme.

lucab:
@ajeddeloh the idea is that every instance (booted with default parameter) has a friendly disk name for its main volume, regardless of its instance type and nvme/non-nvme. This checks that such invariant holds everywhere, as we just fixed the broken corner-case of nvme.

kola/tests/misc/aws.go
```diff
@@ -0,0 +1,47 @@
...
+	})
+}
+
+// Check invariants on AWS instances.
+
+func awsVerifyDiskFriendlyName(c cluster.TestCluster) {
+	friendlyName := "/dev/xvda"
+	out := c.MustSSH(c.Machines()[0], fmt.Sprintf("stat %s", friendlyName))
+
+	if strings.Contains(string(out), "No such file or directory") {
```
ajeddeloh: Is this necesary since stat should return non-zero on ENOTFOUND

lucab: Oh, does it? I think I overlook `MustSSH` returned value, that would be cool. I'll re-check tomorrow.

lucab: Indeed the string check here was duplicating the internal error checking. I just dropped that.

PR #861, opened 2018-05-30
Add kola runner & platforms READMEs

arithx:
Adds some markdown files which detail how the kola runner works
internally as well as what constitutes a platform in mantle (and some
common things that need to be implemented).

PR #877, opened 2018-07-03
kola: add output messages in run and check-console

rfairley:
This adds a verbose option and a Long description to check-console.
A message to indicate a cluster is being created to check the semver
is also given during kola run, since this can be a long wait time.

**Notes**
These are some minor edits I made while reading the cmd/kola and /kola code. Overall there is not much I felt needed changing after getting better practice in general with the command line - the main takeaway was learning how the tests are run, how command line options are added, and the policy on logging (with https://github.com/coreos/mantle/pull/861 as a guide).

Enhancements that could be made in future are:
- Running count of the tests remaining (possibly in addition to a progress bar which updates based on a rating we give for how long the test takes to run). The data in `t.suite.waiting` and `t.suite.running` could be utilized for this in `tRunner` in `harness/harness.go`. Running tests in parallel would need to be considered.
- Messages to indicate why a test was excluded during `filterTests`. Right now `no tests to run` gets printed, but this could be more specific by saying `incorrect version`, `excluded platform` for example. However, doing this would add a lot of output to the console, by printing out every time a test is excluded.
- `man` pages for the commands which sync with the current usage. Cobra can do this (https://github.com/spf13/cobra/blob/master/doc/man_examples_test.go).

### DIFF
--- a/cmd/kola/console.go
+++ b/cmd/kola/console.go
@@ -24,14 +24,24 @@ import (
 	"github.com/coreos/mantle/kola"
 )
 
-var cmdCheckConsole = &cobra.Command{
-	Use:    "check-console [input-file...]",
-	Run:    runCheckConsole,
-	PreRun: preRun,
-	Short:  "Check console output for badness.",
-}
+var (
+	cmdCheckConsole = &cobra.Command{
+		Use:    "check-console [input-file...]",
+		Run:    runCheckConsole,
+		PreRun: preRun,
+		Short:  "Check console output for badness.",
+		Long: `
+Check console output for expressions matching failure messages logged
+by a Container Linux instance.
+
+If no files are specified as arguments, stdin is checked.
+`}
+
+	checkConsoleVerbose bool
+)
 
 func init() {
+	cmdCheckConsole.Flags().BoolVarP(&checkConsoleVerbose, "verbose", "v", false, "output user input prompts")
 	root.AddCommand(cmdCheckConsole)
 }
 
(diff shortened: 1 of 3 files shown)


kola/harness.go
```diff
@@ -280,6 +280,7 @@ func RunTests(pattern, pltfrm, outputDir string) error {
 	}
 
 	if !skipGetVersion {
+		fmt.Printf("Creating cluster to check semver...\n")
```
ajeddeloh: Should probably use plog

rfairley: Thanks, made the change! With `plog.Printf` it now shows the message at log level INFO which is good.

ajeddeloh: I would just `plog.Info` explicitly to make it clear what log level is intended.

rfairley: Done! Makes sense to use the explicit call instead.

kola/harness.go
```diff
@@ -280,6 +280,7 @@ func RunTests(pattern, pltfrm, outputDir string) error {
 	}
 
 	if !skipGetVersion {
+		plog.Info("Creating cluster to check semver...\n")
```
arithx: I don't think you need the extra line break with `plog.Info`

rfairley: Done, removed the line break and it is still on its own line in the output of `kola run -v`, which is good.

rfairley:
Good to merge. If there is anything else please let me know!

cmd/kola/console.go
```diff
@@ -24,14 +24,25 @@ import (
...
+		Use:    "check-console [input-file...]",
+		Run:    runCheckConsole,
+		PreRun: preRun,
+		Short:  "Check console output for badness.",
+		Long: `
+Check console output for expressions matching failure messages logged
+by a Container Linux instance.
+
+If no files are specified as arguments, stdin is checked.
+
```
arithx: Sorry just noticed this, looks like an extra line break snuck in.

rfairley:
Good to merge.

PR #879, opened 2018-07-06
Add --use-host-dns flag to cork enter

ajeddeloh:
Change the default behavior or cork enter to use the 8.8.8.8 and 8.8.4.4 nameservers instead of copying the hosts's `/etc/resolv.conf`. Add the `--use-host-dns` flag to allow the old behavior.

Also remove the deprecated `---experimental` flag.

### DIFF
--- a/cmd/cork/create.go
+++ b/cmd/cork/create.go
@@ -49,6 +49,7 @@ var (
 
 	// only for `enter` command
 	bindGpgAgent bool
+	useHostDNS   bool
 
 	// only for `update` command
 	allowCreate      bool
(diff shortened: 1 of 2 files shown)


ajeddeloh:
Updated the commit message to both indicate there was a change in default behavior and give a general explanation why the change was made

quickwind:
@ajeddeloh Wouldn't it be necessary to add this flag to also cork create command?

ajeddeloh:
@quickwind no, this creates a new one whenever entering. One created when doing `cork create` would be overwritten by a `cork enter --use-host-dns`.

quickwind:
Ok, but without the flag for cork create, I failed to cork create because we need to use our enterprise DNS server, 8888/8844 doesn't work in our corporate env.

ajeddeloh:
@quickwind Oh, I misunderstood. Do you mean `cork update`? `create` shouldn't enter the sdk at all?

quickwind:
@ajeddeloh I mean both `cork create` and `cork update`, they both will enter the chroot, see following code for `create`:
https://github.com/coreos/mantle/blob/master/cmd/cork/create.go#L221
https://github.com/coreos/mantle/blob/master/cmd/cork/create.go#L246
https://github.com/coreos/mantle/blob/master/sdk/repo.go#L132

Because in our intranet we can't use public DNS server but only company ones, we need to use host DNS setting instead of the public default, otherwise `repo init` will fail (can't resolve address to IP).

ajeddeloh:
Oh, whoops, missed those. Thanks. I'll plumb that flag through and cut a new release.

quickwind:
@ajeddeloh thanks!

ajeddeloh:
Just released 0.11.1 which has the change. You'll still need `--use-host-dns` though.

quickwind:
@ajeddeloh Got that, thanks!

PR #881, opened 2018-07-10
kola/tests: distro gate tests

arithx:
Sets test-level flags for supported distributions.


ajeddeloh:
This is a superset of https://github.com/coreos/mantle/pull/874 right?

arithx:
It depends on #874 so I've included the commit.

PR #882, opened 2018-07-10
kola/tests/ignition: stop writing to the root of the filesystem

arithx:
Switches from writing to a folder on the root of the filesystem to one
in /var to allow the tests to be ran on both CL & RHCOS.


arithx:
This change will require us to change one of the configs stored in the s3 `kola-fixtures` bucket. I have not done that yet as it would break our current tests. An alternative would be to upload another config into s3 (say a `authenticated-var.ign`) and change the test to point to that to not break runs against older versions of mantle.

Thoughts @ajeddeloh @bgilbert

ajeddeloh:
I'm fine with either, breakage would only affect us temporarily during the swap.

PR #883, opened 2018-07-10
kola/tests/misc/nfs: add RHCOS specific NFS test

arithx:
Updates the NFS test to no longer write to use /var/mnt rather than
/mnt, adds the ability to specify what is being mounted, and adds a
RHCOS specific NFSv4 test.

### DIFF
--- a/kola/tests/misc/nfs.go
+++ b/kola/tests/misc/nfs.go
@@ -49,15 +49,23 @@ func init() {
 		Run:         NFSv3,
 		ClusterSize: 0,
 		Name:        "linux.nfs.v3",
+		Distros:     []string{"cl"},
 	})
 	register.Register(&register.Test{
 		Run:         NFSv4,
 		ClusterSize: 0,
 		Name:        "linux.nfs.v4",
+		Distros:     []string{"cl"},
+	})
+	register.Register(&register.Test{
+		Run:         RHCOSNFSv4,
+		ClusterSize: 0,
+		Name:        "rhcos.linux.nfs.v4",
+		Distros:     []string{"rhcos"},
 	})
 }
 
-func testNFS(c cluster.TestCluster, nfsversion int) {
+func testNFS(c cluster.TestCluster, nfsversion int, remotePath string) {
 	m1, err := c.NewMachine(nfsserverconf)
 	if err != nil {
 		c.Fatalf("Cluster.NewMachine: %s", err)
@@ -81,7 +89,7 @@ func testNFS(c cluster.TestCluster, nfsversion int) {
       mode: 0644
 systemd:
   units:
-    - name: "mnt.mount"
+    - name: "var-mnt.mount"
       enabled: true
       contents: |-
         [Unit]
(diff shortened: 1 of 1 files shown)


kola/tests/misc/nfs.go
```diff
@@ -123,15 +131,20 @@ systemd:
...
 }
 
 // Test that NFSv4 without security works on CoreOS.
 func NFSv4(c cluster.TestCluster) {
-	testNFS(c, 4)
+	testNFS(c, 4, "/tmp")
+}
+
+// Test that NFSv4 without security works on RHCOS.
+func NFSv4(c cluster.TestCluster) {
```
miabbott: Travis caught this...should be `RHCOSNFSv4`

arithx: Yup, that's what I get for doing a rework right before the end of the day.

miabbott:
The NFS specific commit looks sane to me.  I'll leave any comments about the distro flag in #874

kola/tests/misc/nfs.go
```diff
@@ -92,13 +100,13 @@ systemd:
         Requires=rpc-statd.service
 
         [Mount]
-        What=%s:/tmp
-        Where=/mnt
+        What=%s:%s
+        Where=/var/mnt
```
cgwalters: Related:  https://github.com/systemd/systemd/pull/6293

kola/tests/misc/nfs.go
```diff
@@ -110,9 +118,9 @@ systemd:
 	c.Log("NFS client booted.")
 
 	checkmount := func() error {
-		status, err := c.SSH(m2, "systemctl is-active mnt.mount")
+		status, err := c.SSH(m2, "systemctl is-active var-mnt.mount")
```
cgwalters: Shouldn't this use `mountLocation`?

arithx: `mountLocation` might be a mis-leading variable name. It's intended to be what you're mounting on the remote system (in the service the `What` for the mount is defined as `<PrivateIP>:<mountLocation>`, the `Where` is always `/var/mnt`). No matter it's value the systemd mount will be `var-mnt.mount`.

kola/tests/misc/nfs.go
```diff
@@ -49,15 +49,23 @@ func init() {
...
+	register.Register(&register.Test{
+		Run:         RHCOSNFSv4,
+		ClusterSize: 0,
+		Name:        "rhcos.linux.nfs.v4",
+		Distros:     []string{"rhcos"},
 	})
 }
 
-func testNFS(c cluster.TestCluster, nfsversion int) {
+func testNFS(c cluster.TestCluster, nfsversion int, mountLocation string) {
```
bgilbert: `remotePath`?

kola/tests/misc/nfs.go
```diff
@@ -123,15 +131,20 @@ systemd:
...
 // Test that the kernel NFS server and client work within CoreOS.
 func NFSv3(c cluster.TestCluster) {
-	testNFS(c, 3)
+	testNFS(c, 3, "/tmp")
 }
 
 // Test that NFSv4 without security works on CoreOS.
 func NFSv4(c cluster.TestCluster) {
-	testNFS(c, 4)
+	testNFS(c, 4, "/tmp")
```
bgilbert: Any reason not to just switch to `/`?

arithx: I'll test again tomorrow morning but if I recall correctly that did not work on CL.

arithx:
Pushed with the variable name change and a fixed commit message.

kola/tests/misc/nfs.go
bgilbert: Nah, I wouldn't spend time on it.

arithx:
Will rebase & merge once #874 lands

PR #886, opened 2018-07-13
platform/machine/qemu/cluster: stop passing backing_fmt to qemu-img

arithx:
qemu-img automatically detects the format of the backing image.


bgilbert:
Needed to support testing QCOW images.

PR #887, opened 2018-07-13
platform/platform: update check in CheckMachine

arithx:
Adds an addijtional valid value for the `ID` check of `/etc/os-release`
(`ID="rhcos"`).


miabbott:
Yes please!  LGTM

PR #888, opened 2018-07-24
kola/tests/coretest: add rhcos.basic

arithx:


arithx:
cc @ashcrow @miabbott

kola/tests/coretest/core.go
```diff
@@ -254,13 +319,23 @@ func TestInstalledUpdateEngineRsaKeys() error {
...
 		"systemd-timesyncd.service",
 		"update-engine.service",
-	}
+	})
+}
+
+func TestServicesActiveRHCOS() error {
+	return servicesActive([]string{
+		"multi-user.target",
+		"docker.service",
```
arithx: This is currently active for now, but not sure if there's a better choice of service.

ashcrow:
Thanks! Will review shortly.

kola/tests/coretest/core.go
```diff
@@ -170,6 +195,46 @@ func TestDockerPing() error {
 	}
 }
 
+func TestCrioEcho() error {
+	//t.Parallel()
```
ashcrow: Can drop.

kola/tests/coretest/core.go
```diff
@@ -170,6 +195,46 @@ func TestDockerPing() error {
 	}
 }
 
+func TestCrioEcho() error {
```
ashcrow: Note: `podman` doesn't test `cri-o`. We will need to get @mrunalp or @runcom to give us proper instructions for using `cri-o` without kube and without dropping down to `runc`.

ashcrow:
Is this meant to supersede https://github.com/coreos/mantle/pull/885?

arithx:
@ashcrow: No, I just ported the `coreos.internet` test and switched to `podman` over `docker` in this set. My idea is to just put this in place now and later on we can go back and extend it as necessary.

kola/tests/coretest/core.go
mrunalp: I suggest using crictl like how we do in our integration tests. For e.g. https://github.com/kubernetes-incubator/cri-o/blob/master/test/network.bats#L16

kola/tests/coretest/core.go
```diff
@@ -98,6 +112,17 @@ func init() {
...
 		},
 	})
+	register.Register(&register.Test{
+		Name:             "rhcos.internet",
+		Run:              InternetTests,
+		ClusterSize:      1,
+		ExcludePlatforms: []string{"qemu"},
+		NativeFuncs: map[string]func() error{
+			"CrioPing": TestCrioPing,
+			"CrioEcho": TestCrioEcho,
```
miabbott: Since we are still sorting out how to properly test `crio`, do you want to just make these `TestPodmanPing` and `TestPodmanEcho`?

We could add `crio` specific tests later.

ashcrow: It wouldn't hurt anything, through I don't think there is a use case for using `podman` in RHCOS (other than us building for testing).

arithx:
Switched to WIP, have it working with `crictl` but still need to look at other ways of simplifying the setup. It's in a hacky state atm.

arithx:
Rebased onto master & dropped the `rhcos.internet` test as the functionality is mostly covered inside #885

ashcrow:
Changes look good. Does it pass locally?

arithx:
@ashcrow it does pass locally, the one outstanding thing I'm still looking for input on is if we want different services being checked (https://github.com/coreos/mantle/pull/888/files#r204900927). Ideally I'd like to switch it to checking `crio.service` instead of `docker.service` but it's still not enabled by default on the most recent AMI.

kola/tests/coretest/core.go
miabbott: With openshift/redhat-release-cores#7, we could switch this to `crio.service`?

ashcrow:
@arithx makes sense. My recommendation would be to move without doing the check on `crio.service` and, instead, enabling a test once it is enabled by default in a follow up PR.

kola/tests/coretest/core.go
arithx: The `rdgo` sync is currently broken so that change isn't propogating to the images.

PR #889, opened 2018-07-27
platform/api/aws/ec2: reduce machine creation flakes

arithx:
Ignores `InvalidInstanceID.NotFound` errors during machine creation. This
most commonly happens when AWS itself is having issues.

### DIFF
--- a/platform/api/aws/ec2.go
+++ b/platform/api/aws/ec2.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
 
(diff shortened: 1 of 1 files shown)


lucab:
LGTM. Maybe you want to log such case before throwing the error away? My understanding is that this should be quite rare so logging this shouldn't clog the output.

platform/api/aws/ec2.go
```diff
@@ -127,6 +128,10 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
 			InstanceIds: aws.StringSlice(ids),
 		})
 		if err != nil {
+			// Keep retrying if the InstanceID disappers momentarily
```
bgilbert: `disappears`

arithx:
Comment typo fixed & `plog` debug log added.

PR #892, opened 2018-07-31
kola/tests/misc/selinux:  adapt for RHCOS

miabbott:
This adapts the existing `coreos.selinux.enforce` test to work on
RHCOS and retain its functionality on CL.

Three other tests have been added to validate changing SELinux
booleans and file contexts, as some additional sanity type tests
around SELinux:

  - `coreos.selinux.boolean`
  - `coreos.selinux.boolean.persist`
  - `coreos.selinux.manage`

Signed-off-by: Micah Abbott <miabbott@redhat.com>

### DIFF
--- a/kola/tests/misc/selinux.go
+++ b/kola/tests/misc/selinux.go
@@ -15,8 +15,11 @@
 package misc
 
 import (
+	"regexp"
+
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform"
 )
 
 func init() {
@@ -24,31 +27,45 @@ func init() {
 		Run:         SelinuxEnforce,
 		ClusterSize: 1,
 		Name:        "coreos.selinux.enforce",
-		Flags:       []register.Flag{register.NoEnableSelinux},
 		Distros:     []string{"cl", "rhcos"},
+	})
+	register.Register(&register.Test{
+		Run:         SelinuxBoolean,
+		ClusterSize: 1,
+		Name:        "coreos.selinux.boolean",
+		Distros:     []string{"cl", "rhcos"},
+	})
+	register.Register(&register.Test{
+		Run:         SelinuxBooleanPersist,
+		ClusterSize: 1,
+		Name:        "rhcos.selinux.boolean.persist",
+		Distros:     []string{"rhcos"},
+	})
+	register.Register(&register.Test{
+		Run:         SelinuxManage,
+		ClusterSize: 1,
+		Name:        "rhcos.selinux.manage",
+		Distros:     []string{"rhcos"},
 	})
 }
 
-// SelinuxEnforce checks that some basic things work after `setenforce 1`
-func SelinuxEnforce(c cluster.TestCluster) {
-	m := c.Machines()[0]
+type cmdCheckOutput struct {
+	cmdline     string // command to run
+	checkoutput bool   // should output be checked
+	match       string // regex used to match output from command
+}
 
-	for _, cmd := range []struct {
-		cmdline     string
-		checkoutput bool
-		output      string
-	}{
-		{"getenforce", true, "Permissive"},
-		{"sudo setenforce 1", false, ""},
-		{"getenforce", true, "Enforcing"},
-		{"systemctl --no-pager is-active system.slice", true, "active"},
-		{"sudo cp --remove-destination $(readlink -f /etc/selinux/config) /etc/selinux/config", false, ""},
-		{"sudo sed -i 's/SELINUX=permissive/SELINUX=enforcing/' /etc/selinux/config", false, ""},
-	} {
+// testSelinuxCmds will run a list of commands, optionally check their output, and
+// ultimately reboot the host
+func testSelinuxCmds(c cluster.TestCluster, m platform.Machine, cmds []cmdCheckOutput) {
+	for _, cmd := range cmds {
 		output := c.MustSSH(m, cmd.cmdline)
 
-		if cmd.checkoutput && string(output) != cmd.output {
-			c.Fatalf("command %q has unexpected output: want %q got %q", cmd.cmdline, cmd.output, string(output))
+		if cmd.checkoutput {
+			match := regexp.MustCompile(cmd.match).MatchString(string(output))
+			if !match {
+				c.Fatalf("command %q has unexpected output: tried to match regexp %q, output was %q", cmd.cmdline, cmd.match, string(output))
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


miabbott:
This requires coreos/ignition#569 to get SELinux working correctly on RHCOS.

I opted not to enable `coreos.selinux.boolean.persist` and `coreos.selinux.manage` for CL because it would require additional setup that was specific to CL.

kola/tests/misc/selinux.go
```diff
@@ -56,13 +72,90 @@ func SelinuxEnforce(c cluster.TestCluster) {
...
+	if string(output) != "virt_use_nfs --> on" {
+		c.Fatalf("The SELinux boolean \"virt_use_nfs\" is incorrectly configured: want \"virt_use_nfs --> on\", got %q", string(output))
+	}
+}
+
+// SelinuxManage checks that you can modify an SELinux file context and
+// have it persist across reboots
+func SelinuxManage(c cluster.TestCluster) {
+	cmdList := []cmdCheckOutput{
+		{"sudo /usr/sbin/semanage fcontext -l | grep vasd", true, "/var/opt/quest/vas/vasd(/.*)?                      all files          system_u:object_r:var_auth_t:s0"},
```
arithx: I'm wondering whether or not it might be worth it to switch the output validation string to a regexp rather than doing a direct string comparison, thoughts?

kola/tests/misc/selinux.go
```diff
@@ -56,13 +72,90 @@ func SelinuxEnforce(c cluster.TestCluster) {
 	if err != nil {
 		c.Fatalf("failed to reboot machine: %v", err)
 	}
+}
+
+// SelinuxEnforce checks that some basic things work after `setenforce 1`
+func SelinuxEnforce(c cluster.TestCluster) {
+	cmdList := []cmdCheckOutput{
+		{"/usr/sbin/getenforce", true, "Permissive"},
```
arithx: Would it be worthwhile to track down the `$PATH` issue on RHCOS so we don't have to use the entire path for binaries in `/usr/sbin`?

kola/tests/misc/selinux.go
miabbott: Yeah, that makes sense given how long the output is here.  I'll investigate.

kola/tests/misc/selinux.go
miabbott: Agreed, I'll keep digging into that over in openshift/os#191.  We should be able to leave this as is until that gets sorted out, right?

arithx: I personally think it's fine to carry it like this temporarily; maybe we should just either add a comment (either in code or in the openshift/os issue) to go back and drop them after it's resolved.

miabbott: Smart.  I'll add that in.

kola/tests/misc/selinux.go
```diff
@@ -27,42 +29,148 @@ func init() {
...
 		c.Fatalf("failed to reboot machine: %v", err)
 	}
+}
+
+// SelinuxEnforce checks that some basic things work after `setenforce 1`
+func SelinuxEnforce(c cluster.TestCluster) {
+	// we use the full path to most commands here because the $PATH is
+	// weird in RHCOS.  See https://github.com/openshift/os/issues/191
+	cmdList := []cmdCheckOutput{
+		{"/usr/sbin/getenforce", true, "Permissive"},
```
arithx: Rather than changing the paths in the `SelinuxEnforce` function which is used by CL can we create an RHCOS variant that uses the full paths (like this) and then just drop the function & move back to the `SelinuxEnforce` method if/once the `PATH` variable is fixed on RHCOS?

kola/tests/misc/selinux.go
```diff
@@ -27,42 +29,148 @@ func init() {
...
 
 	err := m.Reboot()
 	if err != nil {
 		c.Fatalf("failed to reboot machine: %v", err)
 	}
+}
+
+// SelinuxEnforce checks that some basic things work after `setenforce 1`
+func SelinuxEnforce(c cluster.TestCluster) {
+	// we use the full path to most commands here because the $PATH is
```
arithx: Let's specifically add a `TODO:` at the front of this

kola/tests/misc/selinux.go
miabbott: Yeah, I can do this.

It gets into a larger question of how much customization do we want to do to support RHCOS?  I was worried about cluttering up the code base with custom functions/tests that run only on RHCOS.  But if you think that would be OK, I'll start using that approach in other places.

arithx: This is all from my perspective, @ajeddeloh or @bgilbert might have a differing one, but I'd prefer us to have the variant functions (even for single distributions). Especially in cases like this where it's very possible that we fix the underlying `PATH` issue and can then switch back to the existing test.

kola/tests/misc/selinux.go
```diff
@@ -25,44 +27,188 @@ func init() {
...
 		Flags:       []register.Flag{register.NoEnableSelinux},
+		Distros:     []string{"cl"},
+	})
+
+	// TODO:  fall back to SelinuxEnforce when PATH on RHCOS is sane
+	// See https://github.com/openshift/os/issues/191
+	register.Register(&register.Test{
+		Run:         SelinuxEnforceRhcos,
+		ClusterSize: 1,
+		Name:        "coreos.selinux.enforce.rhcos",
```
arithx: The pattern we've been using is to replace the `coreos` with `rhcos`

miabbott: Noted.

This might be moot very soon, now that we have SELinux support enabled in RHCOS and the PATH should be sane.

I'll retry this test once we have a new cloud image built with the SELinux support + PATH changes intact.

ashcrow: selinux support is now enabled in the latest composes.

miabbott:
With the fixes for `PATH` in place and SELinux enabled on RHCOS, I was able to drop a lot of the workarounds for the RHCOS case.

I renamed the RHCOS-only tests with the `rhcos.` prefix.

Additionally, I reconfigured the `coreos.selinux.enforce` to not disable SELinux when the machine is started.

kola/tests/misc/selinux.go
```diff
@@ -17,52 +17,150 @@ package misc
 import (
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform"
+	"regexp"
```
arithx: We structure imports in the following order (with an extra line break between categories): stdlib, external, internal

```golang
import (
    "regexp"

    "github.com/coreos/mantle/kola/cluster"
    "github.com/coreos/mantle/kola/register"
)
```

kola/tests/misc/selinux.go
```diff
@@ -17,52 +17,150 @@ package misc
...
 
 func init() {
 	register.Register(&register.Test{
 		Run:         SelinuxEnforce,
 		ClusterSize: 1,
 		Name:        "coreos.selinux.enforce",
-		Flags:       []register.Flag{register.NoEnableSelinux},
 		Distros:     []string{"cl", "rhcos"},
 	})
+
```
arithx: nit: extra line breaks between test registrations

kola/tests/misc/selinux.go
```diff
@@ -17,52 +17,150 @@ package misc
...
 
-	o := c.MustSSH(m, "which getenforce")
-	c.Logf(string(o))
+	testSelinuxCmds(c, m, cmdList)
 
 	output := c.MustSSH(m, "getenforce")
 
 	if string(output) != "Enforcing" {
-		c.Fatalf("command \"getenforce\" has unexpected output: want \"Enforcing\" got %q", string(output))
+		c.Fatalf("command \"getenforce\" has unexpected output: want \"Enforcing\", got %q", string(output))
```
ashcrow: nit: If you want to avoid the escaping you can write this like:

```golang
c.Fatalf(`command "getenforce" has unexpected output: want "Enforcing", got %q`, string(output))
```

arithx: or just do another `%q` & add `"Enforcing"` to the parameters

ashcrow:
Will defer to @arithx for final review.

miabbott:
Thanks for the feedback; addressed the latest nits :arrow_up:

miabbott:
good to merge

PR #893, opened 2018-08-01
platform/api/aws: add AmazonS3ReadOnlyAccess to Role

arithx:
When running kola on a root account that does not own the S3 objects
that are being fetched for tests the AmazonS3ReadOnlyAccess policy needs
to be added to the role (and the corresponding Instance Profile).

This also adds a conditional retry on machine creation to avoid errors
caused by the eventual consistency nature of AWS in relation to the
Instance Profile creation and subsequent use in machine creation.

### DIFF
--- a/platform/api/aws/ec2.go
+++ b/platform/api/aws/ec2.go
@@ -17,9 +17,11 @@ package aws
 import (
 	"encoding/base64"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
 
 	"github.com/coreos/mantle/util"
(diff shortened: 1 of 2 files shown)


platform/api/aws/ec2.go
```diff
@@ -105,7 +107,19 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
 		},
 	}
 
-	reservations, err := a.ec2.RunInstances(&inst)
+	var reservations *ec2.Reservation
+	err = util.RetryConditional(5, 5*time.Second, func(err error) bool {
```
ajeddeloh: Is this going to end up calling the other util.WaitUntilReady in a loop? If so that'll explode the time this could wait for.

arithx: No, the check function is just a validation of the error & the run function is a direct call to AWS.

PR #894, opened 2018-08-02
platform: perform CheckConsole against journal on all platforms

arithx:
Performs the CheckConsole function against journalctl output on all
platforms. This extends the Cluster & Machine interfaces with a new
function JournalOutput.

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -409,6 +409,11 @@ func runTest(h *harness.H, t *register.Test, pltfrm string) {
 				h.Errorf("Found %s on machine %s console", badness, id)
 			}
 		}
+		for id, output := range c.JournalOutput() {
+			for _, badness := range CheckConsole([]byte(output), t) {
+				h.Errorf("Found %s on machine %s journal", badness, id)
+			}
+		}
 	}()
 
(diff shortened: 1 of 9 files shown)


arithx:
@ajeddeloh I did a test run on each platform of `coreos.ignition.*` and nothing got flagged. This shouldn't really trigger much yet, but will allow us in the future to add checks specific to the journal.

kola/harness.go
```diff
@@ -409,6 +409,11 @@ func runTest(h *harness.H, t *register.Test, pltfrm string) {
 				h.Errorf("Found %s on machine %s console", badness, id)
 			}
 		}
+		for id, output := range c.JournalOutput() {
+			for _, badness := range CheckConsole([]byte(output), t) {
+				h.Errorf("Found %s on machine %s journal", badness, id)
```
bgilbert: For messages that go both places, this will produce redundant log entries.  That seems okay as a starting point, though we might want to split out some checks later.

arithx: Yeah, in the future I'd like to switch from using the `CheckConsole` function for the journal checks to a more specific set.

PR #895, opened 2018-08-03
kola/tests: disable systemd.* tests on RHCOS

arithx:
These tests shouldn't have been enabled on RHCOS initially as the
binaries required for them are not shipped.

### DIFF
--- a/kola/tests/systemd/journald.go
+++ b/kola/tests/systemd/journald.go
@@ -58,7 +58,7 @@ func init() {
 		Run:         journalRemote,
 		ClusterSize: 0,
 		Name:        "systemd.journal.remote",
-		Distros:     []string{"cl", "rhcos"},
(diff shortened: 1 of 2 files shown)


cgwalters:
This is OK, but...I feel like we should start doing at least *some* feature detection too.  Fedora CoreOS will definitely have sysusers for example.

bgilbert:
@cgwalters IMO we want the decisions on what to test to be encoded in the test case, not detected from the system, since otherwise we're trusting the system under test to give us correct answers.

cgwalters:
It's certainly got the issue that e.g. if somehow CL's `/usr/bin/docker` binary went missing and our tests keyed of that they'd just silently skip the tests.

We'll see though just how painful it is to maintain these tags when we get up to 3 distributions (at least).

PR #896, opened 2018-08-06
kola/tests/ignition: adapt for RHCOS

mike-nguyen:
The current ignition user tests were failing on RHCOS because RHCOS
does not have a docker group by default like CL.  There was a
subsequent failure due to the UID difference for the core user and
creation of new users.

The original tests were modified to only run on CL.  New tests were
added for RHCOS to account for no docker group and UID difference.

### DIFF
--- a/kola/tests/ignition/passwd.go
+++ b/kola/tests/ignition/passwd.go
@@ -113,7 +113,7 @@ func init() {
 		               ]
 		             }
 		           }`),
-		Distros: []string{"cl", "rhcos"},
+		Distros: []string{"cl"},
 	})
 	register.Register(&register.Test{
 		Name:        "coreos.ignition.v2.users",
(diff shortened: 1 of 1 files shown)


kola/tests/ignition/passwd.go
```diff
@@ -147,18 +147,94 @@ func init() {
...
+	})
+	register.Register(&register.Test{
+		Name:        "rhcos.ignition.v1.users",
+		Run:         usersRhcos,
+		ClusterSize: 1,
+		UserData: conf.Ignition(`{
+		             "ignitionVersion": 1,
+		             "systemd": {
+		               "units": [{
+		                 "name": "system-cloudinit@usr-share-coreos-developer_data.service",
```
arithx: Does this service actually exist on RHCOS machines?

mike-nguyen: No it doesn't exist in RHCOS.

I didn't find it in CL either:
`[bound] core@ip-172-31-21-214 ~ $ systemctl list-units | grep system-cloudinit`

Should it be dropped from both RHCOS and CL?  I wasn't sure if it was a drop in for something I wasn't aware of so I left it in.

dm0-: Yes, developer images are gone, but that password can still be set in prod images with `set_shared_user_password.sh`.  (It can actually take any cloud-config file, but that password script is the normal use for it.)

mike-nguyen: I'll remove it for RHCOS and leave it as is for CL.

mike-nguyen:
Pushed a new commit with the service removed for RHCOS

mike-nguyen:
Good to merge

PR #900, opened 2018-08-15
kola/tests: disable linux.ntp on RHCOS

arithx:


arithx:
@dm0-: the test is using `networkctl` to determine the status of `eth0`

PR #903, opened 2018-08-21
kola/tests/etcd: Run the RHCOS container unconfined

### LINKED ISSUE #902: Kola: rhcos.etcd.cluster.tls test fails due to SELinux error
Running the test through the latest AWS RHCOS instance gives the following error message:

`rhcos.go:178: discovery failed cluster-health check: health polling failed: Process exited with status 125:`

The cause appears to be:

`kernel: type=1400 audit(1534865272.545:4): avc:  denied  { read } for  pid=1679 comm="etcd" name="peer-cert.pem" dev="dm-0" ino=6497731 scontext=system_u:system_r:container_t:s0:c596,c637 tcontext=unconfined_u:object_r:cert_t:s0 tclass=file`

As I'm not very familiar with this test, it would be much appreciated if someone could point out the best fix for this. (FWIW, the test passes if I setenforce 0 so the SELinux issue appears to be the only failure)

dm0-:
The `container_t` process can't access the `cert_t` files from the mounted volume, and podman's :Z and :z options have no effect.

Fixes #902


cgwalters:
> podman's :Z and :z options have no effect.

Well, you wouldn't want to do that anyways; it'd break the labels for non-container processes accessing them.  What flatpak is doing is [pulling the host certs automatically](https://github.com/flatpak/flatpak/issues/677#issuecomment-301079267) - this is one of those flatpak things that'd be nice to have supported in podman too.

PR #908, opened 2018-09-04
kola/tests/misc: Switch CL to the RHCOS NFS test

dm0-:
Part of coreos/coreos-overlay#3348


dm0-:
I've added version-gating since this is an incompatible change.  It will need to be updated depending on when it's merged.

dm0-:
I've rebased this and dropped all the distro/version complexity.  It now specifies `Type=nfs4` and works everywhere.

PR #910, opened 2018-09-11
kola/tests: add kubernetes static pod test

mike-nguyen:
This is a basic smoke test to ensure a node can handle static pods on
RHCOS.  The test configures the static pod through ignition (drops in
kubelet configuration, static pod definition, and kubelet unit file).


kola/tests/kubernetes/kubelet.go
```diff
@@ -0,0 +1,103 @@
...
+                    "source": "data:,kind%3A%20KubeletConfiguration%0AapiVersion%3A%20kubelet.config.k8s.io%2Fv1beta1%0AclusterDomain%3A%20kube.local%0AclusterDNS%3A%20%5B10.254.0.10%5D%0AcgroupDriver%3A%20systemd%0AstaticPodPath%3A%20%2Fetc%2Fkubernetes%2Fmanifests%2F%0Aauthorization%3A%0A%20%20mode%3A%20AlwaysAllow%0Aauthentication%3A%0A%20%20anonymous%3A%0A%20%20%20%20enabled%3A%20true%0A%20%20webhook%3A%0A%20%20%20%20enabled%3A%20false%0A",
+                    "verification": {}
+                },
+                "filesystem": "root",
+                "mode": 420,
+                "path": "/etc/kubernetes/kubeletconfig"
+            },
+            {
+                "contents": {
+                    "source": "data:,apiVersion%3A%20v1%0Akind%3A%20Pod%0Ametadata%3A%0A%20%20name%3A%20static-web%0A%20%20labels%3A%0A%20%20%20%20role%3A%20myrole%0Aspec%3A%0A%20%20containers%3A%0A%20%20%20%20-%20name%3A%20web%0A%20%20%20%20%20%20image%3A%20nginx%0A%20%20%20%20%20%20ports%3A%0A%20%20%20%20%20%20%20%20-%20name%3A%20web%0A%20%20%20%20%20%20%20%20%20%20containerPort%3A%2080%0A%20%20%20%20%20%20%20%20%20%20protocol%3A%20TCP%0A",
```
cgwalters: How hard would it be to have that be literal content so it's easy to view/edit?

kola/tests/kubernetes/kubelet.go
```diff
@@ -0,0 +1,103 @@
...
+                "filesystem": "root",
+                "mode": 420,
+                "path": "/etc/kubernetes/manifests/static-pod.yaml"
+            }
+        ]
+    },
+    "systemd": {
+        "units": [
+            {
+                "contents": "[Unit]\nDescription=Kubernetes Kubelet\nRequires=docker.service\nAfter=docker.service\n\n[Service]\nExecStart=/usr/bin/hyperkube \\\n    kubelet \\\n        --config /etc/kubernetes/kubeletconfig\n\nRestart=always\nRestartSec=10\n\n[Install]\nWantedBy=multi-user.target\n",
```
cgwalters: Just a note, this and below assume kubelet is using docker (and that the host has docker at all) - so it'll need some adaption for Fedora CoreOS at least, and shortly RHCOS isn't going to use Docker either.  But for now we can do this, just let's keep in mind it'll (hopefully shortly) need to change.

kola/tests/kubernetes/kubelet.go
```diff
@@ -0,0 +1,103 @@
...
+        "timeouts": {},
+        "version": "2.2.0"
+    },
+    "networkd": {},
+    "passwd": {},
+    "storage": {
+        "files": [
+            {
+                "contents": {
+                    "source": "data:,kind%3A%20KubeletConfiguration%0AapiVersion%3A%20kubelet.config.k8s.io%2Fv1beta1%0AclusterDomain%3A%20kube.local%0AclusterDNS%3A%20%5B10.254.0.10%5D%0AcgroupDriver%3A%20systemd%0AstaticPodPath%3A%20%2Fetc%2Fkubernetes%2Fmanifests%2F%0Aauthorization%3A%0A%20%20mode%3A%20AlwaysAllow%0Aauthentication%3A%0A%20%20anonymous%3A%0A%20%20%20%20enabled%3A%20true%0A%20%20webhook%3A%0A%20%20%20%20enabled%3A%20false%0A",
```
eparis: Can you translate these into english for review at least?  @sjennings should review this, to make sure it looks like the bootstrap config we are going to use.

kola/tests/kubernetes/kubelet.go
arithx: While this is typically how we have things in kola tests, I'd be okay if we were to do something like this:
```golang
import (
    "github.com/vincent-petithory/dataurl"
)

var (
    staticPodConfig = conf.Ignition(fmt.Sprintf(`{
    ...
    "contents": {
        "source": "data:,%s",
    },
    ...
}`, dataurl.EncodeBytes([]byte("<kubelet_contents_in_plain_text>")), ...)
)
```

kola/tests/kubernetes/kubelet.go
mike-nguyen:
```
kind: KubeletConfiguration
apiVersion: kubelet.config.k8s.io/v1beta1
clusterDomain: kube.local
clusterDNS: [10.254.0.10]
cgroupDriver: systemd
staticPodPath: /etc/kubernetes/manifests/
authorization:
  mode: AlwaysAllow
authentication:
  anonymous:
    enabled: true
  webhook:
    enabled: false
```

I noticed https://github.com/openshift/installer/blob/master/modules/ignition/resources/services/kubelet.service but kubelet in RHCOS is running 1.11 which uses a a kubeletconfiguration file for most of the options.  If there is already an existing kubeletconfiguration, let me know so I can use that instead of the one I created.

sjenning:
```
kind: KubeletConfiguration
apiVersion: kubelet.config.k8s.io/v1beta1
clusterDomain: kube.local
clusterDNS: [10.254.0.10]
cgroupDriver: systemd
staticPodPath: /etc/kubernetes/manifests/
authorization:
  mode: AlwaysAllow
authentication:
  anonymous:
    enabled: true
  webhook:
    enabled: false
```
currently our manifest dir is `/etc/origin/node/pods`.  might change that to match.

eparis: if 4.0 is switching to a kubelet, and the isntaller is going to use /etc/kubernetes/manifests like upstream, I think we should take this change to just abandon /etc/origin/node/pods....

mike-nguyen:
Pushed some new changes:
- Readable YAML / Unit  / Ignition files 
- Changed the validation to use curl instead of docker to check the nginx static pod is running.  I still could not remove docker from the unit file but we can cross that bridge later.

kola/tests/kubernetes/kubelet.go
mike-nguyen: I'll keep it as is if that is the destination going forward.  Is there anything else that sticks out in terms of the kubeletconfiguration?  I created it based on the API with the least amount of options to get it working.

kola/tests/kubernetes/kubelet.go
mike-nguyen: I reduced the verification down to a curl command that doesn't depend on the the container runtime.  I left docker in the kubelet unit file though.

mike-nguyen:
Docker was dropped from RHCOS so I'll be re-working the test to run on crio.

mike-nguyen:
Switched over to using CRI-O / runc but will need cni fix/workaround [0] for the port forward to work.

[0] https://github.com/kubernetes-sigs/cri-o/issues/1804

PR #915, opened 2018-09-14
glide: bump etcd and unpin a lot of packages

ajeddeloh:
Hopefully this doesn't break anything, I'm beginning to untangle the web that is mantle's glide.* files.


ajeddeloh:
To some degree. unit tests pass (duh, travis checks that), and it didn't break anything obvious like kola spawn. But we don't have any sort of comprehensive testing methodology for mantle.

arithx:
@bgilbert might have some extra input but covering the platforms on `kola run` is probably as sufficient as we can get outside of plume stuff.

ajeddeloh:
Tested a full kola run with:
* :+1: qemu
* :+1: aws
* :+1: gce
* :-1: do, but just because DO started handing out 5XX errors
* :question:  packet (still running)

cgwalters:
Looking at `git log ./glide.lock` I don't see any log messages that are like "pin azure sdk since it was broken" or whatever.  So this looks fine to me.

ajeddeloh:
We normally pin thing in the `glide.yaml` instead of the `.lock`, but the only thing I found looking through that was something for Oracle cloud which we no longer support.

ajeddeloh:
Looks like kola still works. Regardless this is doing some much needed updating we'd need to do eventually anyway, so if we hit issues later we should probably fix our code instead of carrying ancient deps.

glide.yaml
```diff
@@ -90,9 +54,5 @@ import:
   version: 045497edb6234273d67dbc25da3f2ddbc4c4cacf
 - package: google.golang.org/api
   version: c858ef4400610cbfd097ffc5f5c6e4a1a51eac86
-- package: google.golang.org/appengine
-  version: 12d5545dc1cfa6047a286d5e853841b6471f4c19
-- package: google.golang.org/cloud
```
bgilbert: We depend pretty heavily on this API, so I'm a little hesitant to unpin it.  Thoughts?

ajeddeloh: We don't directly consume these APIs. We only use the `google.golang.org/api` one; they're transient deps. I'm relying on glide doing the right thing here. They also don't change in glide.lock.

Issue #911, opened 2018-09-11
running upstream project tests via kola?


miabbott:
The `ostree` and `rpm-ostree` projects have a robust set of tests that could be used to validate functionality on RHCOS/FCOS.  Specifically, the [vmcheck](https://github.com/projectatomic/rpm-ostree/tree/master/tests/vmcheck) tests of `rpm-ostree` are such that they can be run against a VM which has an `rpm-ostree` daemon running.

While experimenting running these tests against RHCOS (without using `kola`), I ran into a few snags that made me question how the integration of the tests into `kola` might work.

1.  The upstream tests can change with each release/commit, so `kola` would have to perform a checkout of the repo which matches the version of software under test.  I didn't find any other examples of this kind of approach in the current codebase.  Obviously, this doesn't mean we can't adapt, but I wanted to get a feel if this kind of approach would work out.

2.  I found the `vmcheck` suite has a decent amount of dependencies that need to be installed on the host running the test.  `kola` doesn't really have this problem, as it is just a golang binary after it is all compiled etc.  Would this requirement on certain packages get pushed up to a higher layer to handle?

3.  The alternative to installing packages on the host running the test would be to pull a container with the required dependencies and execute the tests via the container, via `kola`.  I'm not sure this is a better solution than what would be required for point 2.

4.  The extreme alternative is to punt on the upstream tests and re-implement some of the tests natively in 'kola`.

Maybe I'm getting ahead of myself with these kinds of questions, but I figured the folks more experienced with this codebase would be able to steer me in the right direction.

miabbott:
Oh I forgot to add...I had short discussions with @jlebon about finding a generic way for running these kinds of upstream tests via `kola`.  I'm not convinced about the usefulness of such an effort, since there don't seem to be a lot of projects that we work with that execute their tests the same way (i.e. [Fedora STI](https://fedoraproject.org/wiki/CI/Standard_Test_Interface)).

arithx:
Do we want/need these to be added to `kola`? I'm not familiar with `vmcheck`, if it's meant more to test `rpm-ostree` it might make sense to just add a subset of native tests that target core functionality to make sure that what we have packaged in the image is functional rather than testing all of `rpm-ostree`.

-----

Here's some miscellaneous thoughts I've had on the subject of external tests in `kola` that don't necessarily directly target any of your points:

Ideally I'd like to avoid adding more vendor bloat to `mantle` for additional tests unless directly necessary. 

[...]

PR #916, opened 2018-09-17
kola/tests: add rpmostree package

miabbott:
Adds tests for the following `rpm-ostree` commands:
  - status
  - upgrade
  - rollback
  - cleanup
  - install
  - uninstall

The `install-uninstall` test requires network access to fetch an RPM
from EPEL.

Signed-off-by: Micah Abbott <miabbott@redhat.com>

### DIFF
--- a/kola/tests/rpmostree/deployments.go
+++ b/kola/tests/rpmostree/deployments.go
@@ -1,0 +1,302 @@
+// Copyright 2018 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rpmostree
+
+import (
+	"reflect"
+	"regexp"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform/conf"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         rpmOstreeUpgradeRollback,
+		ClusterSize: 1,
+		Name:        "rhcos.rpmostree.upgrade-rollback",
+		Distros:     []string{"rhcos"},
+	})
+	register.Register(&register.Test{
+		Run:         rpmOstreeInstallUninstall,
+		ClusterSize: 1,
+		Name:        "rhcos.rpmostree.install-uninstall",
+		// this Ignition config lands the EPEL repo + key
+		UserData: conf.Ignition(`{
+  "ignition": {
+    "version": "2.2.0"
+  },
+  "storage": {
+    "files": [
+      {
+        "filesystem": "root",
+        "group": {
+          "name": "root"
+        },
+        "path": "/etc/yum.repos.d/epel.repo",
+        "user": {
+          "name": "root"
+        },
+        "contents": {
+          "source": "data:,%5Bepel%5D%0Aname%3DExtra%20Packages%20for%20Enterprise%20Linux%207%20-%20%24basearch%0Ametalink%3Dhttps%3A%2F%2Fmirrors.fedoraproject.org%2Fmetalink%3Frepo%3Depel-7%26arch%3D%24basearch%0Afailovermethod%3Dpriority%0Aenabled%3D1%0Agpgcheck%3D1%0Agpgkey%3Dfile%3A%2F%2F%2Fetc%2Fpki%2Frpm-gpg%2FRPM-GPG-KEY-EPEL-7%0A"
+        },
+        "mode": 420
+      },
+      {
+        "filesystem": "root",
+        "group": {
+          "name": "root"
+        },
+        "path": "/etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-7",
+        "user": {
+          "name": "root"
+        },
+        "contents": {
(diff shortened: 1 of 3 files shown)


miabbott:
I opted to write the tests natively, rather than trying to reuse the `vmcheck` test suite as discussed in #911.  It was faster to do it this way, in my opinion.

Based on the feedback in #911, it seems like we should still explore reusing the upstream tests in some fashion, but this fills a gap in the meantime.

kola/tests/rpmostree/status.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Regex to match the likes of "rhcos:openshift/3.10/x86_64/os"
+//
+// Groups:
+// 0 - rhcos
+// 1 - openshift
+// 2 - 3.10
+// 3 - x86_64
+// 4 - os
+var rhcosOriginRegex string = `^(\w+):(\w+)/(\d\.\d+)/(\w+)/(\w+)`
```
cgwalters: Let's please not hardcode this - not just because we'd like these tests to work on FCOS too, but also because doing it here will make it *extremely* painful to change - and I'd like to change RHCOS to use a [custom origin](https://github.com/ashcrow/pivot/pull/12) for the initial state too.

If this landed as is, then in order to change RHCOS we'd have to have a patch ready for both this repo and the config, then build an assembler container containing this change, which we'd then have to carefully only use for the *new* OS version.

Basically if one has separate git repos that interlock on each other, changes become extremely hard.

kola/tests/rpmostree/deployments.go
```diff
@@ -0,0 +1,300 @@
...
+
+	originalStatus, err := getRpmOstreeStatusJSON(c, m)
+	if err != nil {
+		c.Fatal(err)
+	}
+
+	// create a local branch to act as our upgrade target
+	originalOrigin := originalStatus.Deployments[0].Origin
+	originalCsum := originalStatus.Deployments[0].Checksum
+	createBranch := "sudo ostree refs --create local-branch " + originalCsum
```
cgwalters: This isn't a problem in *this* code but something to watch out for is shell quoting.  I'm not sure if kola or Go have an equivalent of [this API](https://developer.gnome.org/glib/stable/glib-Shell-related-Utilities.html#g-shell-quote).  Basically if some value can e.g. have spaces or shell metacharacters itself.

kola/tests/rpmostree/deployments.go
```diff
@@ -0,0 +1,300 @@
...
+		c.Fatal(err)
+	}
+
+	// should have two deployments
+	if len(postUpgradeStatus.Deployments) != 2 {
+		c.Fatalf("Expected two deployments; found %d deployments", len(postUpgradeStatus.Deployments))
+	}
+
+	// reboot into new deployment
+	rebootErr := m.Reboot()
```
cgwalters: Hooray a reboot API!

kola/tests/rpmostree/deployments.go
```diff
@@ -0,0 +1,300 @@
...
+
+	// version should be new version string
+	if postRebootStatus.Deployments[0].Version != "kola-test-1.0" {
+		c.Fatalf(`New deployment version is incorrect; expected %q, got %q`, "kola-test-1.0", postRebootStatus.Deployments[0].Checksum)
+	}
+
+	// fix original origin file to point to proper ref
+	originalOriginFile := "/ostree/deploy/rhcos/deploy/" + originalCsum + ".0.origin"
+	newSedCmd := "sudo sed -i 's|^refspec=.*$|refspec=" + originalOrigin + "|' " + originalOriginFile
+	c.MustSSH(m, newSedCmd)
```
cgwalters: And then this wouldn't be necessary if we rebased initially right?

kola/tests/rpmostree/deployments.go
```diff
@@ -0,0 +1,300 @@
...
+func rpmOstreeUpgradeRollback(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	originalStatus, err := getRpmOstreeStatusJSON(c, m)
+	if err != nil {
+		c.Fatal(err)
+	}
+
+	// create a local branch to act as our upgrade target
+	originalOrigin := originalStatus.Deployments[0].Origin
```
arithx: Each time you reference `Deployments[0]` (ex: here, `postRebootStatus` & `originalStatus` in `rpmOstreeInstallUninstall`) it would be safer to add a `len(<status_var>.Deployments) > 0` safety check to prevent panic's if unexpected results occur.

kola/tests/rpmostree/status.go
miabbott: Got it; will remove this in an updated commit

kola/tests/rpmostree/deployments.go
```diff
@@ -0,0 +1,300 @@
...
+	// e.g. fpaste-0.3.7.4.1-2.el7.noarch
+	rpmOut := c.MustSSH(m, "rpm -q fpaste")
+	rpmMatch := regexp.MustCompile("^fpaste.*noarch").MatchString(string(rpmOut))
+	if !rpmMatch {
+		c.Fatalf(`Output from "rpm -q" was unexpected: %q`, string(rpmOut))
+	}
+
+	// e.g. fpaste 0.3.7.4.1
+	verOut := c.MustSSH(m, "fpaste --version")
+	verMatch := regexp.MustCompile(`^fpaste \d+\.\d+\.\d+\.\d+\.\d+`).MatchString(string(verOut))
```
cgwalters: Why test the versioning scheme?  Doesn't seem worth it to potentially have a fire drill if fpaste changes its versioning scheme.  Maybe just test `fpaste --version` exits successfully?

kola/tests/rpmostree/deployments.go
```diff
@@ -0,0 +1,300 @@
...
+
+	// create a local branch to act as our upgrade target
+	originalOrigin := originalStatus.Deployments[0].Origin
+	originalCsum := originalStatus.Deployments[0].Checksum
+	createBranch := "sudo ostree refs --create local-branch " + originalCsum
+	c.MustSSH(m, createBranch)
+
+	// make a commit to the new branch
+	refspec := strings.Split(originalStatus.Deployments[0].Origin, ":")[1]
+	createCommit := "sudo ostree commit -b local-branch --tree ref=" + refspec + " --add-metadata-string version=kola-test-1.0"
```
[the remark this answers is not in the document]
miabbott: I was emulating what we did in a-h-t for this kind of test; didn't think to use commit ID

kola/tests/rpmostree/status.go
```diff
@@ -0,0 +1,171 @@
...
+		Run:         rpmOstreeStatus,
+		ClusterSize: 1,
+		Name:        "rhcos.rpmostree.status",
+		Distros:     []string{"rhcos"},
+	})
+}
+
+// hard code the osname for RHCOS
+// TODO: should this also support FCOS?
+var rhcosOsname string = "rhcos"
```
arithx: You could wrap up these three var's inside one block like so:

```golang
var (
  rhcosOsname = "rhcos"

  // comments here
  rhcosOriginRegex = ...
)

File: kola/tests/rpmostree/deployments.go [hunk 9]
```diff
@@ -0,0 +1,300 @@
...
+	newCommit := c.MustSSH(m, createCommit)
+
+	// edit origin file to point to new branch
+	currentDir := c.MustSSH(m, "sudo ostree admin --print-current-dir")
+	originFile := string(currentDir) + ".origin"
+	sedCmd := "sudo sed -i 's|^refspec=.*$|refspec=local-branch|' " + originFile
+	c.MustSSH(m, sedCmd)
+
+	// use rpm-ostree upgrade to get to the "new" commit
+	c.MustSSH(m, "sudo rpm-ostree upgrade")
```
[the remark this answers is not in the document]
miabbott: I was emulating what we did in a-h-t; I guess if the end result is the same I could use `rebase`, but I was trying to tickle the `upgrade` code path.

kola/tests/rpmostree/deployments.go
arithx: I don't believe that there is anything in the stdlib but there are definitely external libs for it.

kola/tests/rpmostree/deployments.go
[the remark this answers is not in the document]
cgwalters: Understood, but internally rpm-ostree is really doing almost exactly the same thing.  `upgrade`  along with most other verbs get unrolled into the generic `UpdateDeployment` DBus API.

One thing missing with this "sed behind rpm-ostree's back" approach is that it might be racy without `rpm-ostree reload`.  I also consider the `.origin` file format mostly private to ostree - in theory we could change, though it is unlikely.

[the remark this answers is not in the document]
cgwalters: (Although there are differences between `upgrade` and `rebase` - one of those is that the latter doesn't check that the timestamp on the commit is newer, but we're not testing that here)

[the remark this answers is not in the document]
miabbott: OK, since it's basically the same, I'll go with `rebase`.  Thanks for the additional details!

miabbott:
Updated to address comments from initial feedback! :arrow_up:

kola/tests/rpmostree/deployments.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	// validate we are back to the original deployment by comparing the
+	// the two rpmOstreeDeployment structs
+	if !reflect.DeepEqual(originalStatus.Deployments[0], rollbackStatus.Deployments[0]) {
+		c.Fatalf(`Differences found in "rpm-ostree status"; original %v, current: %v`, originalStatus.Deployments[0], rollbackStatus.Deployments[0])
+	}
+
+	// cleanup our mess
+	cleanupErr := rpmOstreeCleanup(c, m)
```
arithx: Unless we're wanting to test the cleanup we can drop this as the machine will just be terminated after the test is run (same for this in `rpmOstreeInstallUninstall`).

miabbott: So if I `kola run ... rhcos.rpmostree.*`, a new machine is started for each of the tests?

arithx: That's correct. Only `NativeFuncs` (e.x.: `rhcos.basic`) in the same test are ran on the same machine.

miabbott: Huh...TIL.  :smile:

miabbott: I think I'll keep those in there just as a sanity test of `cleanup`

kola/tests/rpmostree/deployments.go
```diff
@@ -0,0 +1,302 @@
...
+
+	m := c.Machines()[0]
+
+	originalStatus, err := getRpmOstreeStatusJSON(c, m)
+	if err != nil {
+		c.Fatal(err)
+	}
+
+	if len(originalStatus.Deployments) < 1 {
+		c.Fatalf(`Unexpected results from "rpm-ostree status"; no deployments present?`)
```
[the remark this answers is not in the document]
miabbott: Added in new commit.

miabbott:
I pushed a new commit yesterday that had one small change suggested.  :arrow_up:

miabbott:
good to merge

PR #917, opened 2018-09-17
crio: remove tty flag from commands

yuqi-zhang:
Fix tests as tty flag needs interactive flag, causing fails.


yuqi-zhang:
Good to merge

PR #920, opened 2018-09-21
platform: Drop checks for OS IDs

cgwalters:
This way if someone uses e.g. `-b rhcos`, we will believe them and
try.  Theoretically for example Flatcar should have its own OS ID
and set `ID_LIKE=coreos` or so - in general we should also support
`ID_LIKE` but that's a distinct path.


cgwalters:
OK, now with no comment.

bgilbert:
This is okay, but I'd rather have the safety check for the reasons @ajeddeloh mentioned.  Could we go ahead and add the additional case for FCOS?

PR #922, opened 2018-09-27
*: add fcos distribution

arithx:
Note: The test related changes are only for tests that run on the `qemu` platform.

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -494,6 +494,15 @@ func scpKolet(c cluster.TestCluster, mArch string) {
 			if err := c.DropFile(kolet); err != nil {
 				c.Fatalf("dropping kolet binary: %v", err)
 			}
+			// The default SELinux rules do not allow init_t to execute user_home_t
+			if Options.Distribution == "rhcos" || Options.Distribution == "fcos" {
+				for _, machine := range c.Machines() {
+					out, stderr, err := machine.SSH("sudo chcon -t bin_t kolet")
+					if err != nil {
+						c.Fatalf("running chcon on kolet: %s: %s: %v", out, stderr, err)
+					}
+				}
+			}
 			return
 		}
 	}
--- a/platform/platform.go
+++ b/platform/platform.go
@@ -299,14 +299,17 @@ func CheckMachine(ctx context.Context, m Machine) error {
 		return fmt.Errorf("ssh unreachable: %v", err)
 	}
 
-	// ensure we're talking to a Container Linux system
-	out, stderr, err := m.SSH("grep ^ID= /etc/os-release")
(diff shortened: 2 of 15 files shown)


platform/platform.go
```diff
@@ -299,14 +299,35 @@ func CheckMachine(ctx context.Context, m Machine) error {
...
 
-	if !bytes.Equal(out, []byte("ID=coreos")) && !bytes.Equal(out, []byte(`ID="rhcos"`)) {
-		return fmt.Errorf("not a supported instance")
+	supportedSystemCheck := func() (bool, error) {
+		// ensure we're talking to a supported system
+		switch string(out) {
+		case `ID=coreos`, `ID="rhcos"`:
+			return true, nil
+		case `ID=fedora`:
+			variantid, _, err := m.SSH("grep ^VARIANT_ID= /etc/os-release")
```
cgwalters: I'd do `. /etc/os-release && echo "$VARIANT_ID"` - the file is shell syntax, and your "parser" here doesn't handle quotes.

cgwalters: Also worth noting this obsoletes https://github.com/coreos/mantle/pull/920

kola/harness.go
```diff
@@ -493,6 +493,14 @@ func scpKolet(c cluster.TestCluster, mArch string) {
 			if err := c.DropFile(kolet); err != nil {
 				c.Fatalf("dropping kolet binary: %v", err)
 			}
+			if Options.Distribution == "rhcos" || Options.Distribution == "fcos" {
+				for _, machine := range c.Machines() {
+					out, stderr, err := machine.SSH("sudo chcon -t bin_t kolet")
```
jlebon: Would be good to add some context comment here on why this is needed.

arithx:
Updates pushed addressing comments :arrow_up:

arithx:
Pushed addressing comments :arrow_up:

platform/platform.go
```diff
@@ -299,14 +299,24 @@ func CheckMachine(ctx context.Context, m Machine) error {
...
-	// ensure we're talking to a Container Linux system
-	out, stderr, err := m.SSH("grep ^ID= /etc/os-release")
+	out, stderr, err := m.SSH(`. /etc/os-release && echo "$ID-$VARIANT_ID"`)
 	if err != nil {
 		return fmt.Errorf("no /etc/os-release file: %v: %s", err, stderr)
 	}
 
-	if !bytes.Equal(out, []byte("ID=coreos")) && !bytes.Equal(out, []byte(`ID="rhcos"`)) {
-		return fmt.Errorf("not a supported instance")
+	supportedSystemCheck := func() bool {
```
bgilbert: The inline function isn't needed any longer.

PR #923, opened 2018-09-28
kola/tests: add podman tests

mike-nguyen:
Add base (info, resources, network reliability) and network
connectivity tests between machines for podman.  These tests are based
off the existing docker/crio tests in the repo.

There is an addition workflow test (requires internet) that uses
ngingx to test:
  - running/pulling a container with volume mount and port forwarding
  - executing commands in an running container
  - stopping a container
  - removing a container
  - deleting a container

### DIFF
--- a/kola/tests/podman/podman.go
+++ b/kola/tests/podman/podman.go
@@ -1,0 +1,343 @@
+// Copyright 2018 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package podman
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"golang.org/x/crypto/ssh"
+	"golang.org/x/net/context"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	tutil "github.com/coreos/mantle/kola/tests/util"
+	"github.com/coreos/mantle/lang/worker"
+	"github.com/coreos/mantle/platform"
+	"github.com/coreos/mantle/util"
+)
+
+// init runs when the package is imported and takes care of registering tests
+func init() {
+	register.Register(&register.Test{
+		Run:         podmanBaseTest,
+		ClusterSize: 1,
+		Name:        `podman.base`,
+		Distros:     []string{"rhcos"},
+	})
+	register.Register(&register.Test{
+		Run:              podmanWorkflow,
+		ClusterSize:      1,
+		Name:             `podman.workflow`,
+		ExcludePlatforms: []string{"qemu"}, // Requires internet for pulling nginx
+		Distros:          []string{"rhcos"},
+	})
+	register.Register(&register.Test{
+		Run:         podmanNetworkTest,
+		ClusterSize: 2,
+		Name:        `podman.network`,
+		Distros:     []string{"rhcos"},
+	})
+}
+
+// simplifiedContainerPsInfo represents a container entry in podman ps -a
+type simplifiedContainerPsInfo struct {
+	Id     string `json:"id"`
+	Image  string `json:"image"`
+	Status string `json:"status"`
+}
+
+// simplifiedPsInfo represents the results of podman ps -a
+type simplifiedPsInfo struct {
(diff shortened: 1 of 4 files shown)


kola/tests/podman/podman.go
```diff
@@ -0,0 +1,352 @@
...
+	c.Run("info", podmanInfo)
+	c.Run("resources", podmanResources)
+	c.Run("network", podmanNetworksReliably)
+}
+
+// Test: Run basic podman commands
+func podmanWorkflow(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	// Test: Verify container can run with volumen mount and port forwarding
```
miabbott: nit: s/volumen/volume

kola/tests/podman/podman.go
```diff
@@ -0,0 +1,352 @@
...
+}
+
+func podmanBaseTest(c cluster.TestCluster) {
+	c.Run("info", podmanInfo)
+	c.Run("resources", podmanResources)
+	c.Run("network", podmanNetworksReliably)
+}
+
+// Test: Run basic podman commands
+func podmanWorkflow(c cluster.TestCluster) {
```
arithx: Ideally we'd like to keep tests self-contained, having unreported sub tests that depend on each other isn't optimal. In this specific case I would be okay with this for now.

It does start me thinking about adding some sort of `failfast` setting inside a `TestCluster` though. If we had a feature like that we could pretty easily wrap each comment marked block in a `c.Run` and let the harness keep track of whether a previous sub test has failed (immediately marking any sub tests following a failed sub test as failed without running them, keeping the functionality of this code).

kola/tests/podman/podman.go
```diff
@@ -0,0 +1,352 @@
...
+	if info.Store.GraphRoot != expectedGraphRoot {
+		c.Errorf("Unexected graph root: %v != %v", expectedGraphRoot, info.Store.GraphRoot)
+	}
+}
+
+// Test: Run podman with various options
+func podmanResources(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	genPodmanContainer(c, m, "sleep", []string{"sleep"})
```
arithx: Why `sleep` over say `echo`?

kola/tests/podman/podman.go
```diff
@@ -0,0 +1,352 @@
...
+
+	err = json.Unmarshal(pInfoJson, &target)
+	if err != nil {
+		return target, fmt.Errorf("Could not unmarshal info %q into known JSON: %v", string(pInfoJson), err)
+	}
+	return target, nil
+}
+
+// make a podman container out of binaries on the host
+func genPodmanContainer(c cluster.TestCluster, m platform.Machine, name string, binnames []string) {
```
arithx: This seems to be the same as a sub-section of the [`genContainer`](https://github.com/coreos/mantle/blob/master/kola/tests/crio/crio.go#L216-L220) function in the cri-o tests. This should be split out into the `kola/tests/util` package and referenced in both locations to prevent code duplication.

kola/tests/podman/podman.go
```diff
@@ -0,0 +1,352 @@
...
+		if err != nil {
+			return err
+		}
+		if !bytes.Contains(b, []byte("TEST PAGE")) {
+			return fmt.Errorf("nginx pod is not running %s", b)
+		}
+		return nil
+	}
+
+	if err := util.Retry(10, 10*time.Second, podIsRunning); err != nil {
```
arithx: Does this really need 100+ seconds? Can we bump this down to 30 seconds with something like `util.Retry(6, 5*time.Second, podIsRunning)`?

mike-nguyen: Probably not.  I will bump down to 30 seconds.

kola/tests/podman/podman.go
mike-nguyen: Sleep was used in the docker test but there's no particular reason we need it here.  `echo` works and its faster!  Thanks for the suggestion!

mike-nguyen:
Pushed a new commit to address comments.

kola/tests/util/containers.go
```diff
@@ -0,0 +1,31 @@
...
+package util
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/platform"
+)
+
```
miabbott: Could some comments be added here about what this is doing or why you would use it?

arithx: Would `GenPodmanScratchContainer` convey the intent of the function better?

mike-nguyen: I changed the name of the function and also added a comment for the exported function (golint told me I should have done that).

mike-nguyen:
Another update pushed to address comments :arrow_up:

kola/tests/podman/podman.go
arithx: I have up a PR for a `FailFast` flag in #926 

That patch would allow you to wrap each commented test block in a `c.Run` and then if one sub-test fails all future sub-tests would be skipped.

mike-nguyen:
ready to merge

PR #924, opened 2018-10-01
glide: pin golang.org/x/sys to current version

arithx:


ajeddeloh:
Do we know if this is versioned with the golang standard library or just "we promise master is stable"?

arithx:
@ajeddeloh afaict "we promise master is stable"

arithx:
@bgilbert re-pushed with context in commit message

PR #926, opened 2018-10-01
kola: Add FailFast flag to test registration

arithx:
Adds a new flag (`FailFast`) to test registration, when set to true any
sub-tests will be skipped if a previous sub-test has failed.

### DIFF
--- a/kola/register/register.go
+++ b/kola/register/register.go
@@ -49,6 +49,10 @@ type Test struct {
 	Architectures    []string // whitelist of machine architectures supported -- defaults to all
 	Flags            []Flag   // special-case options for this test
 
+	// FailFast skips any sub-test that occurs after a sub-test has
+	// failed.
+	FailFast bool
+
 	// MinVersion prevents the test from executing on CoreOS machines
 	// less than MinVersion. This will be ignored if the name fully
 	// matches without globbing.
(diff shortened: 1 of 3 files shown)


arithx:
Works for both tests that perform `c.Run` inside of the test code as well as `NativeFuncs` which perform that under the hood.

arithx:
Pause on review while I investigate parallel sub-tests a bit more to make sure there won't be any future potential edge cases that aren't covered by this.

kola/register/register.go
```diff
@@ -49,6 +49,10 @@ type Test struct {
 	Architectures    []string // whitelist of machine architectures supported -- defaults to all
 	Flags            []Flag   // special-case options for this test
 
+	// FailFast skips any sub-test that occurs after a sub-test has
+	// failed.
+	FailFast bool
```
bgilbert: More descriptive name?

arithx: Any ideas? This sort of functionality in other test runners (including [golang](https://go-review.googlesource.com/c/go/+/74450)) usually uses `FailFast` or some derivative.

bgilbert: Ah, if it's standard terminology let's stick with it.

PR #929, opened 2018-10-04
kola/tests: use FailFast for podman workflow test

mike-nguyen:
Use the FailFast functionality introduced in #926 for the podman
workflow test so test reporting is clearer and dependent tests are
skipped when a failure occurs.

Also fixed some minor variable casing caught by golint.


mike-nguyen:
ready to merge

PR #930, opened 2018-10-09
platform/api/aws/networking: detect broken networking

### LINKED ISSUE #914: Kola fails on AWS if the kola security group doesn't have a VPC attached
Minor bug: If the VPC gets pruned, but the security group is not deleted, it will cause a failure like:

`error resolving subnet: no subnets found for vpc vpc-xxxxxx`

Can be worked around by deleting security group and have Kola re-create everything.

arithx:
Add detection for missing networking pieces when selecting a security
group (tearing down and recreating them if resources are missing).

Fixes #914

### DIFF
--- a/platform/api/aws/network.go
+++ b/platform/api/aws/network.go
@@ -44,6 +44,14 @@ func (a *API) getSecurityGroupID(name string) (string, error) {
 		return "", fmt.Errorf("unable to get security group named %v: %v", name, err)
 	}
 
+	// Validates that the required networking resources are present, if not attempts to
+	// delete the Security Group, VPC, and relevant networking resources
+	if valid, err := a.validateNetworkResources(sgIds.SecurityGroups[0]); err != nil {
+		return "", err
+	} else if !valid {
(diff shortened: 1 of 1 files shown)


platform/api/aws/network.go
```diff
@@ -340,3 +348,115 @@ func (a *API) getVPCID(sgId string) (string, error) {
...
+
+	return true, nil
+}
+
+func (a *API) DeleteVPC(sgId *string, vpc *ec2.Vpc) error {
+	_, err := a.ec2.DeleteSecurityGroup(&ec2.DeleteSecurityGroupInput{
+		GroupId: sgId,
+	})
+	if err != nil {
+		return fmt.Errorf("deleting security group: %v", err)
```
bgilbert: Are the resource IDs included in `err`?  If not, they should be added to these error messages.

arithx:
After OOB discussions this PR is going to get split up once Flight (#940) merges into the creation/deletion of networking resources in AWS as `ore` commands & a networking validation inside of the AWS Flight object creation.

PR #932, opened 2018-10-11
cmd/kola: list improvements

arithx:
Adds distribution information to the list subcommand. Also adds a new
flag `--json` to the list subcommand.

### DIFF
--- a/cmd/kola/kola.go
+++ b/cmd/kola/kola.go
@@ -60,11 +60,15 @@ will be ignored.
 		Short: "List kola test names",
 		Run:   runList,
 	}
+
+	listJSON bool
 )
 
 func init() {
 	root.AddCommand(cmdRun)
 	root.AddCommand(cmdList)
+
+	cmdList.Flags().BoolVar(&listJSON, "json", false, "format output in JSON")
 }
 
 func main() {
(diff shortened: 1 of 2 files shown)


cgwalters:
Seems sane to me, though your commit message says "what" but not "why" - I can guess at the "why" but I like to have commit messages at least include a single line rationale if possible; e.g. "Will use this for Jenkins" or something.

arithx:
Ah yeah, I'll update it.

The main content for why is to a.) keep it in sync with the filtering options available & b.) allow for something like a Jenkins job to be able to easily parse out the test list and farm them out to individual pods.

arithx:
Updated commit message. ⬆️

cmd/kola/kola.go
```diff
@@ -250,12 +269,50 @@ func (i item) String() string {
...
+			i.Platforms = []string{"all"}
+		}
 	}
 	if len(i.Architectures) == 0 {
-		i.Architectures = []string{"all"}
+		if listJSON {
+			i.Architectures = kolaArchitectures
+		} else {
+			i.Architectures = []string{"all"}
+		}
```
cgwalters: I don't know this codebase well, so this is a question - is the idea here that when emitting JSON, we want to declare what architectures we support, but when evaluating the tests, we want to assume that the test supports all of them?

Offhand, it feels a bit weird to do it this way versus having e.g. toplevel JSON elements `platforms: $kolaPlatforms, architectures: $kolaArchitectures` and then the tests default to "all" the same as they do internally?

cgwalters:
One optional question/comment; from my POV free to merge as is too.

cmd/kola/kola.go
arithx: With CL dropping arm64 support (and mantle subsequently following) there is only one architecture.

As far as why it's `all` on the stdout approach vs naming them all in the JSON, I found that there'd be more value in enumerating them for CI in JSON while still maintaining the current functionality of using `all`.

miabbott:
I like the enhancements.  The alignment is a bit off but, not worth reason enough to block a merge.

```
Test Name                                       Platforms       Architectures   Distributions       
                                                                                                    
cl.basic                                        [all]                           [all]   [cl]   
cl.cloudinit.basic                              [all]                           [all]   [cl]           
cl.cloudinit.script                             [all]                           [all]   [cl]           
cl.disk.raid.data                               [all]                           [all]   [cl]        
cl.disk.raid.root                               [qemu]                          [all]   [cl]           
cl.etcd-member.discovery                        [aws do esx gce packet]         [all]   [cl]           
```

cmd/kola/kola.go
```diff
@@ -204,37 +208,52 @@ func writeProps() error {
...
+		var w = tabwriter.NewWriter(os.Stdout, 0, 8, 0, '\t', 0)
+
+		fmt.Fprintln(w, "Test Name\tPlatforms\tArchitectures\tDistributions")
+		fmt.Fprintln(w, "\t")
+		for _, item := range testlist {
+			fmt.Fprintf(w, "%v\n", item)
+		}
+		w.Flush()
+	} else {
+		out, err := json.Marshal(testlist)
```
bgilbert: `MarshalIndent`?

cmd/kola/kola.go
```diff
@@ -204,37 +208,52 @@ func writeProps() error {
...
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "marshalling test list: %v", err)
+			os.Exit(1)
+		}
+		fmt.Fprintf(os.Stdout, string(out))
```
bgilbert: Variable data is being used as a format string here.  You probably just want `fmt.Println`.

cmd/kola/kola.go
```diff
@@ -204,37 +208,52 @@ func writeProps() error {
...
 	Name             string
 	Platforms        []string
 	ExcludePlatforms []string
 	Architectures    []string
+	Distros          []string
+	ExcludeDistros   []string
 }
 
-func (i item) String() string {
+func (i *item) UpdateValues() {
```
bgilbert: Unexported would be a bit clearer.

cmd/kola/kola.go
```diff
@@ -204,37 +208,52 @@ func writeProps() error {
...
-	w.Flush()
 }
 
 type item struct {
 	Name             string
 	Platforms        []string
 	ExcludePlatforms []string
 	Architectures    []string
+	Distros          []string
+	ExcludeDistros   []string
```
bgilbert: The `Exclude` fields should be `omitempty`.

cmd/kola/kola.go
```diff
@@ -250,12 +269,50 @@ func (i item) String() string {
 			}
 		}
 		i.Platforms = platforms
+	} else {
+		i.ExcludePlatforms = []string{""}
```
bgilbert: These are unnecessary with `omitempty`.

cmd/kola/kola.go
bgilbert: > With CL dropping arm64 support (and mantle subsequently following) there is only one architecture.

Side note: mantle isn't going to drop arm64 support because we'll need it in the future.

cmd/kola/kola.go
```diff
@@ -250,12 +269,50 @@ func (i item) String() string {
...
 	if len(i.Architectures) == 0 {
-		i.Architectures = []string{"all"}
+		if listJSON {
+			i.Architectures = kolaArchitectures
+		} else {
+			i.Architectures = []string{"all"}
+		}
 	}
-	return fmt.Sprintf("%v\t%v\t%v", i.Name, i.Platforms, i.Architectures)
+	if len(i.ExcludeDistros) > 0 {
```
bgilbert: Rather than copypasting this code, there could be a helper function (perhaps an anonymous one) that takes three string slices.

cmd/kola/kola.go
```diff
@@ -250,12 +269,50 @@ func (i item) String() string {
...
+		for _, distro := range i.Distros {
+			if _, ok := excludeDistros[distro]; !ok {
+				distros = append(distros, distro)
+			}
+		}
+		i.Distros = distros
+	} else {
+		i.ExcludeDistros = []string{""}
+	}
+	if len(i.Distros) == 0 {
```
bgilbert: This block could also be abstracted into a helper, though there seems less value in it.

cmd/kola/kola.go
arithx: Ah yeah, might as well, even if the primary use case is for CI / scripting it doesn't hurt to do that.

cmd/kola/kola.go
arithx: There's a subtle difference in `jq` between omit'd fields & fields containing empty lists (doing `.foo` if it's omit'd results in `null` vs `[]`). My thought was to leave it this way intentionally for easier scripting.

bgilbert: Ah, I misread the code.  Do consumers need `Exclude*` at all?  Would it be sufficient to produce a list of platforms/distros which _are_ enabled for each test?

arithx: Yeah, it should be sufficient to just provide an enabled list.

arithx:
Pushed an update that no longer outputs `Exclude*` properties in the JSON, utilizes a generic helper method, and uses the `MarshalIndent` function

cgwalters:
(superficial) LGTM

miabbott:
Built and worked for me in the simple `kola/list` case  :+1:

cmd/kola/kola.go
```diff
@@ -204,58 +208,88 @@ func writeProps() error {
...
-	if len(i.Platforms) == 0 {
-		i.Platforms = []string{"all"}
+		return retItems
 	}
-	if len(i.Architectures) == 0 {
-		i.Architectures = []string{"all"}
-	}
-	return fmt.Sprintf("%v\t%v\t%v", i.Name, i.Platforms, i.Architectures)
+	i.Platforms = buildItems(i.Platforms, i.ExcludePlatforms, kolaPlatforms)
+	i.Architectures = buildItems(i.Architectures, []string{}, kolaArchitectures)
```
bgilbert: nit: `nil` should be sufficient for the second argument.

cmd/kola/kola.go
```diff
@@ -204,58 +208,88 @@ func writeProps() error {
...
+			}
+			retItems = items
+		} else {
+			retItems = include
 		}
-		platforms := []string{}
-		for _, platform := range i.Platforms {
-			if _, ok := excludePlatforms[platform]; !ok {
-				platforms = append(platforms, platform)
+		if len(retItems) == 0 {
```
bgilbert: If `exclude` contains every element of `all`, this concludes that the test will always run, which doesn't match the behavior of `filterTests()`.

arithx: Good catch, I'll add a case for ensuring that `exclude` isn't the same as `include`.

bgilbert: `exclude` could also directly contain everything in `all`, inadvertently or otherwise.

arithx: Should be addressed now, added a `sliceCompare` function and use it at the beginning of `buildItems` to compare if `exclude` & `all` are equal (returning an empty string list immediately if this is the case).

PR #934, opened 2018-10-15
Add functionality to upload Fedora images to AWS

sayanchowdhury:
This PR downloads the Fedora Cloud images and creates the AMIs for them in AWS. This PR also make the AMIs public


cmd/plume/fedoraspecs.go
```diff
@@ -0,0 +1,72 @@
...
+				"us-gov-west-1",
+			},
+		},
+	}
+
+	fedoraSpecs = map[string]channelSpec{
+		"prod": channelSpec{
+			BaseURL: "https://koji.fedoraproject.org/",
+			AWS: awsSpec{
+				BaseName:        "Fedora",
```
cgwalters: Is this "classic" Fedora Cloud?

cmd/plume/prerelease.go
```diff
@@ -62,7 +62,19 @@ var (
...
 	platformList []string
 
+	systems = map[string]system{
+		"cl": system{
+			displayName: "ContainerLinux",
+			handler:     runCLPreRelease,
+		},
+		"fedora": system{
+			displayName: "Fedora",
+			handler:     runFedoraPreRelease,
```
cgwalters: Have you looked at all at https://github.com/coreos/coreos-assembler/ - I have just started looking at the Plume code and I'm still trying to figure out how these intersect.

See https://github.com/coreos/coreos-assembler/issues/162

Are you thinking that we'd add a 3rd entry here for "FedoraCoreOS"?  Or are they close enough that they're the same?  I am not sure about that since currently Fedora content comes in composes, but that won't be true of FCOS...unless we change something there.

cgwalters:
Do you envison that plume would be run...as a container in Fedora Infrastructure?  And that we'd wrap it with some code to parse fedmsgs emitted by composes?

cmd/plume/fedoraspecs.go
sayanchowdhury: Yes, this is based on the classic Fedora Cloud

arithx:
Might make sense to create a `specs` sub-folder under `cmd/plume` that has something like `types.go`, `containerlinux.go`, & `fedora.go` to have easy separation.

cmd/plume/prerelease.go
sayanchowdhury: What I've noticed that the Fedora way of releasing is almost similar to that of CL, I don't think the same is true for FCOS. We need to add a different way to handle the FCOS scenario.

sayanchowdhury:
> Do you envison that plume would be run...as a container in Fedora Infrastructure? And that we'd wrap it with some code to parse fedmsgs emitted by composes?

Yes, the plan for the rewrite is that there would be some external code that would be calling of plume and performing the upload and would also be responsible for sending the fedmsg messages.

sayanchowdhury:
> Might make sense to create a `specs` sub-folder under `cmd/plume` that has something like `types.go`, `containerlinux.go`, & `fedora.go` to have easy separation.

That sounds like a sane idea.

cmd/plume/prerelease.go
cgwalters: Maybe we should have a quick meeting about this?  It'd be useful to have representation from people who understand all 3 release processes and figure out how we intersect them.  The "iron is hot" so to speak for coreos-assembler; we can make breaking changes if we need to.

cmd/plume/fedoraspecs.go
```diff
@@ -0,0 +1,72 @@
...
+				Image:           "fedora_production_ami_raw_image.raw.xz",
+				Partitions:      awsPartitions,
+			},
+		},
+	}
+)
+
+func ChannelFedoraSpec() channelSpec {
+	if specChannel == "" {
+		plog.Fatal("--channel is required")
```
lucab: In general, please consider returning a `(channelSpec, err)` instead, as `plog.Fatal` abruptly exit the process skipping any pending `defer` at caller sites.

cmd/plume/prerelease.go
```diff
@@ -162,14 +218,24 @@ func runPreRelease(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	return nil
+}
+
+func runPreRelease(cmd *cobra.Command, args []string) error {
+	if systemInfo, ok := systems[selectedSystem]; !ok {
+		return fmt.Errorf("Unknown system %q", selectedSystem)
+	} else {
```
lucab: You can flatten this `else` branch.

util/xz.go
```diff
@@ -0,0 +1,48 @@
...
+	reader, err := xz.NewReader(in)
+	if err != nil {
+		return err
+	}
+
+	_, err = io.Copy(out, reader)
+	if err != nil {
+		os.Remove(dst)
+	}
+	return err
```
lucab: In general, avoid doing this, as `error` is an interface: https://golang.org/doc/faq#nil_error

Just return the error in the conditional above and `return nil` here.

util/xz.go
```diff
@@ -0,0 +1,48 @@
...
+		return err
+	}
+
+	defer in.Close()
+
+	out, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return err
+	}
+
```
lucab: There is a missing `defer out.Close()` here.

cmd/plume/fedoraspecs.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+type awsPartitionFedoraSpec struct {
```
arithx: Why do we need to duplicate these structs with a different name?

EDIT: Looks like they are flattened in a later commit, just merge that one down.

cmd/plume/prerelease.go
```diff
@@ -127,6 +127,7 @@ func runFedoraPreRelease(system string, cmd *cobra.Command, args []string) error
 
 	spec := ChannelFedoraSpec()
 	ctx := context.Background()
+	client := http.Client{}
```
arithx: Some of the changes in this commit don't seem relevant to the commit title.

cmd/plume/fedora.go
```diff
@@ -102,6 +105,13 @@ var (
 	}
 )
 
+func AddFedoraSpecFlags(flags *pflag.FlagSet) {
```
arithx: Any reason to not just make a more generic `AddReleaseFlags` and put a note in the description of each that they're targeting `fedora` releases?

cmd/plume/prerelease.go
```diff
@@ -692,6 +701,25 @@ func awsPreRelease(ctx context.Context, system string, client *http.Client, src
...
+		if specChannel == "updates" || specChannel == "twoweek" {
+			builder := &strings.Builder{}
+			if err := t.Execute(builder, imageMetadata); err != nil {
+				panic(err)
+			}
+			s := builder.String()
+			fmt.Println(s)
+
+		} else if specChannel == "version" {
+			fmt.Println("In version")
```
arithx: Debug code

cmd/plume/prerelease.go
```diff
@@ -696,35 +697,23 @@ func awsPreRelease(ctx context.Context, system string, client *http.Client, src
...
-			fmt.Println("In version")
-		} else if specChannel == "branched" {
-			fmt.Println("In branched")
-		} else if specChannel == "cloud" {
-			fmt.Println("In cloud")
-		} else {
-			fmt.Println("To be skipped")
+		builder := &strings.Builder{}
+		if err := t.Execute(builder, imageMetadata); err != nil {
+			panic(err)
```
arithx: Plume shouldn't be using `panic`, return the error and exit more gracefully.

cmd/plume/prerelease.go
```diff
@@ -287,7 +288,7 @@ func getFedoraImageFile(client *http.Client, src *storage.Bucket, fileName strin
 		return imagePath, nil
 	}
 
-	rawxzURI, err := url.Parse(fileName)
+	rawxzURI, err := url.Parse("https://kojipkgs.fedoraproject.org/compose/cloud/Fedora-Cloud-29-20181114.0/compose/Cloud/x86_64/images/Fedora-Cloud-Base-29-20181114.0.x86_64.raw.xz")
```
arithx: Probably don't want this hard-coded.

cmd/plume/prerelease.go
```diff
@@ -475,7 +476,27 @@ func azurePreRelease(ctx context.Context, system string, client *http.Client, sr
 	return nil
 }
 
-func awsUploadToPartition(spec *channelSpec, part *awsPartitionSpec, imageName, imageDescription, imagePath string) (map[string]string, map[string]string, error) {
+func getSpecAWSImageName(spec *channelSpec, imageMetadata *map[string]interface{}) (string, string, string) {
+	imageFileName := spec.AWS.Image
+	imageName := fmt.Sprintf("%v-%v-%v", spec.AWS.BaseName, specChannel, specVersion)
+	imageName = regexp.MustCompile(`[^A-Za-z0-9()\\./_-]`).ReplaceAllLiteralString(imageName, "_")
```
arithx: Is it expected that `\` & `/` are allowed characters? And will `/` not conflict with the s3 buckets?

cmd/plume/prerelease.go
sayanchowdhury: Yeah! working on this one.

cmd/plume/prerelease.go
sayanchowdhury: This is something which was already there, and I just moved it into a util method.

arithx: cc @bgilbert 

Is there a potential bug if images were to have `/` in the file name that would dump it into a new bucket?

sayanchowdhury:
@arithx I've rearranged the commits to match the title of the commit and the code in it? Can you give one more review? I've to work on the rework thing on the hardcoded URL. I'm working on it. Rest should be good.

@lucab Can you go through the PR again as I pushed more changed to the PR? I've addressed the comments you posted but it was before I did a major push

cmd/plume/prerelease.go
bgilbert: In the original code, at least, the `imageName` is used for the human-readable snapshot and AMI names but _not_ for the S3 object name.

cmd/plume/prerelease.go
```diff
@@ -359,7 +476,33 @@ func azurePreRelease(ctx context.Context, client *http.Client, src *storage.Buck
...
+		buffer := &bytes.Buffer{}
+		if err := t.Execute(buffer, imageMetadata); err != nil {
+			return nil, err
+		}
+		imageFileName = buffer.String()
+		imageName = strings.TrimSuffix(imageFileName, ".raw.xz")
+		imageDescription = fmt.Sprintf("%v %v %v", spec.AWS.BaseDescription, specChannel, specVersion)
+	}
+
+	awsImageMeta := map[string]string{
```
bgilbert: This should be returning a struct, not a map.

cmd/plume/prerelease.go
```diff
@@ -86,17 +105,66 @@ func init() {
...
+			}
+		}
+		if !run {
+			continue
+		}
+
+		platform := platforms[platformName]
+		plog.Printf("Running %v pre-release...", platform.displayName)
+		if err := platform.handler(ctx, system, &client, nil, &spec, &imageInfo, &imageMetadata); err != nil {
+			plog.Fatal(err)
```
lucab: You can return this `err` instead.

cmd/plume/prerelease.go
```diff
@@ -162,14 +230,24 @@ func runPreRelease(cmd *cobra.Command, args []string) error {
...
 	}
 
+	return nil
+}
+
+func runPreRelease(cmd *cobra.Command, args []string) error {
+	if systemInfo, ok := systems[selectedSystem]; !ok {
+		return fmt.Errorf("Unknown system %q", selectedSystem)
+	} else if err := systemInfo.handler(selectedSystem, cmd, args); err != nil {
+		plog.Fatal(err)
```
lucab: You can un-nest this else branch, and return the `err` if non-nil.

cmd/plume/prerelease.go
```diff
@@ -200,6 +278,45 @@ func getImageFile(client *http.Client, src *storage.Bucket, fileName string) (st
...
+	if err := util.XZ2File(imagePath, rawxzPath); err != nil {
+		return "", err
+	}
+	return imagePath, nil
+}
+
+// getImageFile downloads a bzipped CoreOS image, verifies its signature,
+// decompresses it, and returns the decompressed path.
+func getImageFile(system string, client *http.Client, src *storage.Bucket, fileName string) (string, error) {
+	if system == "cl" {
```
lucab: Minor nit: this is usual written as a `switch` statement with an explicit error default case.

cmd/plume/specs.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 CoreOS, Inc.
+// Copyright 2018 Red Hat Inc.
```
arithx: None of this code is new, thus the date shouldn't be moved forward.

cmd/plume/prerelease.go
```diff
@@ -288,6 +288,12 @@ func getFedoraImageFile(client *http.Client, src *storage.Bucket, fileName strin
 	if err := sdk.UpdateFile(rawxzPath, rawxzURI.String(), client); err != nil {
 		return "", err
 	}
+
+	//decompress it
```
arithx: Please include a space after the `//` before starting your comment.

cmd/plume/fedora.go
```diff
@@ -0,0 +1,125 @@
+// Copyright 2018 Red Hat Inc.
```
arithx: This entire `plume: Draft the initial specs for Fedora images` commit is moving files to a different location that were originally moved/created in the `plume: Move the Fedora Specs into a different file` commit. Why not just move them into the correct location in the original commit?

cmd/plume/prerelease.go
```diff
@@ -699,30 +724,23 @@ func awsPreRelease(ctx context.Context, system string, client *http.Client, src
...
-		imageName = strings.TrimSuffix(imageFileName, ".raw.xz")
-		imageDescription = fmt.Sprintf("%v %v %v", spec.AWS.BaseDescription, specChannel, specVersion)
+	imageFileName, imageName, imageDescription := getSpecAWSImageName(spec, imageMetadata)
+	imagePath, err := getImageFile(system, client, src, imageFileName)
+	if err != nil {
+		return err
 	}
 
-	imagePath, err := getImageFile(system, client, src, imageFileNameTmpl)
+	vmdkImagePath := strings.Replace(imagePath, "raw", "vmdk", 1)
```
arithx: Isn't this block of code going to run on CL images as well which are already `vmdk` images?

cmd/plume/fedora.go
sayanchowdhury: Did this to segregate the flags for `fedora` and `containerlinux`

cmd/plume/prerelease.go
sayanchowdhury: I updated this part of code later in https://github.com/coreos/mantle/pull/934/commits/c7197d25e571317664032ec0dd32dec260f3f2e2

cmd/plume/specs.go
sayanchowdhury: I've migrated this into a new file `types.go`. So, I updated the date forward.

cmd/plume/prerelease.go
sayanchowdhury: This was fixed in a later commit

cmd/plume/prerelease.go
sayanchowdhury: Done!

cmd/plume/prerelease.go
sayanchowdhury: Changed this to `struct`

cmd/plume/prerelease.go
```diff
@@ -71,7 +85,12 @@ var (
 
 type platform struct {
 	displayName string
-	handler     func(context.Context, *http.Client, *storage.Bucket, *channelSpec, *imageInfo) error
+	handler     func(context.Context, string, *http.Client, *storage.Bucket, *channelSpec, *imageInfo, *map[string]interface{}) error
```
arithx: Why do both the `platform.handler` & `system.handler` take the `system` string? They should be able to reference the `selectedSystem` variable, similar to what's being done [here](https://github.com/coreos/mantle/pull/934/files#diff-6876a8022c0a64d94d3da2742d1114c3R125) to access the `selectedPlatforms` variable.

cmd/plume/prerelease.go
```diff
@@ -86,17 +105,70 @@ func init() {
...
+	imageMetadata := map[string]interface{}{
+		"Env":       specEnv,
+		"Version":   specFedoraVersion,
+		"Timestamp": specTimestamp,
+		"Respin":    specRespin,
+		"ImageType": specImageType,
+		"Arch":      specArch,
+	}
+
+	for _, platformName := range platformList {
```
arithx: I know this is most likely copy-pasted from the existing code but figured I'd post this up anyways.

Given we've already validated the `selectedPlatforms` list it seems like this could be simplified to:
```golang
for _, platformName := range selectedPlatforms {
    platform := platforms[platformName]
    plog.Printf("Running %v pre-release...", platform.displayName)
    if err := platform.handler(ctx, system, &client, nil, &spec, &imageInfo, &imageMetadata); err != nil {
        return nil
    }
}
```

cmd/plume/prerelease.go
```diff
@@ -575,24 +741,44 @@ func awsUploadAmiLists(ctx context.Context, bucket *storage.Bucket, spec *channe
...
 	}
 
+	if system == "fedora" {
+		vmdkImagePath := strings.Replace(imagePath, "raw", "vmdk", 1)
+		if _, err := os.Stat(vmdkImagePath); err == nil {
+			plog.Printf("Reusing existing vmdk image %q", vmdkImagePath)
+			return nil
+		}
+
+		cmd := exec.Command("qemu-img", "convert", "-f", "raw", "-O", "vmdk", "-o", "adapter_type=lsilogic,subformat=streamOptimized,compat6", imagePath, vmdkImagePath)
```
arithx: I'm not sure how I feel about plume doing this convert. We should probably either download existing bits that are in the proper format or have some other tool/script, outside of plume itself, that downloads & preps the correct release bits.

cc @bgilbert

cmd/plume/release.go
```diff
@@ -40,21 +39,62 @@ var (
 	cmdRelease    = &cobra.Command{
 		Use:   "release [options]",
 		Short: "Publish a new CoreOS release.",
-		Run:   runRelease,
+		RunE:  runRelease,
```
arithx: Why is this being switched to `RunE`? Given `runRelease` is the top level function it should be able to utilize `plog.Fatal` as it hasn't actually spawned any awaiting to be cleaned up once the error has bubbled up.

cmd/plume/specs.go
bgilbert: When copying existing code, you should always retain the existing copyright header.  The CoreOS acquisition is a bit of a special case, but anyway the old date shouldn't be dropped.  `2016-2018` is okay.

cmd/plume/prerelease.go
```diff
@@ -191,6 +184,14 @@ func runPreRelease(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	return nil
+}
+
+func runPreRelease(cmd *cobra.Command, args []string) error {
+	if _, ok := systems[selectedSystem]; !ok {
```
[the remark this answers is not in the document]
sayanchowdhury: I did a strip down of the commits, but I also agree there might be commits which later gets changed. But the issue is if we need to squash, we need to squash all the commits in between those two commits. I don't see any other option.

cmd/plume/fedora.go
sayanchowdhury: I skipped this squash as there was a commit in between these two, which might lead to conflicts.

cmd/plume/fedora.go
```diff
@@ -0,0 +1,120 @@
...
+				Partitions:      awsFedoraPartitions,
+			},
+		},
+	}
+)
+
+func AddFedoraSpecFlags(flags *pflag.FlagSet) {
+	flags.StringVarP(&specEnv, "environment", "E", "prod", "instance environment")
+	flags.StringVarP(&specImageType, "imagetype", "I", "Cloud-Base", "type of image")
+	flags.StringVarP(&specFedoraVersion, "fedoraversion", "F", "29", "fedora release version")
```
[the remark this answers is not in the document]
sayanchowdhury: I too +1 on this, but I could not come around with a solution to avoid this, as there were too many permutations & combinations. Another way would have been to pass all the names & URLs from the script that would call plume, but that would lead to a bunch of work on script side, and also removing the relevant code from this PR.

cmd/plume/fedora.go
```diff
@@ -0,0 +1,120 @@
...
+	flags.StringVarP(&specTimestamp, "timestamp", "T", "20181101", "compose timestamp")
+	flags.StringVarP(&specRespin, "respin", "R", "0", "compose respin")
+	flags.StringVarP(&specArch, "arch", "A", "x86_64", "compose arch")
+	flags.StringVarP(&specComposeID, "composeid", "O", "Fail", "compose id")
+}
+
+func ChannelFedoraSpec() (channelSpec, error) {
+	if specChannel == "" {
```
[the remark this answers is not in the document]
sayanchowdhury: Done

cmd/plume/specs.go
sayanchowdhury: Okay

cmd/plume/fedora.go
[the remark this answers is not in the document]
arithx: I'm not sure I understand, I meant this exclusively towards things that had actual duplicates in the existing set (like `version` & `fedoraversion` or `board` & `arch`).

Are you trying to make it so your script just calls plume without arguments?

cmd/plume/prerelease.go
```diff
@@ -83,9 +83,18 @@ var (
 	imageInfoFile      string
 )
 
+type imageMetadataAbsctract struct {
```
arithx: Typo: should be `imageMetadataAbstract`

cmd/plume/prerelease.go
sayanchowdhury: Again this would resolve when the commits would be sorted.

cmd/plume/fedora.go
sayanchowdhury: This would be fixed when the commits would be squashed into 3

cmd/plume/fedora.go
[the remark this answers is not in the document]
sayanchowdhury: @arithx @bgilbert @dustymabe So do we go here with the current set of arguments now, and later change them?

cmd/plume/prerelease.go
sayanchowdhury: cc @dustymabe

cmd/plume/prerelease.go
dustymabe: @arithx is this better now?

cmd/plume/prerelease.go
dustymabe: This still shows as no space for me.

cmd/plume/prerelease.go
dustymabe: is this resolved?

cmd/plume/prerelease.go
arithx: Yep.

cmd/plume/prerelease.go
dustymabe: If we don't feel comfortable doing the convert here then I think it would be OK to do the convert in the fedimg glue code that we will still need. @arithx do you see any other need ever for converting the input disk to the required format for upload?

cmd/plume/prerelease.go
arithx: This is now running against all platforms in `platformList` rather than the list of selected platforms. Same with the repeat block in `runCLPreRelease`

cmd/plume/prerelease.go
arithx: @dustymabe not inside of plume. Plume is built around taking already built release artifacts and publishing them. They should already be built and simply be consumed.

cmd/plume/release.go
```diff
@@ -55,6 +62,14 @@ func init() {
...
 
 func runRelease(cmd *cobra.Command, args []string) {
+	if systemInfo, ok := releaseSystems[selectedSystem]; !ok {
+		plog.Fatal("Unknown system %q", selectedSystem)
+	} else if err := systemInfo.handler(cmd, args); err != nil {
+		plog.Fatal(err)
+	}
+}
+
+func runCLRelease(cmd *cobra.Command, args []string) error {
```
arithx: There are a lot of remaining `plog.Fatal` calls still inside this function that can be changed to just returning the error and letting https://github.com/coreos/mantle/pull/934/commits/f73d8de4ade96285f13eb2864fe6e29a1e118cea#diff-70d5486f73eaac934075d6c18abdcc27R68 take care of it.

cmd/plume/prerelease.go
dustymabe: @sayanchowdhury can we add this to the fedimg glue code?

dustymabe: after a brief discussion in IRC we identified our options as:

1 - upload the raw (takes longer and wastes bandwidth)
2 - download the raw.xz uncomporess and convert to vmdk outside of plume
3 - do the convert inside plume (against philosophy) 
4 - make fedora produce a vmdk

let's go with 1 and I'll see about talking to releng to see if we can do 4

dustymabe: for `4` I opened this PR to see if we can make progress there: https://pagure.io/pungi-fedora/pull-request/683

sayanchowdhury:
I've made a few changes:
- Upload RAW Image for Fedora
- Not create PV images for Fedora
- Fix the failing tests [minor fix]

/cc @bgilbert @dustymabe

cmd/plume/prerelease.go
```diff
@@ -641,8 +793,10 @@ func awsPreRelease(ctx context.Context, client *http.Client, src *storage.Bucket
 		}
 	}
 
-	if err := awsUploadAmiLists(ctx, src, spec, &amis); err != nil {
-		return fmt.Errorf("uploading AMI IDs: %v", err)
+	if selectedSystem == "cl" {
+		if err := awsUploadAmiLists(ctx, src, spec, &amis); err != nil {
```
bgilbert: Do the Fedora AMI IDs not need to be stored anywhere?

cmd/plume/fedora.go
```diff
@@ -0,0 +1,116 @@
...
+
+	fedoraSpecs = map[string]channelSpec{
+		"rawhide": channelSpec{
+			BaseURL: "https://koji.fedoraproject.org/compose/rawhide",
+			System:  "Fedora",
+			AWS: awsSpec{
+				BaseName:        "Fedora",
+				BaseDescription: "Fedora AMI",
+				Prefix:          "fedora_{{.Env}}_ami_",
+				Image:           "Fedora-AtomicHost-{{.Version}}-{{.Timestamp}}.{{.Respin}}.{{.Arch}}.raw.xz",
```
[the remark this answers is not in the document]
sayanchowdhury: Yes, `AtomicHost` is hardcoded here.

cmd/plume/prerelease.go
```diff
@@ -235,6 +280,39 @@ func getCLImageFile(client *http.Client, src *storage.Bucket, fileName string) (
...
+	rawxzPath := filepath.Join(cacheDir, fileName)
+	imagePath := strings.TrimSuffix(rawxzPath, filepath.Ext(rawxzPath))
+
+	if _, err := os.Stat(imagePath); err == nil {
+		plog.Printf("Reusing existing image %q", imagePath)
+		return imagePath, nil
+	}
+
+	if specImageType == "Cloud-Base" {
+		specImageType = "Cloud"
```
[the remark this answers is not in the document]
sayanchowdhury: The reason being Cloud images being the exceptions where the URL contains `Cloud` where are the image name is `Cloud-Base`. For all other the name in URL and image name is the same.

cmd/plume/prerelease.go
```diff
@@ -74,6 +82,11 @@ type platform struct {
 	handler     func(context.Context, *http.Client, *storage.Bucket, *channelSpec, *imageInfo) error
 }
 
+type system struct {
+	displayName string
```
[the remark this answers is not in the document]
sayanchowdhury: I guess this was for the `system` struct. If not, please un-resolve it.

cmd/plume/fedora.go
```diff
@@ -0,0 +1,116 @@
...
+	specImageType       string
+	specTimestamp       string
+	specFedoraVersion   string
+	specFedoraBoard     string
+	awsFedoraBoards     = []string{"amd64-usr"}
+	awsFedoraPartitions = []awsPartitionSpec{
+		awsPartitionSpec{
+			Name:         "AWS",
+			Profile:      "default",
+			Bucket:       "fedora-s3-bucket-fedimg-test",
```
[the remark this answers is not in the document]
sayanchowdhury: I've added an 'user' channel to test.

cmd/plume/prerelease.go
bgilbert: Debug code shouldn't be included in any commit.

cmd/plume/prerelease.go
sayanchowdhury: @dustymabe Do we need to store this?

sayanchowdhury: This is something we are not doing right now. In the past, we had the necessity of the list of ami ids to delete the older ones periodically but I had a different approach there. I wanted to distinguish the AMIs using the AWS tags.

cmd/plume/fedora.go
```diff
@@ -0,0 +1,116 @@
...
+			BucketRegion: "us-east-1",
+			LaunchPermissions: []string{
+				"013116697141",
```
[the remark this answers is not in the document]
sayanchowdhury: We don't have an account alias. Should I create one? I can add a comment which is the production user id and which is the staging one.

cmd/plume/prerelease.go
dustymabe: We are essentially using fedmsg right now to retrieve AMI IDs. As long as we are still sending out fedmsg I would say we're fine not storing them otherwise and we can improve in the future.

dustymabe: @sayanchowdhury - can you confirm we'll still be sending out fedmsg?

cmd/plume/fedora.go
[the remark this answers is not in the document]
dustymabe: @bgilbert would it suffice to add a comment saying `#fedora aws main account` or `#fedora aws stage account` - i.e. terminology fedora infra/releng would understand?

cmd/plume/prerelease.go
sayanchowdhury: Yes, we will be sending out the fedmsg messages.

cmd/plume/fedora.go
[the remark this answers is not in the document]
sayanchowdhury: @dustymabe Exactly. That should suffice.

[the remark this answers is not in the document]
bgilbert: SGTM.

cmd/plume/fedora.go
```diff
@@ -0,0 +1,154 @@
...
+				Partitions:      awsFedoraPartitions,
+			},
+		},
+		"branched": channelSpec{
+			BaseURL: "https://koji.fedoraproject.org/compose/branched",
+			AWS: awsSpec{
+				BaseName:        "Fedora",
+				BaseDescription: "Fedora AMI",
+				Prefix:          "fedora_{{.Env}}_ami_",
+				Image:           "Fedora-{{.ImageType}}-{{.Version}}-{{.Timestamp}}.n.{{.Respin}}.{{.Arch}}.raw.xz",
```
arithx: Should the `.n.` be a templated variable or is that the actual separator used to differentiate between `cloud` & `branched` builds?

cmd/plume/fedora.go
```diff
@@ -0,0 +1,154 @@
...
+				Image:           "Fedora-{{.ImageType}}-{{.Version}}-{{.Timestamp}}.{{.Respin}}.{{.Arch}}.raw.xz",
+				Partitions:      awsFedoraUserPartitions,
+			},
+		},
+	}
+)
+
+func AddFedoraSpecFlags(flags *pflag.FlagSet) {
+	flags.StringVarP(&specEnv, "environment", "E", "prod", "instance environment")
+	flags.StringVarP(&specImageType, "imagetype", "I", "Cloud-Base", "type of image")
```
arithx: Typically we hyphenate multiple words rather than combining them for arguments (`--image-type` & `--compose-id`).

cmd/plume/prerelease.go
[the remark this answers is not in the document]
arithx: @sayanchowdhury changing the `specImageType` could have unintended side effects if/when multiple platforms are being ran. Having instead a function like so:
```golang
func getImageTypeURI() string {
    if specImageType == "Cloud-Base" {
        return "Cloud"
    }
    return specImageType
}
```
will achieve the same functionality without modify the global state.

cmd/plume/prerelease.go
```diff
@@ -388,7 +506,19 @@ func awsUploadToPartition(spec *channelSpec, part *awsPartitionSpec, imageName,
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("Could not generate the image metadata: %v", err)
+	}
+
+	imageFileName := awsImageMetadata["imageFileName"]
+	imageName = awsImageMetadata["imageName"]
+	imageDescription = awsImageMetadata["imageDescription"]
+
+	s3ObjectPath := fmt.Sprintf("%s/%s/%s", specBoard, specVersion, strings.TrimSuffix(imageFileName, filepath.Ext(imageFileName)))
+	if selectedDistro == "fedora" {
```
arithx: This `if` block seems redundant. It's setting `s3ObjectPath` to the same value as it was initially instantiated to above.

cmd/plume/prerelease.go
```diff
@@ -404,7 +534,13 @@ func awsUploadToPartition(spec *channelSpec, part *awsPartitionSpec, imageName,
 		}
 
 		plog.Printf("Creating EBS snapshot...")
-		snapshot, err = api.CreateSnapshot(imageName, s3ObjectURL, aws.EC2ImageFormatVmdk)
+
+		format := aws.EC2ImageFormatVmdk
```
arithx: Same comment as https://github.com/coreos/mantle/pull/934#discussion_r253616375

Don't instantiate the value and then have a special case for fedora, use an `if` or `switch` and set the value depending on the distro.

cmd/plume/prerelease.go
arithx: It looks like @dustymabe's PR has merged, can you please update this to remove the VMDK creation and use the VMDK images produced by Fedora rather than converting the raw image in this code.

cmd/plume/prerelease.go
arithx: Out of curiousity how are you getting the AMIs from to send out with `fedmsg`?

cmd/plume/prerelease.go
dustymabe: @sayanchowdhury you can try with the vmdk from https://koji.fedoraproject.org/koji/taskinfo?taskID=32866399. For some reason the vmdk isn't getting copied to the output directory yet. Opened an issue for that: https://pagure.io/releng/issue/8148

cmd/plume/fedora.go
dustymabe: yeah the `.n` means nightly in Fedora releng terms.

sayanchowdhury:
@arithx so to tackle the issue, i added a compose id tag to fedora images. a script would pick those amis and push messages for them.

arithx:
> @arithx so to tackle the issue, i added a compose id tag to fedora images. a script would pick those amis and push messages for them.

Is there a reason to not have plume output a file with the AMIs it generated rather than using a script to search for them in AWS?

cmd/plume/fedora.go
```diff
@@ -0,0 +1,154 @@
...
+	if specComposeID == "" {
+		plog.Fatal("--composeid is required")
+	}
+	if specTimestamp == "" {
+		plog.Fatal("--timestamp is required")
+	}
+	if specVersion == "" {
+		plog.Fatal("--version is required")
+	}
+	if specBoard == "" {
```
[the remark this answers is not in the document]
sayanchowdhury: Yes, I use the board to generate the architecture of the image, x86_64 & aarch64 are the two we will be using.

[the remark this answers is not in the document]
arithx: You don't specify any allowed board types on the spec so `x86_64` & `aarch64` are not valid options. 

Here's the specific code that is disallowing boards that are not present in the spec.
https://github.com/coreos/mantle/blob/bd4cd38e57f7f65cca993752150fc8d268d631f4/cmd/plume/containerlinux.go#L366-L375

sayanchowdhury:
> > @arithx so to tackle the issue, i added a compose id tag to fedora images. a script would pick those amis and push messages for them.
> 
> Is there a reason to not have plume output a file with the AMIs it generated rather than using a script to search for them in AWS?

I would in for of storing it via files but all I can think now creating the file in the fedimg box itself where the code would run. But, I am not sure about the availibilty of the files. That would need discussion maybe we can dump to S3. But for now going with the tagging option.

cmd/plume/prerelease.go
```diff
@@ -212,6 +267,35 @@ func getCLImageFile(client *http.Client, src *storage.Bucket, fileName string) (
...
+	}
+
+	rawxzURI, err := url.Parse(fmt.Sprintf("%v/%v/compose/%v/%v/images/%v", spec.BaseURL, specComposeID, getImageTypeURI(), specBoard, fileName))
+	if err != nil {
+		return "", err
+	}
+
+	plog.Printf("Downloading image %q to %q", rawxzURI, rawxzPath)
+
+	if err := sdk.UpdateFile(rawxzPath, rawxzURI.String(), client); err != nil {
```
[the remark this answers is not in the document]
sayanchowdhury: No, we don't have any signed images within Fedora

cmd/plume/fedora.go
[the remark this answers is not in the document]
arithx: Per OOB discussions: I must have initially missed adding the `--system fedora` flag when initially testing it and added it later when I got a working test case. We should still update this to only allow the boards that are valid `x86_64` & `aarch64` as mentioned above.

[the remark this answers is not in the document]
sayanchowdhury: I've added the code to allow specified boards from spec and  disallow any invalid argument. Please review.

arithx:
Local testing looks good, going to do a test run with the CL nightlies and then this should be good.

dustymabe:
Really glad to see this merged! Thanks to everyone involved!

dustymabe:
> Really glad to see this merged! Thanks to everyone involved!

Do we need to do a release so we can use it?

PR #935, opened 2018-10-16
kola/tests/ignition: add test for versioned s3 objects

### LINKED ISSUE #933: Add test for Ignition PR 643
https://github.com/coreos/ignition/pull/643 needs a test that needs to be in kola, not the bb tests.

arithx:
Adds a test for pulling versioned s3 objects (a new feature in
Ignition).

Tests both a null versionId (AWS uses this for the original version of a
file) and a later versionId.

Closes #933

### DIFF
--- a/kola/tests/ignition/resource.go
+++ b/kola/tests/ignition/resource.go
@@ -147,6 +147,40 @@ func init() {
 	      }`),
 		Distros: []string{"cl", "rhcos"},
 	})
+	// TODO: once Ignition supports this on all channels/distros
+	//       this test should be rolled into coreos.ignition.v2_1.resources.remote
+	// Test specifically for versioned s3 objects
+	register.Register(&register.Test{
+		Name:        "coreos.ignition.v2_1.resource.s3.versioned",
+		Run:         resourceS3Versioned,
+		ClusterSize: 1,
+		UserData: conf.Ignition(`{
+		  "ignition": {
+		      "version": "2.1.0"
+		  },
+		  "storage": {
+		      "files": [
+			  {
+			      "filesystem": "root",
+			      "path": "/var/resource/original",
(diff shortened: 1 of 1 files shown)


arithx:
This PR needs to be updated once coreos/ignition#643 merges and propogates to the relevant distributions / channels to have the `MinVersion` set.

lucab:
:heart:

kola/tests/ignition/resource.go
```diff
@@ -147,6 +147,40 @@ func init() {
...
+			      "contents": {
+				  "source": "http://s3-us-west-2.amazonaws.com/kola-fixtures/resources/versioned?versionId=null"
+			      },
+			      "mode": 420
+			  },
+			  {
+			      "filesystem": "root",
+			      "path": "/var/resource/latest",
+			      "contents": {
+				  "source": "http://s3-us-west-2.amazonaws.com/kola-fixtures/resources/versioned?versionId=RDWqxfnlcJOSDf1.5jy6ZP.oK9Bt7_Id"
```
lucab: Do we have content in this resource that changed across versions and that we can check once the node is provisioned?

arithx: Yes, in the original version (where `versionId` is `null`), the contents of the file are `original`, while in this version the contents are `updated`, see lines 245-248.

lucab:
@arithx I think we can update this `DNM` now. This feature was released in Ignition `0.29.0` and reached CL Alpha in 1995.0.0.

arithx:
@lucab yup, `0.29.1` is also in RPM form, merging now.

bgilbert:
Ignition 0.29.x isn't in beta or stable yet, so this will need a `MinVersion`.

PR #937, opened 2018-10-23
kola/tests: add ostree tests

miabbott:
This adds some simple tests to validate basic `ostree` functionality.  It should cover some of the most common use cases that a FCOS/RHCOS user would encounter.

I used the tests from `atomic-host-tests` as a guide, but did not port the full complement of `ostree` related tests from there.

### DIFF
--- a/kola/tests/ostree/basic.go
+++ b/kola/tests/ostree/basic.go
@@ -1,0 +1,292 @@
+// Copyright 2018 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ostree
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/kola/tests/util"
+	"github.com/coreos/mantle/platform"
+)
+
+// the "basic" test is only supported on 'rhcos' for now because of how
+// the refs are defined. if 'fcos' goes in the same direction, we can
+// expand support there.
+func init() {
+	register.Register(&register.Test{
+		Run:         ostreeBasicTest,
+		ClusterSize: 1,
+		Name:        "ostree.basic",
+		Distros:     []string{"rhcos"},
+		FailFast:    true,
+	})
+
+	register.Register(&register.Test{
+		Run:              ostreeRemoteTest,
+		ClusterSize:      1,
+		Name:             "ostree.remote",
+		Distros:          []string{"rhcos", "fcos"},
+		ExcludePlatforms: []string{"qemu"}, // need network to contact remote
+		FailFast:         true,
+	})
+}
+
+type ostreeAdminStatus struct {
+	Checksum string
+	Origin   string
+	Version  string
+}
+
+// getOstreeAdminStatus stuffs the important output of `ostree admin status`
+// into an `ostreeAdminStatus` struct
+func getOstreeAdminStatus(c cluster.TestCluster, m platform.Machine) (ostreeAdminStatus, error) {
+	oaStatus := ostreeAdminStatus{}
+
+	oasOutput, err := c.SSH(m, "ostree admin status")
+	if err != nil {
+		return oaStatus, fmt.Errorf(`Could not get "ostree admin status": %v`, err)
+	}
+
+	oasSplit := strings.Split(string(oasOutput), "\n")
+	if len(oasSplit) != 3 {
+		return oaStatus, fmt.Errorf(`Unexpected output from "ostree admin status": %v`, string(oasOutput))
+	}
+
+	// we use a bunch of regexps to find the content in each line of output
+	// from "ostree admin status".  the `match` for each line sticks the
+	// captured group as the last element of the array; that is used as the
+	// value for each field of the struct
+	reCsum, _ := regexp.Compile(`^\* [\w\-]+ ([0-9a-f]+)\.\d`)
+	csumMatch := reCsum.FindStringSubmatch(oasSplit[0])
+	if csumMatch == nil {
+		return oaStatus, fmt.Errorf(`Could not parse first line from "ostree admin status": %q`, oasSplit[0])
+	}
+	oaStatus.Checksum = csumMatch[len(csumMatch)-1]
+
+	reVersion, _ := regexp.Compile(`^Version: (.*)`)
+	versionMatch := reVersion.FindStringSubmatch(strings.TrimSpace(oasSplit[1]))
+	if versionMatch == nil {
+		return oaStatus, fmt.Errorf(`Could not parse second line from "ostree admin status": %q`, oasSplit[1])
+	}
+	oaStatus.Version = versionMatch[len(versionMatch)-1]
+
+	reOrigin, _ := regexp.Compile(`^origin refspec: (.*)`)
+	originMatch := reOrigin.FindStringSubmatch(strings.TrimSpace(oasSplit[2]))
+	if originMatch == nil {
+		return oaStatus, fmt.Errorf(`Could not parse third line from "ostree admin status": %q`, oasSplit[2])
+	}
+	oaStatus.Origin = originMatch[len(originMatch)-1]
+
+	return oaStatus, nil
+}
+
+// ostreeBasicTest performs sanity checks on the output from `ostree admin status`,
+// `ostree rev-parse`, and `ostree show` by comparing to the output from `rpm-ostree status`
+func ostreeBasicTest(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	ros, err := util.GetRpmOstreeStatusJSON(c, m)
+	if err != nil {
+		c.Fatal(err)
+	}
+
+	oas, err := getOstreeAdminStatus(c, m)
+	if err != nil {
+		c.Fatal(err)
+	}
+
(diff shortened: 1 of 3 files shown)


miabbott:
cc: @mike-nguyen

miabbott:
@arithx Yeah, you are probably right.  I just wanted to get the bones out there, but I can rework to use `FailFast`

miabbott:
Pushed a new commit that reorgs a lot of what I had done.

I used `FailFast` on all the tests since they all are relying on the success of a previous operation.

kola/tests/rpmostree/status.go
```diff
@@ -57,9 +59,9 @@ type simplifiedRpmOstreeStatus struct {
 	Deployments []rpmOstreeDeployment
 }
 
-// getRpmOstreeStatusJSON returns an unmarshal'ed JSON object that contains
+// GetRpmOstreeStatusJSON returns an unmarshal'ed JSON object that contains
 // a limited representation of the output of `rpm-ostree status --json`
-func getRpmOstreeStatusJSON(c cluster.TestCluster, m platform.Machine) (simplifiedRpmOstreeStatus, error) {
+func GetRpmOstreeStatusJSON(c cluster.TestCluster, m platform.Machine) (simplifiedRpmOstreeStatus, error) {
```
arithx: I think this function might fit in the [tests util folder](https://github.com/coreos/mantle/tree/master/kola/tests/util).

kola/tests/ostree/basic.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/kola/tests/rpmostree"
+	"github.com/coreos/mantle/platform"
+)
+
+func init() {
+	register.Register(&register.Test{
+		Run:         ostreeBasicTest,
+		ClusterSize: 1,
+		Name:        "rhcos.ostree.basic",
```
arithx: The test names (and the test names in `ostree/unlock.go`) should be changed to conform with the [kola test namespacing policy](https://github.com/coreos/mantle/blob/master/README.md#kola-test-namespacing) to either `coreos.ostree.*` or `ostree.*` as they support multiple distributions (the `rhcos.*` top level namespace would be if the tests only supported `rhcos`)

kola/tests/ostree/basic.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	oasSplit := strings.Split(string(oasOutput), "\n")
+	if len(oasSplit) != 3 {
+		return oaStatus, fmt.Errorf(`Unexpected output from "ostree admin status": %v`, string(oasOutput))
+	}
+
+	// take the first line of `ostree admin status`, split on whitespace,
+	// return the csum+serial, trim the serial suffix to get the csum
+	oaStatus.Checksum = strings.TrimSuffix(strings.Fields(oasSplit[0])[2], ".0")
```
arithx: I'm a bit weary about adding direct index references without size checks. Could we either a) split this out and add checks or b) switch to regex where a `nil` match is checked?

Same with the `Version` & `Origin` assignments below.

kola/tests/ostree/basic.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	oas, err := getOstreeAdminStatus(c, m)
+	if err != nil {
+		c.Fatal(err)
+	}
+
+	// verify the output from `ostree admin status`
+	c.Run("admin status", func(c cluster.TestCluster) {
+		if oas.Checksum != ros.Deployments[0].Checksum {
```
arithx: Should add a check that `len(ros.Deployments) > 0`

kola/tests/ostree/basic.go
```diff
@@ -0,0 +1,243 @@
...
+	})
+
+	// verify the output of 'ostree show'
+	c.Run("show", func(c cluster.TestCluster) {
+		oShowOut := c.MustSSH(m, ("ostree show " + oas.Checksum))
+		oShowOutSplit := strings.Split(string(oShowOut), "\n")
+
+		// convert the 'timestamp' from `rpm-ostree status` into a date that
+		// we can compare to the date in `ostree admin status`
+		// also, wtf is up with formatting date/time in golang?!
```
arithx: I ask myself that every time I use it 😃

kola/tests/ostree/basic.go
```diff
@@ -0,0 +1,243 @@
...
+		osRemoteListOut := c.MustSSH(m, "ostree remote list -u")
+		if len(strings.Split(string(osRemoteListOut), "\n")) < 1 {
+			c.Fatalf(`Did not find expected amount of remotes: %v`, string(osRemoteListOut))
+		}
+
+		osRemoteListSplit := strings.Split(string(osRemoteListOut), "\n")
+		var remoteFound bool = false
+		for _, v := range(osRemoteListSplit) {
+			lSplit := strings.Fields(v)
+			if lSplit[0] == remoteName && lSplit[1] == remoteUrl {
```
arithx: Should probably pre-check the length of `lSplit`

kola/tests/ostree/unlock.go
```diff
@@ -0,0 +1,263 @@
...
+		return fmt.Errorf(`RPM incorrectly in rpmdb after RPM uninstall: %v`, missRpmErr)
+	}
+
+	return nil
+}
+
+// ostreeUnlockTest verifies the simplest use of `ostree admin unlock` by
+// trying to install a dummy RPM on the host, rebooting, and ensuring
+// the RPM is gone after reboot
+func ostreeUnlockTest (c cluster.TestCluster) {
```
arithx: Would it be worthwhile to add a section checking re-locking the deployment?

kola/tests/ostree/unlock.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+var rpmUrl string = "https://raw.githubusercontent.com/projectatomic/atomic-host-tests/master/rpm/aht-dummy-1.0-1.noarch.rpm"
+var rpmName string = "aht-dummy"
+
+// ostreeAdminUnlock will unlock the deployment and verify the success of the operation
+func ostreeAdminUnlock (c cluster.TestCluster, m platform.Machine, hotfix bool) error {
+		var unlockCmd string = "sudo ostree admin unlock"
+		if hotfix {
+			unlockCmd = "sudo ostree admin unlock --hotfix"
```
arithx: `unlockCmd += " --hotfix"`?

kola/tests/ostree/unlock.go
```diff
@@ -0,0 +1,263 @@
...
+			return fmt.Errorf(`Failed to unlock deployment: %v`, unlockCmdErr)
+		}
+
+		status, err := rpmostree.GetRpmOstreeStatusJSON(c, m)
+		if err != nil {
+			c.Fatal(err)
+		}
+
+		// verify that the unlock was successful
+		if hotfix && status.Deployments[0].Unlocked != "hotfix" {
```
arithx: Should add a length check on `status.Deployments` to avoid the potential segfaults in these two if statements.

miabbott:
> There's probably a few more potential segfaults from not length checking the Deployments size I missed.

Ack, I'll keep that in mind when I go back to fix everything.  Thanks!

kola/tests/rpmostree/status.go
miabbott: OK, I'm going to break out this change into a separate PR.

Should the `structs` get defined in the `util` package too?

kola/tests/ostree/basic.go
miabbott: Makes sense.  Should I migrate the `kola/tests/rpmostree` tests similarly?

kola/tests/rpmostree/status.go
arithx: Yeah, if the structs are primarily created by this function they should probably live there as well.

As for breaking it out into a new PR, it's fine to me either way.

kola/tests/ostree/basic.go
arithx: Yeah, if I missed any of them feel free to add a commit or a separate PR.

miabbott:
I created another PR (#938) to clean up some of the `rpmostree` tests + func.  I think this PR would benefit from the changes there, so let's wait for that to merge first.

kola/tests/ostree/basic.go
```diff
@@ -0,0 +1,299 @@
...
+		preNumRemotes := len(strings.Split(string(preRemotesOut), "\n"))
+		if preNumRemotes < 1 {
+			c.Fatalf(`No remotes configured on host: %q`, string(preRemotesOut))
+		}
+
+		c.MustSSH(m, ("sudo ostree remote delete " + remoteName))
+
+		delRemoteListOut := c.MustSSH(m, "ostree remote list")
+		delNumRemotes := len(strings.Split(string(delRemoteListOut), "\n"))
+		if delNumRemotes < 1 {
```
arithx: I don't think this check is required

kola/tests/ostree/unlock.go
```diff
@@ -0,0 +1,263 @@
...
+		ClusterSize:      1,
+		Name:             "ostree.hotfix",
+		Distros:          []string{"rhcos", "fcos"},
+		ExcludePlatforms: []string{"qemu"}, // need network to pull RPM
+		FailFast:         true,
+	})
+
+}
+
+var rpmUrl string = "https://raw.githubusercontent.com/projectatomic/atomic-host-tests/master/rpm/aht-dummy-1.0-1.noarch.rpm"
```
arithx: minor nit: generally when declaring variables package wide we do it like so:

```golang
var (
  rpmUrl = "foo"
  rpmName = "bar"
)
```

kola/tests/ostree/unlock.go
```diff
@@ -0,0 +1,263 @@
...
+		return fmt.Errorf(`Failed to unlock deployment: %v`, unlockCmdErr)
+	}
+
+	status, err := util.GetRpmOstreeStatusJSON(c, m)
+	if err != nil {
+		c.Fatal(err)
+	}
+
+	// verify that the unlock was successful
+	if hotfix && status.Deployments[0].Unlocked != "hotfix" {
```
arithx: Should add a check on `len(status.Deployments)` like in `ostreeBasicTest`

miabbott:
Addressed the nits. etc above :arrow_up:

miabbott:
good to merge

PR #938, opened 2018-10-25
kola/tests: cleanup rpmostree tests + related funcs

miabbott:
This started with just moving `GetRpmOstreeStatusJSON` to the `util` package, but grew into re-working the `rpmostree` tests too.

I moved the `rpmostree` tests into their own namespace to be more compliant with the [kola test namespacing policy](https://github.com/coreos/mantle/blob/master/README.md#kola-test-namespacing).

Additionally, I opted to make some of the `rpmostree` tests use the `FailFast` capability.


miabbott:
good to merge

PR #939, opened 2018-10-29
rpmostree: Don't hardcode required osname/stateroot

cgwalters:
I changed this recently in the RHCOS pipeline, and I think we
want to support people running `kola` against e.g. custom FCOS
builds.

Changing the name really shouldn't break anything.

### DIFF
--- a/kola/tests/rpmostree/status.go
+++ b/kola/tests/rpmostree/status.go
@@ -35,8 +35,6 @@ func init() {
 }
 
 var (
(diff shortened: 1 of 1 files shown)


cgwalters:
Currently the stateroot is derived from: https://github.com/coreos/fedora-coreos-config/blob/master/fedora-coreos.yaml#L10

miabbott:
LGTM

cgwalters:
good to merge

Issue #803, opened 2018-02-01
Add platform interface representing an entire kola run


bgilbert:
Tentatively called a `Flight`.  It would serve as a Cluster factory and also contain infrastructure that shouldn't be recreated per Cluster.

This would allow:
- [x] Sharing platform API objects between clusters, allowing e.g. AWS SDK rate limiting to work correctly
- [x] Creating one SSH agent and platform SSH key for an entire kola run
- [ ] Creating one Azure resource group for an entire kola run
- [ ] On qemu, adding a caching proxy for quay.io (container images) and the torcx bucket
- [x] On qemu, launching one discovery etcd and one dnsmasq for an entire kola run
- [x] On qemu, integrating `kola mkimage` into `kola run` rather than requiring a separate step

bgilbert:
Leaving the qemu caching proxy for a followup project if needed.

Issue #799, opened 2018-01-29
RFE: kola should be able to run tests N times


ajeddeloh:
`kola run` should have a `--number` / `-n` flag to run the test(s) up to n times and stop on the first failure. It would also be good to utilize the parallel flag so runnning a single test can be done faster, but even a serial implementation would be nice. Furthermore a `--keep-going` flag could be useful to always run n tests and report the success/failure rate.

arithx:
I think that if an implementation of the `-n` flag is done we'd probably want to force the `--keep-going` functionality if the parallel flag is given. To properly implement the stop on first failure we'd have to plumb a lot of channels through the test clusters so when we do hit a failure we can initiate a clean exit/resource cleanup while also not reporting the status.

PR #940, opened 2018-11-05
Flight

bgilbert:
Add Flight object, which serves as a parent for all Clusters in a kola run.  Make the Flight responsible for platform infrastructure that shouldn't be recreated per Cluster.

Move the following into Flight:

- Platform API objects
- SSH agent and lifecycle of platform SSH keys
- On qemu: dnsmasq, discovery etcd, and NTP server
- On qemu: `kola mkimage` functionality

Fixes https://github.com/coreos/mantle/issues/803.


bgilbert:
The GitHub UI reordered the commits.  The correct order is:

- 1125b405b10f18fdb90e66833f93a8aec2ef65ef cmd/kola: fix gofmt
- 4122ea7f7faffdf3779156defcc9fa572cc9f407 platform: rename base.go to cluster.go
- 94a75122e191e92cfd4b73ca05bd310de945bb26 platform: move logger into platform.go
- 5515f490411c40e394c28b92b91b4d06d56a0335 platform: add Name() to Cluster interface
- f1af767188718fad86bc469d80a513d431f2a437 platform: add Flight
- 4ae4017f632d64a6d8c2042639aeddc1d2fb7d8f kola: drop rate-limiting hack
- 63b83c73cbc97a197a07aac71a557941de39d2b2 *: don't hardcode hostport of local Omaha server
- c0f3982bf8c84f0de1089312eb87350df179a657 platform/local: dynamically select Omaha listen port
- f502e17c67d4400e66525716834496d6a3362b3d platform/local: only create one network segment
- e7ee538256885850cad2114c42dd53608f746d87 platform/local: increase interface limit beyond 8 bits
- fc7bdd84c11ae9ee0d71dbfbb5433358de9f5734 platform/local: use one netns per flight
- c36fefacf8441b6f604507786342ada0a7173060 platform: add Flight.Name()
- 172a285fea96821534655fb3bb90942ef09c3fad platform: move SSH agent into flight
- 4fdd92ba20f73166dd6548705e98070f07236561 platform/qemu: move disk-related code into disk.go
- 0aef98a0c37bfce7ed90068444666609336e1a22 platform/qemu: unexport Disk methods
- 58629d9c2e87ffdf3be4c7453133b6ba0a580d1f cmd/kola: drop mkimage subcommand
- d9641f819ce57a6b73b6d83137ac234ba133acb8 platform/qemu: add helper function to make safe tempfile name
- df856e9f51006112fc1adf025626cf022de4312e platform/qemu: enable console logging in CL images by default

dustymabe:
nice work @bgilbert !

platform/machine/qemu/disk.go
```diff
@@ -195,3 +193,12 @@ func setupDisk(additionalOptions ...string) (*os.File, error) {
 
 	return os.OpenFile(dstFileName, os.O_RDWR, 0)
 }
+
+func mkpath(basedir string) (string, error) {
+	f, err := ioutil.TempFile(basedir, "mantle-qemu")
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
```
arithx: This seems to have a subtle difference in functionality as this will be closed at the exit of `mkpath` rather than `setupDisk` but given the rest of the function it doesn't seem to change anything.

platform/machine/aws/flight.go
```diff
@@ -0,0 +1,91 @@
...
+	api *aws.API
+}
+
+// NewFlight creates an instance of a Flight suitable for spawning
+// instances on Amazon Web Services' Elastic Compute platform.
+//
+// NewFlight will consume the environment variables $AWS_REGION,
+// $AWS_ACCESS_KEY_ID, and $AWS_SECRET_ACCESS_KEY to determine the region to
+// spawn instances in and the credentials to use to authenticate.
+func NewFlight(opts *aws.Options) (platform.Flight, error) {
```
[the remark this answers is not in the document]
arithx: To add onto this comment, I'm fine with having these actions done in separate PRs after the landing of this one and am more than happy to file issues once this lands.

kola/tests/kubernetes/setup.go
```diff
@@ -73,7 +82,7 @@ func setupCluster(c cluster.TestCluster, nodes int, version, runtime string) *kC
 	}
 
 	// create worker nodes
-	workers, err := platform.NewMachines(c, conf.CloudConfig(""), nodes)
+	workers, err := platform.NewMachines(clusterWrapper{&c}, conf.CloudConfig(""), nodes)
```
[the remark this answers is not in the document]
bgilbert: See abea610c2036f2716a778f83f55ab2e36d17256f.

platform/local/cluster.go
```diff
@@ -85,6 +90,14 @@ func (lc *LocalCluster) GetDiscoveryURL(size int) (string, error) {
 	return baseURL, nil
 }
 
+func (lc *LocalCluster) GetOmahaHostPort() (string, error) {
```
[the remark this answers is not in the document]
bgilbert: Package `net` calls that string a `hostport`.  I didn't call it `GetOmahaHostAddr` to avoid confusion with `net.Addr`.

kola/tests/kubernetes/setup.go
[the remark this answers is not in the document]
arithx: Ah missed that commit message.

kola/harness.go
```diff
@@ -391,13 +390,6 @@ func parseCLVersion(input string) (*semver.Version, error) {
 func runTest(h *harness.H, t *register.Test, pltfrm string, flight platform.Flight) {
 	h.Parallel()
 
-	// don't go too fast, in case we're talking to a rate limiting api like AWS EC2.
```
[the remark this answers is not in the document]
bgilbert: Apparently rate-limiting doesn't work the way I thought it did.  The AWS bindings automatically retry rate-limited requests, but from a quick grep I think the GCE, DO, and Packet bindings just pass through rate limit errors.  So maybe this change should be dropped?

platform/machine/qemu/disk.go
bgilbert: `setupDisk` was closing the file immediately, not from a `defer`.

platform/machine/aws/flight.go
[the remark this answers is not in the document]
bgilbert: Hmm, that would require some refactoring of `platform/api/*`, e.g. to have a setup function returning a state struct which is then passed to the instance-creation function.  (Or instance creation could become a method on the state struct.)  Makes sense, but probably for a separate PR.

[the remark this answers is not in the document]
arithx: I hadn't thought of creating a `State` struct that has a built-in method for instance-creation but I do like the idea. My initial thought was around the caching of global / entire test run variables with simple accessor methods.

Once this PR lands I'll file a couple of issues for the different clouds that have distinct overarching state.

kola/harness.go
```diff
@@ -382,7 +388,7 @@ func parseCLVersion(input string) (*semver.Version, error) {
 // runTest is a harness for running a single test.
 // outputDir is where various test logs and data will be written for
 // analysis after the test run. It should already exist.
-func runTest(h *harness.H, t *register.Test, pltfrm string) {
+func runTest(h *harness.H, t *register.Test, pltfrm string, flight platform.Flight) {
```
arithx: What do you think about moving the `architecture` function in this file into `Flight`? The determination could be handled at it's instantiation which would also allow us to drop passing the `pltfrm` string here (and add an easy way to access the platform architecture from inside of tests in the future).

platform/local/flight.go
```diff
@@ -35,6 +36,10 @@ const (
 type LocalFlight struct {
 	destructor.MultiDestructor
 	*platform.BaseFlight
+	Dnsmasq    *Dnsmasq
+	SimpleEtcd *SimpleEtcd
```
arithx: I'm a bit weary about having a shared `etcd` instance across the entire `Flight`.

One of the RFE's I've had in mind to implement for a while is #799 which could potentially encounter some problems with a singular shared `etcd` instance on some tests (the fleet test comes to mind).

platform/flight.go
```diff
@@ -16,13 +16,17 @@
 package platform
 
 import (
+	"fmt"
 	"sync"
+
+	"github.com/satori/go.uuid"
```
arithx: Ignition has switched over to the [pborman/uuid](https://github.com/pborman/uuid) repo, should we switch here as well? See https://github.com/coreos/ignition/pull/560#issuecomment-398147683 for more info

platform/machine/qemu/disk.go
```diff
@@ -195,3 +193,12 @@ func setupDisk(additionalOptions ...string) (*os.File, error) {
 
 	return os.OpenFile(dstFileName, os.O_RDWR, 0)
 }
+
+func mkpath(basedir string) (string, error) {
```
arithx: Given this function isn't exported or used anywhere else should we just drop the `basedir` parameter and revisit it if we decide to allow the setting of `TempDir` locations later?

bgilbert: The next commit uses the argument.

platform/local/flight.go
bgilbert: Unless I'm missing something, this etcd is only used for the discovery service.

platform/flight.go
bgilbert: Probably, but `go.uuid` is used in a few places so that seems like a different PR.

kola/harness.go
bgilbert: `architecture` is also called from `filterTests`, which doesn't have a flight.  I agree that the arch stuff needs to be cleaned up, especially now that we can spell it `architecture` instead of `board`, but I think that's a larger project.

platform/machine/qemu/disk.go
arithx: agh, grepping around fails me again :(

kola/harness.go
```diff
@@ -278,9 +278,15 @@ func RunTests(pattern, pltfrm, outputDir string) error {
 		torcxManifestFile.Close()
 	}
 
+	flight, err := NewFlight(pltfrm)
+	if err != nil {
+		plog.Fatalf("Flight failed: %v", err)
```
ajeddeloh: nit: This is `"new flight: %v"` some places. Maybe standardize unless there's a reason not to.

bgilbert: There's a pretty good mix of message formats within `cmd/kola`, and I used the local convention in each case.  I'll leave for now.

PR #944, opened 2018-11-08
ore: Introspect disk for size rather than hardcode 8GB

cgwalters:
We want to expand the disk size of Red Hat CoreOS.  Use the JSON
output of `qemu-img info` to get the size of the VMDK, and pass
that on to AWS.

The hardcoded 8GB moves to Plume.  I think to support larger disks
there then we'd need to query the snapshot or so?

### DIFF
--- a/cmd/ore/aws/upload.go
+++ b/cmd/ore/aws/upload.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
@@ -122,6 +123,31 @@ func defaultBucketURL(urlPrefix, imageName, board, file, region string) (*url.UR
 	return s3URL, nil
 }
 
+type qemuInfo struct {
+	VirtualSize uint64 `json:"virtual-size"`
+}
+
+func getDiskSizeGiB(path string) (uint32, error) {
+	out, err := exec.Command("qemu-img", "info", "--output=json", path).Output()
+	if err != nil {
+		return 0, err
+	}
+
+	var info qemuInfo
+	err = json.Unmarshal(out, &info)
+	if err != nil {
+		return 0, err
+	}
+
+	const GiB = 1024 * 1024 * 1024
+	quotient, remainder := info.VirtualSize/GiB, info.VirtualSize%GiB
+	// Round up if there's leftover
+	if remainder > 0 {
+		quotient = quotient + 1
+	}
+	return uint32(quotient), nil
+}
+
 func runUpload(cmd *cobra.Command, args []string) error {
 	if len(args) != 0 {
 		fmt.Fprintf(os.Stderr, "Unrecognized args in aws upload cmd: %v\n", args)
@@ -131,6 +157,13 @@ func runUpload(cmd *cobra.Command, args []string) error {
 		fmt.Fprintf(os.Stderr, "At most one of --source-object and --source-snapshot may be specified.\n")
 		os.Exit(2)
 	}
+
+	imageSizeGiB, err := getDiskSizeGiB(uploadFile)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Unable to query size of disk: %v\n", err)
+		os.Exit(1)
+	}
+	plog.Debugf("Image size: %v\n", imageSizeGiB)
 
 	// if an image name is unspecified try to use version.txt
 	imageName := uploadImageName
@@ -155,7 +188,6 @@ func runUpload(cmd *cobra.Command, args []string) error {
 	}
 
 	var s3URL *url.URL
-	var err error
 	if uploadSourceObject != "" {
 		s3URL, err = url.Parse(uploadSourceObject)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/ore/aws/upload.go
```diff
@@ -122,6 +123,26 @@ func defaultBucketURL(urlPrefix, imageName, board, file, region string) (*url.UR
...
+	if err != nil {
+		return 0, err
+	}
+
+	var info qemuInfo
+	err = json.Unmarshal(out, &info)
+	if err != nil {
+		return 0, err
+	}
+	return uint32(info.VirtualSize / (uint64(1024) * 1024 * 1024)), nil
```
ajeddeloh: This will round down, is that intended?

cgwalters: I thought about that, but right now the disk images we're providing are exact multiples of GiB.  We could make it a fatal error to have it not be exact?

I am not clear what the semantics are here - presumably if it wasn't a GiB multiple one would need to round up?

Although now that I actually look at [the docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_EbsBlockDevice.html) I notice, yeah we'd need to round up but also:

> Default: If you're creating the volume from a snapshot and don't specify a volume size, the default is the snapshot size. 

Which makes sense...so why are we specifying it in the first place?  A little git blame later:

https://github.com/coreos/mantle/commit/f0f298ad37003956ffb9abbae28bb48e956188ed

"The default if nothing is set is 5GB" is confusing - were CL disk images 5GB at that time and he wanted AWS to be larger?

bgilbert: CL images are 4.5 GB right now, so this is (a) broken, and (b) a behavior change.

cmd/ore/aws/upload.go
```diff
@@ -122,6 +123,26 @@ func defaultBucketURL(urlPrefix, imageName, board, file, region string) (*url.UR
...
 }
 
+type qemuInfo struct {
+	VirtualSize uint64 `json:"virtual-size"`
+}
+
+// We used to hardcode 8GB for Container Linux, but let's support
+// differently-sized images.
+func getVMDKSizeGB(path string) (uint32, error) {
+	out, err := exec.Command("qemu-img", "info", "--output=json", path).Output()
```
bgilbert: This is called unconditionally for both VMDK and raw images and will work correctly for both, but the function name implies otherwise.

cmd/ore/aws/upload.go
```diff
@@ -122,6 +123,26 @@ func defaultBucketURL(urlPrefix, imageName, board, file, region string) (*url.UR
 	return s3URL, nil
 }
 
+type qemuInfo struct {
+	VirtualSize uint64 `json:"virtual-size"`
+}
+
+// We used to hardcode 8GB for Container Linux, but let's support
```
bgilbert: Nit: history is what `git` is for.  :grin:

cmd/ore/aws/upload.go
cgwalters: Ah.  Well that explains the "5GB" bit probably.  But then where did 8GB come from?

cmd/ore/aws/upload.go
cgwalters: Yeah, though I find it nice to avoid people needing to take that trip; when functions get moved around you need to traverse multiple layers of history, etc.

cgwalters:
Updated for comments, also tried to consistently use `GiB` here since it's important.

I didn't test it though beyond compilation yet.

cmd/ore/aws/upload.go
bgilbert: We create 8 GB images in AWS, so that the root FS has a reasonable amount of free space.

For compatibility, maybe use max(8 GB, actual image size)?

cgwalters: > We create 8 GB images in AWS, so that the root FS has a reasonable amount of free space.

Only for AWS?  Like Azure and other clouds are still 5GB? 

> For compatibility, maybe use max(8 GB, actual image size)?

Hmm, if I do that unconditionally then this entire PR would be pointless since I want to grow RHCOS beyond 8GB, right?

It seems like we'd need a new `--im-not-cl` flag?  Or maybe we only clamp if `amiName.contains("Container-Linux")` ?

cgwalters: Sorry, I was confused...yeah, we can do `max`.

cgwalters: TFW one finds https://github.com/golang/go/issues/20775

Oh well, these are small integers, we can be confident we're not going to hit FP accuracy issues.

cgwalters:
FWIW we're currently carrying this patch for RHCOS which also implies we have a forked coreos-assembler which is going to be annoying to maintain; I can retest it, any other code comments?

platform/api/aws/images.go
```diff
@@ -36,6 +36,13 @@ var (
 
 type EC2ImageType string
 
+// In GiB. This used to be hardcoded in ore.  See discussion in
```
bgilbert: It was hardcoded in this file, actually.  I'm still in favor of dropping that sentence, but :man_shrugging:

cgwalters: Updated to tweak the comment :arrow_down:

cgwalters:
Also I did test this latest PR, but only with RHCOS, not CL yet.  I *think* it'll work - if not we'll find out on the next nightly job right?

bgilbert:
We'll find out on the plume side if we check the uploaded image by hand.  We don't run ore automatically.  You could [download an AWS image](https://alpha.release.core-os.net/amd64-usr/current/) for testing ore.

cgwalters:
> You could download an AWS image for testing ore.

I did that and uploaded it and it failed, and I didn't have time to debug it in depth.  Turns out the problem is it's a `raw` image and not `vmdk`, which I found surprising.  Converting it to vmdk before uploading works for me.

bgilbert:
Ah, yeah, for historical reasons there's a separate VMDK image, `ami_vmdk`.

Good to merge?

cgwalters:
good to merge

cgwalters:
This was reverted in https://github.com/coreos/mantle/pull/948

PR #948, opened 2018-11-17
Revert "ore: Introspect disk for size rather than hardcode 8GB"

bgilbert:
It breaks all cases that don't include an explicit `--file` argument, including `--source-object`, `--source-snapshot`, and using the latest build from the CL SDK.  Revert for now.

This reverts commit 8244c5df6feafb06bc21b7e96ceb9d8371861f95.

cc @cgwalters

### DIFF
--- a/cmd/ore/aws/upload.go
+++ b/cmd/ore/aws/upload.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


cgwalters:
LGTM, will look at fixing this in a next attempt.

cgwalters:
> bgilbert:8gb-is-defined-to-be-enough-for-everyone

:laughing:  I see you noticed my branch name...

OK so what was our thought on fixing this?  We could only do it if `--file` is provided?

bgilbert:
I think it'd be cleanest for the size to be the larger of:

- A new command-line argument specifying the desired AMI disk size, defaulting to 8 GiB to handle the historical case.
- The disk size from the image file (from `--file` or autodetected), object (from `--source-object`), or snapshot (from `--source-snapshot`).

Unfortunately that requires a nontrivial amount of new code.

cgwalters:
> The disk size from the image file (from --file or autodetected), object (from --source-object), or snapshot (from --source-snapshot).

One thing I don't quite understand is the use cases for the latter two.  Are they used today in CL?

The flow we have for RHCOS at least today is to upload via a file to a single region, get an AMI, then use AWS' builtin region-copy API (which is what plume is doing I believe).  So at least we don't need object/snapshot bits (we always pass `--delete-object`).

How about `--image-inspect-size` or so which would be used with `--file`?

bgilbert:
> Are they used today in CL?

Remember, ore is a development tool, not a production tool.  I don't think anyone actively uses them, but they might be useful for working with images in various stages of creation.

> How about --image-inspect-size or so which would be used with --file?

I'm not thrilled with lack of consistency across image sources, but I can live with it.  We should still have the manual option though:

- `--image-size`, defaulting to 8 GiB
- `--image-inspect-size`, which works with `--file` (or autodetected local files) and overrides `--image-size`

PR #949, opened 2018-11-17
platform/qemu: don't try to mangle qcow disk images

bgilbert:
Fixes `kola spawn` on RHCOS images without `-b rhcos`, which isn't otherwise necessary.

cc @miabbott


cgwalters:
LGTM

PR #951, opened 2018-11-19
kola/tests: fix rhcos.tls.fetch-urls namespace

bgilbert:


arithx:
That test currently only targets `rhcos`, should it also be targeting CL or FCOS?

bgilbert:
The test currently targets everything but `rhcos`:

https://github.com/coreos/mantle/blob/4c2089262a132642634dfcc0667f8b80940241a8/kola/tests/misc/tls.go#L24

arithx:
Ah yeah, misread the flag.

PR #954, opened 2018-11-29
ore/upload: Add --size-gib and --size-inspect

cgwalters:
We want to expand the disk size of Red Hat CoreOS; previously
there was a hardcoded 8GiB size for Container Linux which
(I believe entirely coincidentally) matched RHCOS.

See https://github.com/coreos/mantle/pull/948
and https://github.com/coreos/mantle/pull/944

### DIFF
--- a/cmd/ore/aws/upload.go
+++ b/cmd/ore/aws/upload.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/coreos/mantle/platform/api/aws"
 	"github.com/coreos/mantle/sdk"
+	"github.com/coreos/mantle/util"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


cgwalters:
I have only compile-tested this; let me know if the design and overall approach seems OK and if so I'll do some manual tests.

cgwalters:
Having a forked mantle for this is still a big pain.  It's going to hold up other work like adding tests in kola.   I still find myself doubting the path of keeping mantle git master working for both CL and {F,RH}COS.  If we did a `cl` branch it would relieve us of the worry about trying to keep both working for every patch.

cgwalters:
(Today the mantle fork implies a coreos-assembler fork which in turn is causing more pain around things like dealing with the [firstboot rename](https://github.com/coreos/coreos-assembler/pull/220))

bgilbert:
Obsoleted by #959.

PR #959, opened 2018-12-13
ore/aws/upload: --disk-size-gib and --disk-size-inspect, revisited

bgilbert:
We were uploading the full disk image (4.5 GiB for CL) but hardcoding an 8 GiB disk size in AMI metadata.  This worked until RHCOS images became larger than 8 GiB.  Continue defaulting `ore aws upload` and plume to an 8 GiB AMI disk, but allow overriding from the ore command line, and add an ore option to autodetect the size from a local image file.

Based on, and obsoletes, https://github.com/coreos/mantle/pull/954.  Tested with a CL image and:

- `ore aws upload`, autodetected file
- `ore aws upload --disk-size-gib`, autodetected file
- `ore aws upload --disk-size-inspect`, autodetected file
- `ore aws upload --disk-size-inspect --file`
- `ore aws upload --disk-size-inspect --source-object` (fails)
- `ore aws upload --disk-size-inspect --source-snapshot` (fails)
- `bin/plume pre-release --platform aws -C user`

cc @cgwalters for review (GitHub isn't letting me request one from you directly)

### DIFF
--- a/cmd/ore/aws/upload.go
+++ b/cmd/ore/aws/upload.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/coreos/mantle/platform/api/aws"
 	"github.com/coreos/mantle/sdk"
+	"github.com/coreos/mantle/util"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


cgwalters:
LGTM!  Tested and works here.

PR #964, opened 2019-01-04
platform/machine: only run DeleteKey if one has been added

arithx:
Currently on both AWS & DO the `DeleteKey` API methods will be executed
regardless of if a key has been added. This will cause the platform
level Flight `Destroy` method to fail before entering the BaseFlight's
`Destroy`.


ajeddeloh:
This LGTM, but I almost wonder if destroy should actually have a list of functions that get registered with it that it calls on Destroy(). Then you can wait until the key was created before adding the callback. Thoughts?

arithx:
@ajeddeloh For certain platforms that could have more clean up operations I could see that being a more convenient pattern. At the moment these are all modifications to individual platform level flights that only really have the SSH key being added.

bgilbert:
@ajeddeloh You mean a [`MultiDestructor`](https://github.com/coreos/mantle/blob/2367cfff2d7d555f80a9d84cf81c2d2a2ce70921/lang/destructor/destructor.go#L43-L58)?  :smile:

platform/machine/aws/flight.go
```diff
@@ -68,6 +69,8 @@ func NewFlight(opts *aws.Options) (platform.Flight, error) {
 		return nil, err
 	}
 
+	af.keyAdded = true
```
bgilbert: Please snug this up in the same block as the `AddKey` call.  As is, it's too easy for another code block to be inserted between them.

platform/machine/do/flight.go
```diff
@@ -108,12 +111,14 @@ func (df *flight) NewCluster(rconf *platform.RuntimeConfig) (platform.Cluster, e
 }
 
 func (df *flight) Destroy() {
-	for _, keyID := range []int{df.sshKeyID, df.fakeSSHKeyID} {
-		if keyID == 0 {
-			continue
-		}
-		if err := df.api.DeleteKey(context.TODO(), keyID); err != nil {
-			plog.Errorf("Error deleting key %v: %v", keyID, err)
+	if df.keyAdded {
```
bgilbert: Unneeded for DO, since `sshKeyID == 0` or `fakeSSHKeyID == 0` serve that purpose.

arithx:
Updated ⬆️

Issue #572, opened 2017-04-19
kola: Mechanism for flagging tests that need Internet access


bgilbert:
Right now we're tagging these with `ExcludePlatform: []string{"qemu"}`, which is not ideal.

euank:
Ideally, tests wouldn't need internet access either.

A lot of the tests needing internet access really need access to a docker registry, which the test framework should spin up (just as it does for an etcd cluster).

The registry could be populated from known images stored as artifacts in GCS

PR #967, opened 2019-01-15
kola: add RequiresInternetAccess register flag

rfairley:
Tests with this flag set will not be run when running on a platform
that does not provide the networking setup required for Internet.
Currently, the only platform where this is the case is `"qemu"`.

This avoids needing to manually exclude platforms using
`ExcludePlatforms:` for this case.

Fixes: #572 

---

A few notes:

- I looked at the tests that were already excluding `"qemu"`. For the tests that did this, I replaced the `ExcludePlatforms: []string{"qemu"}` with the internet flag, apart from the following:

    - coreos.ignition.v2_2.security.tls (kola/tests/ignition/security.go)
    - coreos.ignition.v1.ssh.key (kola/tests/ignition/ssh.go)
    - coreos.ignition.v2.ssh.key (kola/tests/ignition/ssh.go)

    These tests look to exclude QEMU for reasons other than requiring Internet access. (security/TLS flakes (https://github.com/coreos/mantle/commit/3957a903613c724eef87506564102ccf5298893a), and SSH tests above not applicable for QEMU).

- Added a debug log message to indicate that the test was excluded because Internet is not available. Think it could be helpful, but can understand if having that message would be noisier than we'd like.

- Tested this running `kola run` on a built QEMU image of CL (built in the CL SDK version 1995.0.0).

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -192,6 +192,20 @@ func filterTests(tests map[string]*register.Test, pattern, platform string, vers
 			continue
 		}
 
+		existsIn := func(item string, entries []string) bool {
+			for _, i := range entries {
+				if i == item {
+					return true
+				}
+			}
+			return false
+		}
+
+		if existsIn(platform, register.PlatformsNoInternet) && t.HasFlag(register.RequiresInternetAccess) {
+			plog.Debugf("skipping test %s: Internet required but not supported by platform %s", t.Name, platform)
+			continue
+		}
+
 		isAllowed := func(item string, include, exclude []string) bool {
 			allowed := true
 			for _, i := range include {
--- a/kola/tests/ignition/empty.go
+++ b/kola/tests/ignition/empty.go
@@ -21,7 +21,8 @@ import (
 )
 
 // These tests require the kola key to be passed to the instance via cloud
-// provider metadata since it will not be injected into the config.
+// provider metadata since it will not be injected into the config. Platforms
+// where the cloud provider metadata system is not available have been excluded.
 func init() {
 	// Tests for https://github.com/coreos/bugs/issues/1184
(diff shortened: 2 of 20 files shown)


rfairley:
↑ Force pushed a change I committed but didn't push earlier

kola/harness.go
```diff
@@ -192,6 +192,11 @@ func filterTests(tests map[string]*register.Test, pattern, platform string, vers
 			continue
 		}
 
+		if platform == "qemu" && t.HasFlag(register.RequiresInternetAccess) {
```
arithx: Generally I'd rather have a list of platforms that do not have the feature and then querying if the platform is in the list. However, I don't think we'll be adding any new platforms that don't have internet access so I don't think it will matter.

rfairley: Thanks, SGTM to have it a list. Will be clearer with it being named something like `platformsNoInternet`. Will make that edit in

rfairley:
↑ Added the list `register.PlatformsNoInternet` in, and gave a run to make sure the right tests are still excluded. Let me know if anything needs adjusting.

kola/tests/ignition/empty.go
```diff
@@ -28,7 +28,8 @@ func init() {
 		Name:             "cl.ignition.misc.empty",
 		Run:              empty,
 		ClusterSize:      1,
-		ExcludePlatforms: []string{"qemu", "esx"},
+		Flags:            []register.Flag{register.RequiresInternetAccess},
```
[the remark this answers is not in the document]
rfairley: Amended this - added a note (alongside the existing comment)

rfairley:
Good to merge from my end.

PR #968, opened 2019-01-31
*: add unprivileged-qemu platform

arithx:
Adds a new platform `unprivileged-qemu` which is meant to be ran without
root access. It has a restricted set of functionality, such as a lack of
the Local cluster and no networking between machines.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -30,7 +30,7 @@ var (
 	kolaPlatform       string
 	defaultTargetBoard = sdk.DefaultBoard()
 	kolaArchitectures  = []string{"amd64"}
-	kolaPlatforms      = []string{"aws", "do", "esx", "gce", "openstack", "packet", "qemu"}
+	kolaPlatforms      = []string{"aws", "do", "esx", "gce", "openstack", "packet", "qemu", "qemu-unpriv"}
 	kolaDistros        = []string{"cl", "fcos", "rhcos"}
 	kolaDefaultImages  = map[string]string{
 		"amd64-usr": sdk.BuildRoot() + "/images/amd64-usr/latest/coreos_production_image.bin",
--- a/cmd/kola/spawn.go
+++ b/cmd/kola/spawn.go
@@ -174,7 +174,7 @@ func doSpawn(cmd *cobra.Command, args []string) error {
 				return fmt.Errorf("Could not read machine options: %v", err)
 			}
 
-			var machineOpts qemu.MachineOptions
+			var machineOpts platform.MachineOptions
 			err = json.Unmarshal(b, &machineOpts)
 			if err != nil {
 				return fmt.Errorf("Could not unmarshal machine options: %v", err)
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -47,6 +47,7 @@ import (
 	"github.com/coreos/mantle/platform/machine/openstack"
 	"github.com/coreos/mantle/platform/machine/packet"
 	"github.com/coreos/mantle/platform/machine/qemu"
+	"github.com/coreos/mantle/platform/machine/unprivqemu"
 	"github.com/coreos/mantle/system"
 )
 
@@ -170,14 +171,23 @@ func NewFlight(pltfrm string) (flight platform.Flight, err error) {
 		flight, err = packet.NewFlight(&PacketOptions)
 	case "qemu":
 		flight, err = qemu.NewFlight(&QEMUOptions)
+	case "qemu-unpriv":
+		flight, err = unprivqemu.NewFlight(&QEMUOptions)
 	default:
 		err = fmt.Errorf("invalid platform %q", pltfrm)
 	}
 	return
 }
 
-func filterTests(tests map[string]*register.Test, pattern, platform string, version semver.Version) (map[string]*register.Test, error) {
+func filterTests(tests map[string]*register.Test, pattern, pltfrm string, version semver.Version) (map[string]*register.Test, error) {
 	r := make(map[string]*register.Test)
+
+	checkPlatforms := []string{pltfrm}
+
+	// qemu-unpriv has the same restrictions as QEMU but might also want additional restrictions due to the lack of a Local cluster
+	if pltfrm == "qemu-unpriv" {
+		checkPlatforms = append(checkPlatforms, "qemu")
+	}
 
 	for name, t := range tests {
 		match, err := filepath.Match(pattern, t.Name)
@@ -202,13 +212,13 @@ func filterTests(tests map[string]*register.Test, pattern, platform string, vers
 			return false
 		}
 
-		if existsIn(platform, register.PlatformsNoInternet) && t.HasFlag(register.RequiresInternetAccess) {
-			plog.Infof("skipping test %s: Internet required but not supported by platform %s", t.Name, platform)
+		if existsIn(pltfrm, register.PlatformsNoInternet) && t.HasFlag(register.RequiresInternetAccess) {
+			plog.Debugf("skipping test %s: Internet required but not supported by platform %s", t.Name, pltfrm)
 			continue
 		}
 
-		isAllowed := func(item string, include, exclude []string) bool {
-			allowed := true
+		isAllowed := func(item string, include, exclude []string) (bool, bool) {
+			allowed, excluded := true, false
 			for _, i := range include {
 				if i == item {
 					allowed = true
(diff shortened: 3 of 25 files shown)


arithx:
Removing WIP, ready for review.

From offline discussions, once this is merged we should open up an issue for refactoring the QEMU based platforms. A refactor would likely include a new Command interface that supports both namespaced & non-namespaced command sets as well as some form of a base QEMU cluster struct / interface that has methods which can be overwritten to support both clusters requirements (Local cluster, networking differences, etc.)

platform/machine/unprivqemu/cluster.go
```diff
@@ -0,0 +1,197 @@
...
+	cmd := qm.qemu.(*exec.ExecCmd)
+	cmd.Stderr = os.Stderr
+
+	cmd.ExtraFiles = append(cmd.ExtraFiles, extraFiles...)
+
+	if err = qm.qemu.Start(); err != nil {
+		return nil, err
+	}
+
+	// parse /proc/net/tcp to determine the port selected by QEMU
```
cgwalters: Eeek.  Ideally there'd be a way for us to bind the socket and pass it to qemu...may require a qemu patch though.

Or maybe the port is exposed over QMP?

arithx: A specific port can be selected and given in the `hostfwd` parameter, but it seemed easier to allow QEMU's auto-selection code to be used and then determine what they were using.

As for using QMP I delved into it for a bit and while it is probably possible I didn't find an easy/clean way of doing it that way.

arithx:
Yes, `unprivileged-qemu` works on all supported distributions.

cgwalters:
I gave this a try and it seems to work for me.  Let's ship it?

cgwalters:
Hey so...what's blocking this?

cgwalters:
I don't have merge rights to mantle it looks like.  Any objections to adding me?

From my PoV @arithx this is also good to merge.

bgilbert:
@cgwalters It's blocked on a full code review.  :smile:

platform/machine/unprivqemu/cluster.go
```diff
@@ -0,0 +1,197 @@
...
+	getAddress := func() (string, error) {
```
ajeddeloh: is there a reason for this being a closure instead of a normal function?

cmd/kola/spawn.go
```diff
@@ -174,7 +174,7 @@ func doSpawn(cmd *cobra.Command, args []string) error {
 				return fmt.Errorf("Could not read machine options: %v", err)
 			}
 
-			var machineOpts qemu.MachineOptions
+			var machineOpts platform.MachineOptions
```
ajeddeloh: I think we should find a better home for this than `platform`, though not sure if that's worth blocking this PR.

kola/harness.go
```diff
@@ -169,15 +170,24 @@ func NewFlight(pltfrm string) (flight platform.Flight, err error) {
...
 }
 
-func filterTests(tests map[string]*register.Test, pattern, platform string, version semver.Version) (map[string]*register.Test, error) {
+func filterTests(tests map[string]*register.Test, pattern, pltfrm string, version semver.Version) (map[string]*register.Test, error) {
 	r := make(map[string]*register.Test)
 
+	checkPlatforms := []string{pltfrm}
+
+	// unprivileged-qemu has the same restrictions as QEMU but might also want additional restrictions due to the lack of a Local cluster
+	if pltfrm == "unprivileged-qemu" {
```
ajeddeloh: nit for later, not this PR, we should define these as `const` strings in each platform.

kola/tests/misc/raid.go
```diff
@@ -105,12 +106,20 @@ systemd:
...
 func RootOnRaid(c cluster.TestCluster) {
+	var m platform.Machine
+	var err error
 	options := platform.MachineOptions{
 		AdditionalDisks: []platform.Disk{
 			{Size: "520M", DeviceOpts: []string{"serial=secondary"}},
 		},
 	}
-	m, err := c.Cluster.(*qemu.Cluster).NewMachineWithOptions(raidRootUserData, options)
+	if qc, ok := c.Cluster.(*qemu.Cluster); ok {
```
ajeddeloh: nit: could use a type switch instead.

platform/machine/unprivqemu/cluster.go
arithx: No particular reason, was originally just in-function code but ended up needing to be inside of a sub-function for `util.Retry` so I just wrapped it into a closure rather than creating a separate function and passing through `qm.qemu.Pid()`

cmd/kola/spawn.go
arithx: If I recall correctly I originally had it in a `platform/qemu` sub-package but I don't remember why I changed it back to just being in `platform`. I wouldn't personally be opposed to either having it in a sub-package or leaving it inside of `platform`.

@bgilbert did you have a particularly strong opinion either way?

platform/machine/unprivqemu/cluster.go
ajeddeloh: I'd rather have it as a function in that case, make it explicit we're not trying to capture variables or anything fancy.

cmd/kola/spawn.go
bgilbert: `platform` seems like the obvious place for it.  I don't think it's important to start a subpackage just for these structs.  I'm okay with other approaches though.

cmd/kola/options.go
```diff
@@ -30,7 +30,7 @@ var (
 	kolaPlatform       string
 	defaultTargetBoard = sdk.DefaultBoard()
 	kolaArchitectures  = []string{"amd64"}
-	kolaPlatforms      = []string{"aws", "do", "esx", "gce", "openstack", "packet", "qemu"}
+	kolaPlatforms      = []string{"aws", "do", "esx", "gce", "openstack", "packet", "qemu", "unprivileged-qemu"}
```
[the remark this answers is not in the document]
arithx: `qemu-unpriv` works for me, I don't have any better ideas.

cmd/kola/spawn.go
arithx: I'll leave it in it's current place for this PR and we can decide later if we want to move it.

arithx:
Update pushed addressing comments

platform/qemu.go
```diff
@@ -214,3 +219,102 @@ func mkpath(basedir string) (string, error) {
 	defer f.Close()
 	return f.Name(), nil
 }
+
+func CreateQEMUCommand(board, uuid, biosImage, consolePath, confPath, diskImagePath string, isIgnition bool, options MachineOptions) ([]string, []*os.File, error) {
+	var qmCmd []string
+	combo := runtime.GOARCH + "--" + board
+	switch combo {
+	case "amd64--amd64-usr":
```
jlebon: Aside: ahh right, we'll eventually need to expand this list for more archs, which... yeah would be good to share this with `cosa run`. Maybe for now a comment here to that effect?

kola/harness.go
```diff
@@ -169,15 +170,24 @@ func NewFlight(pltfrm string) (flight platform.Flight, err error) {
...
+	case "qemu-unpriv":
+		flight, err = unprivqemu.NewFlight(&QEMUOptions)
 	default:
 		err = fmt.Errorf("invalid platform %q", pltfrm)
 	}
 	return
 }
 
-func filterTests(tests map[string]*register.Test, pattern, platform string, version semver.Version) (map[string]*register.Test, error) {
+func filterTests(tests map[string]*register.Test, pattern, pltfrm string, version semver.Version) (map[string]*register.Test, error) {
```
jlebon: Minor: any reason for the variable rename here?

kola/tests/docker/docker.go
```diff
@@ -206,6 +212,12 @@ systemd:
   units:
    - name: docker.service
      enable: true`),
+
+		// <link to mantle bug>
```
jlebon: Missing link here?

kola/harness.go
arithx: I chose to rename it because having a variable named `platform` as well as a `platform` package used in the same file felt like it could lead to confusion; better to just short-name the variable.

platform/qemu.go
arithx: Isn't `cosa run` libvirt based? That level of abstraction should eliminate a lot of the need for this native arch + intended  arch matrix.

kola/tests/docker/docker.go
arithx: I was originally intending to post up the issue and force-push an amended commit that had the correct links but I'll post it later and update.

platform/qemu.go
jlebon: No, it uses `qemu-kvm` directly, see: https://github.com/coreos/coreos-assembler/blob/f42b840cfe1b4c3d564b1c33b91768222435da8f/src/cmd-run#L181. We already have some minor arch-specific optioneering going on there.

kola/harness.go
jlebon: Gotcha, makes sense!

platform/qemu.go
arithx: Ah, TIL, will add a comment.

arithx:
Rebased, updated per comments, and pushed

PR #977, opened 2019-02-25
sdk/enter: make compatible with Silverblue

jlebon:
On Silverblue, `$HOME` is in `/var/home`, so that throws off code that
assumes otherwise. The rest of the SDK assumes `/home` is where the user
home directory will be inside the chroot, so let's keep that assumption
going by building the path to `/home` here too rather than passing
through the original path.

### DIFF
--- a/sdk/enter.go
+++ b/sdk/enter.go
@@ -200,7 +200,8 @@ func (e *enter) MountGnupgHome() error {
 	// gpg gets confused when GNUPGHOME isn't ~/.gnupg, so mount it there.
 	// Additionally, set the GNUPGHOME variable so commands run with sudo
 	// can also use it.
-	newHome := filepath.Join(e.Chroot, e.User.HomeDir, ".gnupg")
+	newHomeInChroot := filepath.Join("/home", e.User.Username, ".gnupg")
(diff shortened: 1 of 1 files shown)


sdk/enter.go
```diff
@@ -200,7 +200,8 @@ func (e *enter) MountGnupgHome() error {
 	// gpg gets confused when GNUPGHOME isn't ~/.gnupg, so mount it there.
 	// Additionally, set the GNUPGHOME variable so commands run with sudo
 	// can also use it.
-	newHome := filepath.Join(e.Chroot, e.User.HomeDir, ".gnupg")
+	newHomeInChroot := filepath.Join("home", e.User.Username, ".gnupg")
```
dm0-: Won't this string be a relative path?  Is that a problem when using it directly for `GNUPGHOME` below?

jlebon: Yeah, good catch. I suspect I would've hit this when I reached the signing part of the release process. Fixed!

jlebon:
Hmm weird, the Travis CI test passed: https://travis-ci.org/coreos/mantle/builds/498364481), but it's still pending here.

PR #978, opened 2019-02-27
kola/tests/rpmostree/status: relax regex matching for version

miabbott:
When RHCOS changed the version numbering to something that looks like
`400.7.20190227.0`, it broke the `rpm-ostree status` test.  Example
version output from RHCOS:

`Version: 400.7.20190227.0 (2019-02-27T16:19:34Z)`

This relaxes the regex to match one or more digits, then anything
afterwards, until the first paren is hit.  Since I assume RHCOS and
FCOS are going to use numbers for their versions, this should work for
both situations.

### DIFF
--- a/kola/tests/rpmostree/status.go
+++ b/kola/tests/rpmostree/status.go
@@ -36,7 +36,7 @@ func init() {
 
 var (
 	// Regex to extract version number from "rpm-ostree status"
-	rpmOstreeVersionRegex string = `^Version: (\d+\.\d+\.?\d*).*`
+	rpmOstreeVersionRegex string = `^Version: ([\d.]+).*`
 )
 
 // rpmOstreeCleanup calls 'rpm-ostree cleanup -rpmb' on a host and verifies


ashcrow:
Just verifying the regex seems to work as expected:

```
>>> re.findall('^Version: (\d+.*) \(.*', 'Version: 400.7.20190227.0 (2019-02-27T16:19:34Z)')
['400.7.20190227.0']
```

ashcrow:
Requesting a review from @arithx

kola/tests/rpmostree/status.go
```diff
@@ -36,7 +36,7 @@ func init() {
 
 var (
 	// Regex to extract version number from "rpm-ostree status"
-	rpmOstreeVersionRegex string = `^Version: (\d+\.\d+\.?\d*).*`
+	rpmOstreeVersionRegex string = `^Version: (\d+.*) \(.*`
```
arithx: We should probably specifically call out only the allowed values.

Maybe something like:
`^Version: ([\d\.]+).*`

ashcrow: Works as expected as well:

```
>>> re.findall("^Version: ([\d\.]+).*", 'Version: 400.7.20190227.0 (2019-02-27T16:19:34Z)')
['400.7.20190227.0']
```

miabbott: I tried a few more version strings we've seen in the `ostree` world and they all seem to work, as long as they are just digits.  If we introduce a letter, the results are not expected.  I think this is fine, as I don't expect us to use any letters in version strings.

```
>>> v = re.compile("^Version: ([\d.]+).")
>>> s = "Version: 400.7.20190227.0 (2019-02-27T16:19:34Z)"
>>> v.findall(s)
['400.7.20190227.0']
>>> s = "Version: 47.334 (2019-02-27T16:19:34Z)"
>>> v.findall(s)
['47.334']
>>> s = "Version: 29.20190226.0 (2019-02-27T16:19:34Z)"
>>> v.findall(s)
['29.20190226.0']
>>> s = "Version: 29.20190226.a (2019-02-27T16:19:34Z)"
>>> v.findall(s)
['29.20190226.']
```

miabbott:
Pushed a new version based on @arithx suggestion :arrow_up:

kola/tests/rpmostree/status.go
arithx: Hmm, it seems that the POSIX brackets are removing the need to escape the `.`; I would have expected `[\d.]` to be essentially a useless character class as it includes the wildcard but it's actually matching `[0-9]` or the `.` character.

miabbott: Heh, I saw this comment after it was merged.  I changed the regex to include the escaped `.`, ala
`"^Version: ([\d\.]+).*"` and things worked the same for the test strings:

```
>>> v = re.compile("^Version: ([\d\.]+).*")
>>> s = "Version: 400.7.20190227.0 (2019-02-27T16:19:34Z)"
>>> v.findall(s)
['400.7.20190227.0']
>>> s = "Version: 47.334 (2019-02-27T16:19:34Z)"
>>> v.findall(s)
['47.334']
>>> s = "Version: 29.20190226.0 (2019-02-27T16:19:34Z)"
>>> v.findall(s)
['29.20190226.0']
>>> s = "Version: 29.20190226.a (2019-02-27T16:19:34Z)"
>>> v.findall(s)
['29.20190226.']
```

So good, either way.

arithx: Yeah, the non-escaped `.` inside of a character class definition seems to be the same as an escaped `.`, not a wild-card. TIL

PR #983, opened 2019-03-01
kola/tests/misc/selinux: refactor SELinux boolean tests

miabbott:
Previously, the SELinux boolean tests assumed a default value for the
boolean that was being tested.  When RHCOS flipped this value during
its compose, the tests broke.

This changes the tests to query the boolean first, then flip the value
of it during the test.

This introduces `toggleSelinuxBoolean()`, which accepts a SELinux
boolean to query and returns an array with the flipped values to be
used later.

### DIFF
--- a/kola/tests/misc/selinux.go
+++ b/kola/tests/misc/selinux.go
@@ -15,7 +15,9 @@
 package misc
 
 import (
+	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/coreos/mantle/kola/cluster"
 	"github.com/coreos/mantle/kola/register"
@@ -49,10 +51,20 @@ func init() {
 	})
 }
 
+// cmdCheckOutput is used by `testSelinuxCmds()`. It contains a
+// command to run, a bool indicating if output should be matched,
+// and the regex used for the match
 type cmdCheckOutput struct {
 	cmdline     string // command to run
 	checkoutput bool   // should output be checked
 	match       string // regex used to match output from command
+}
+
+// seBooleanState is used by `getSelinuxBooleanState()` to store the
+// original value of the boolean and the flipped value of the boolean
+type seBooleanState struct {
+	originalValue string // original boolean value from `getseboolean`
+	newValue      string // new, opposite boolean value
 }
 
 // testSelinuxCmds will run a list of commands, optionally check their output, and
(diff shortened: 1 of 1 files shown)


miabbott:
It's been a minute since I've had to write `golang`, so hopefully this isn't too bad.

Also, I recognize that `SelinuxBoolean()` and `SelinunxBooleanPersist()` are nearly identical now; let me know if there is a better way to approach this.

kola/tests/misc/selinux.go
```diff
@@ -101,42 +143,62 @@ func SelinuxEnforce(c cluster.TestCluster) {
...
-	cmdList := []cmdCheckOutput{
-		{"getsebool virt_use_nfs", true, ".*off"},
-		{"sudo setsebool -P virt_use_nfs 1", false, ""},
-		{"getsebool virt_use_nfs", true, ".*on"},
-	}
+	var seBoolean string = "virt_use_nfs"
 
 	m := c.Machines()[0]
 
+	newBool, err := toggleSelinuxBoolean(c, m, seBoolean)
```
ashcrow: This test looks almost identical to `SelinuxBoolean` ... but so was the previous version. Eventually it would be nice to reuse but for now :+1:

kola/tests/misc/selinux.go
```diff
@@ -75,6 +77,46 @@ func testSelinuxCmds(c cluster.TestCluster, m platform.Machine, cmds []cmdCheckO
 	}
 }
 
+// toggleSelinuxBoolean checks the original value of a provided SELinux boolean
+// and returns an array that contains the opposite values
+func toggleSelinuxBoolean(c cluster.TestCluster, m platform.Machine, seBool string) ([3]string, error) {
```
arithx: The function name seems to imply that it's performing the action of toggling the boolean, maybe `getSelinuxBooleanState`?

Also, let's make a struct for the return contents rather than returning a list of strings. Then we can have names that imply what each field should be explicitly.

kola/tests/misc/selinux.go
```diff
@@ -101,42 +143,62 @@ func SelinuxEnforce(c cluster.TestCluster) {
 
 // SelinuxBoolean checks that you can tweak a boolean in the current session
 func SelinuxBoolean(c cluster.TestCluster) {
-	cmdList := []cmdCheckOutput{
-		{"getsebool virt_use_nfs", true, ".*off"},
-		{"sudo setsebool virt_use_nfs 1", false, ""},
-		{"getsebool virt_use_nfs", true, ".*on"},
-	}
+	var seBoolean string = "virt_use_nfs"
```
arithx: No need to declare it this way, let's just do `seBoolean := "virt_use_nfs"`

miabbott:
Pushed a new version that uses a struct, etc based on @arithx suggestion :arrow_up:

kola/tests/misc/selinux.go
```diff
@@ -55,6 +57,11 @@ type cmdCheckOutput struct {
 	match       string // regex used to match output from command
 }
 
+type seBooleanState struct {
```
ashcrow: nit: Types should generally have a doc line

kola/tests/misc/selinux.go
```diff
@@ -75,6 +82,40 @@ func testSelinuxCmds(c cluster.TestCluster, m platform.Machine, cmds []cmdCheckO
 	}
 }
 
+// getSelinuxBooleanState checks the original value of a provided SELinux boolean
+// and returns a seBooleanState struct with original + opposite/new value
+func getSelinuxBooleanState(c cluster.TestCluster, m platform.Machine, seBool string) (seBooleanState, error) {
```
ashcrow: :+1:

miabbott:
Added some doc lines on the structs as requested by @ashcrow :arrow_up:

kola/tests/misc/selinux.go
```diff
@@ -101,42 +146,61 @@ func SelinuxEnforce(c cluster.TestCluster) {
...
 
+	persistBoolState, err := getSelinuxBooleanState(c, m, seBoolean)
+	if err != nil {
+		c.Fatalf(`Failed to gather SELinux boolean state: %v`, err)
+	}
+
+	// construct a regexp that looks like ".*off" or ".*on"
+	persistBoolRegexp := ".*" + persistBoolState.newValue
+	cmdList := []cmdCheckOutput{
+		{"sudo setsebool -P " + seBoolean + " " + persistBoolState.newValue, false, ""},
```
arithx: Minor: I generally prefer doing `fmt.Sprintf` for cases where there are multiple variables/strings being shoved together.

miabbott: Easy enough...will address this

miabbott:
One last version (I hope) with the `fmt.Sprintf()` change :arrow_up:

miabbott:
Good to merge

Issue #981, opened 2019-03-01
`kola spawn` doesn't work with FCOS


miabbott:
I know that FCOS isn't even ready, but putting this out there...

I built an FCOS image locally via `coreos-assembler` and tried to boot it with `kola spawn`

```
$  sudo ./bin/kola spawn -b fcos --qemu-image /var/srv/fedora-coreos/builds/latest/fedora-coreos-29-qemu.qcow2                                  
Spawning instance failed: machine "24d5b086-5fb3-45f1-af7a-8f1c169c1a04" failed to enable selinux: Unable to enable SELinux: Process exited with status 1: We trust you have received the usual lecture from the local System
Administrator. It usually boils down to these three things:

    #1) Respect the privacy of others.
    #2) Think before you type.
    #3) With great power comes great responsibility.

sudo: no tty present and no askpass program specified
```

This also affects the ability to `kola run` any tests against FCOS, obviously.

Might be a configuration issue on FCOS that we can fix in `fedora-coreos-config`, but I haven't dug in.

bgilbert:
Thanks for reporting.  From the symptoms, this is probably https://github.com/coreos/fedora-coreos-config/issues/41.  I'll close as duplicate.

ashcrow:
FWIW: We now see the same issue in RHCOS on qemu.

PR #984, opened 2019-03-01
WIP: kola/tests: Sense initial remotes before running ostree remotes tests

ashcrow:
A fresh image has 0 or more remotes. Get the initial list and use that in tests.

### DIFF
--- a/kola/tests/ostree/basic.go
+++ b/kola/tests/ostree/basic.go
@@ -54,6 +54,20 @@ type ostreeAdminStatus struct {
 	Version  string
 }
 
+// getOstreeRemotes returns the current number of ostree remotes on a machine
+func getOstreeRemotes(c cluster.TestCluster, m platform.Machine) (int, []string) {
+	remoteListOut := string(c.MustSSH(m, "ostree remote list"))
+	numRemotes := 0
+	// If we get anything other than an empty string calculate the results
+	// NOTE: This is needed as splitting "" ends up providing a count of 1
+	//       when the count should be 0
+	remoteListRaw := strings.Split(remoteListOut, "\n")
+	if remoteListOut != "" {
+		numRemotes = len(remoteListRaw)
+	}
+	return numRemotes, remoteListRaw
+}
+
 // getOstreeAdminStatus stuffs the important output of `ostree admin status`
 // into an `ostreeAdminStatus` struct
 func getOstreeAdminStatus(c cluster.TestCluster, m platform.Machine) (ostreeAdminStatus, error) {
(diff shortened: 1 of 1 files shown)


kola/tests/ostree/basic.go
```diff
@@ -195,7 +195,7 @@ func ostreeRemoteTest(c cluster.TestCluster) {
 
 		osRemoteListSplit := strings.Split(string(osRemoteListOut), "\n")
 		// should have original remote + newly added remote
-		if len(osRemoteListSplit) != 2 {
+		if len(osRemoteListSplit) != 1 {
```
arithx: The above comment seems to indicate that there should also be the original remote in addition to the newly added remote. Also, is this tested against FCOS?

cc @miabbott

ashcrow: No it's not tested against FCOS  .... good call.

miabbott: RHCOS has no remote nowadays, but FCOS likely will.

FCOS tests are going to be a non-starter right now #981 

You should still be able to inspect what the state of FCOS is like with `coreos-assembler run`, though.

ashcrow: I'm downloading a copy right now to inspect.

arithx: We'll probably have to split the test then to accommodate both distros, e.x.: `rhcos.ostree.remote` which calls a RHCOS specific function that calls `ostreeRemoteTest` passing in the expected amount of remotes existing in the base image.

miabbott: My locally built FCOS shows:

```
$ rpm-ostree status
State: idle
Warning: failed to query journal: Cannot assign requested address (os error 99)
AutomaticUpdates: disabled
Deployments:
● ostree://fedora-coreos:fedora/29/x86_64/coreos
                   Version: 29 (2019-02-28T22:02:19Z)
                    Commit: d18e5962c1771c0eb3afccf694d61cce1ebafb26ae50104dee9e3284cbce2b17
```

ashcrow:
Re-looking at this before my next meeting

ashcrow:
Updated.

kola/tests/ostree/basic.go
```diff
@@ -277,16 +293,23 @@ func ostreeRemoteTest(c cluster.TestCluster) {
...
+
 		if preNumRemotes < 1 {
 			c.Fatalf(`No remotes configured on host: %q`, string(preRemotesOut))
 		}
 
 		c.MustSSH(m, ("sudo ostree remote delete " + remoteName))
 
-		delRemoteListOut := c.MustSSH(m, "ostree remote list")
-		delNumRemotes := len(strings.Split(string(delRemoteListOut), "\n"))
+		delRemoteListOut := string(c.MustSSH(m, "ostree remote list"))
```
arithx: Why not use `getOstreeRemotes` for initializing this variable & `preNumRemotes`?

ashcrow: Only because of the failure message printing out the raw output. I can update `getOstreeRemotes` to return the output and the number and re-use it. Will update shortly.

ashcrow: Updated

ashcrow:
Merge at will

PR #985, opened 2019-03-01
kola/tests/podman: update output validation

mike-nguyen:
Podman has some minor formatting changes in its command output:
  - displays only 12 characters of container IDs
  - id/image/status changed to ID/Image/Status
  - more detailed statuses

This commit accounts for these changes and relaxes the check on the
status.

### DIFF
--- a/kola/tests/podman/podman.go
+++ b/kola/tests/podman/podman.go
@@ -130,7 +130,7 @@ func podmanWorkflow(c cluster.TestCluster) {
 
 		cmd = fmt.Sprintf("sudo podman run -d -p 80:80 -v %s/index.html:%s/index.html:z %s", string(dir), wwwRoot, image)
 		out := c.MustSSH(m, cmd)
-		id = string(out)
+		id = string(out)[0:12]
 
 		podIsRunning := func() error {
 			b, err := c.SSH(m, `curl -f http://localhost 2>/dev/null`)
@@ -150,7 +150,7 @@ func podmanWorkflow(c cluster.TestCluster) {
 
 	// Test: Execute command in container
 	c.Run("exec", func(c cluster.TestCluster) {
-		cmd := fmt.Sprintf("sudo podman exec %s echo hello", id[0:12])
+		cmd := fmt.Sprintf("sudo podman exec %s echo hello", id)
 		out := c.MustSSH(m, cmd)
 
 		if string(out) != "hello" {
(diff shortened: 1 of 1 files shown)


kola/tests/podman/podman.go
```diff
@@ -58,9 +58,9 @@ func init() {
 
 // simplifiedContainerPsInfo represents a container entry in podman ps -a
 type simplifiedContainerPsInfo struct {
-	ID     string `json:"id"`
-	Image  string `json:"image"`
-	Status string `json:"status"`
+	ID     string `json:"ID"`
```
arithx: golang's `json.Unmarshal` is case-insensitive (https://golang.org/pkg/encoding/json/#Unmarshal) so this isn't strictly necessary but if all versions of Podman are producing it in this format this is fine.

kola/tests/podman/podman.go
```diff
@@ -169,9 +169,9 @@ func podmanWorkflow(c cluster.TestCluster) {
 
 		found := false
 		for _, container := range psInfo.containers {
-			if container.ID == id {
+			if container.ID == id[0:12] {
```
arithx: It looks like every time `id` is used we're only using this range, should we just initially set it to that value rather than selecting a limited set? (e.x.: `id = string(out)[0:12]`)

kola/tests/podman/podman.go
```diff
@@ -212,7 +212,7 @@ func podmanWorkflow(c cluster.TestCluster) {
 		cmd = fmt.Sprintf("sudo podman images | grep %s", imageID)
 		out, err := c.SSH(m, cmd)
 		if err == nil {
-			c.Fatalf("Image should be deleted but found %s", string(out))
+			c.Fatalf("Image should be deleted but found %s", string(out)[0:12])
```
arithx: Why trim the image name?

mike-nguyen: I went on autopilot there a bit.  I'll remove it.

kola/tests/podman/podman.go
mike-nguyen: :+1: I didn't realize I changed them all

kola/tests/podman/podman.go
mike-nguyen: I changed it back to lowercase since older versions were using lowercase before.

mike-nguyen:
Changes pushed :arrow_up:

mike-nguyen:
ok to merge

PR #986, opened 2019-03-04
kola/test/util/containers.go: don't reuse layers when building images

miabbott:
By default `podman` will reuse layers when building images.  This can
cause problems when building different images using the same
`Dockerfile` as is done in `GenPodmanScratchContainer()`.


miabbott:
Thanks to @mike-nguyen who helped me debug this

mike-nguyen:
LGTM Great work finding the root cause!

miabbott:
Good to merge

PR #990, opened 2019-03-11
*: add oscontainer pivot support

### LINKED ISSUE #987: enable `kola` to test oscontainers
In the near future, the build pipeline for RHCOS will not be producing AMIs at the same cadence as `oscontainers`.  This presents a problem for using `kola` to test new RHCOS content, as we expect to have an AMI to test against.

I think we'd like to be able to provide something like a `registry/container@sha256sum` value via a flag to `kola run`, which would instruct the hosts to upgrade to that `oscontainer` before running any test.  @cgwalters suggested we use Ignition to pivot the host under test to the desired `oscontainer`

arithx:
Adds a new flag `--oscontainer` to `kola` which allows the specification
of an oscontainer image pull spec when running against the `rhcos`
distribution. When specified it will create the relevant files & the
systemd units required to perform a `pivot` operation before `sshd`
starts.

Closes #987

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -59,6 +59,9 @@ func init() {
 	ss("debug-systemd-unit", []string{}, "full-unit-name.service to enable SYSTEMD_LOG_LEVEL=debug on. Specify multiple times for multiple units.")
 	sv(&kola.UpdatePayloadFile, "update-payload", "", "Path to an update payload that should be made available to tests")
 
+	// rhcos-specific options
+	sv(&kola.Options.OSContainer, "oscontainer", "", "oscontainer image pullspec for pivot (RHCOS only)")
+
 	// aws-specific options
 	defaultRegion := os.Getenv("AWS_REGION")
 	if defaultRegion == "" {
@@ -165,6 +168,10 @@ func syncOptions() error {
 			Name:     "10-debug.conf",
 			Contents: "[Service]\nEnvironment=SYSTEMD_LOG_LEVEL=debug",
 		})
+	}
+
+	if kola.Options.OSContainer != "" && kola.Options.Distribution != "rhcos" {
+		return fmt.Errorf("oscontainer is only supported on rhcos")
 	}
 
 	return nil
--- a/kola/tests/rpmostree/deployments.go
+++ b/kola/tests/rpmostree/deployments.go
@@ -117,9 +117,9 @@ func rpmOstreeUpgradeRollback(c cluster.TestCluster) {
 			c.Fatal(err)
 		}
 
-		// should have two deployments
-		if len(postUpgradeStatus.Deployments) != 2 {
-			c.Fatalf("Expected two deployments; found %d deployments", len(postUpgradeStatus.Deployments))
+		// should have an additional deployment
+		if len(postUpgradeStatus.Deployments) != len(originalStatus.Deployments)+1 {
+			c.Fatalf("Expected %d deployments; found %d deployments", len(originalStatus.Deployments)+1, len(postUpgradeStatus.Deployments))
 		}
 
 		// reboot into new deployment
@@ -134,9 +134,9 @@ func rpmOstreeUpgradeRollback(c cluster.TestCluster) {
 			c.Fatal(err)
 		}
 
-		// should still have 2 deployments
+		// should have 2 deployments, the previously booted deployment and the test deployment due to rpm-ostree pruning
 		if len(postRebootStatus.Deployments) != 2 {
-			c.Fatalf("Expected two deployments; found %d deployment", len(postRebootStatus.Deployments))
+			c.Fatalf("Expected %d deployments; found %d deployment", 2, len(postRebootStatus.Deployments))
 		}
 
 		// origin should be new branch
@@ -176,7 +176,7 @@ func rpmOstreeUpgradeRollback(c cluster.TestCluster) {
 
 		// still 2 deployments...
 		if len(rollbackStatus.Deployments) != 2 {
-			c.Fatalf("Expected two deployments; found %d deployments", len(rollbackStatus.Deployments))
+			c.Fatalf("Expected %d deployments; found %d deployments", 2, len(rollbackStatus.Deployments))
 		}
 
 		// validate we are back to the original deployment by comparing the
@@ -291,7 +291,7 @@ func rpmOstreeInstallUninstall(c cluster.TestCluster) {
 
 		// check the metadata to make sure everything went well
 		if len(postUninstallStatus.Deployments) != 2 {
-			c.Fatal("Expected two deployments, got %d", len(postUninstallStatus.Deployments))
+			c.Fatal("Expected %d deployments, got %d", 2, len(postUninstallStatus.Deployments))
 		}
 
 		if postUninstallStatus.Deployments[0].Checksum != originalCsum {
--- a/platform/cluster.go
+++ b/platform/cluster.go
@@ -162,6 +162,28 @@ func (bc *BaseCluster) RenderUserData(userdata *conf.UserData, ignitionVars map[
 		conf.CopyKeys(keys)
 	}
 
+	if bc.bf.baseopts.OSContainer != "" {
+		if bc.Distribution() != "rhcos" {
+			return nil, fmt.Errorf("oscontainer is only supported on the rhcos distribution")
+		}
+		conf.AddSystemdUnitDropin("pivot.service", "00-before-sshd.conf", `[Unit]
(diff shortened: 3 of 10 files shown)


arithx:
Notably: both NFS related tests (`coreos.ignition.systemd.enable-service` & `linux.nfs.v4`) that are enabled on `rhcos` are broken when doing an oscontainer pivot (https://github.com/openshift/os/issues/379)

kola/tests/rpmostree/deployments.go
```diff
@@ -291,7 +293,7 @@ func rpmOstreeInstallUninstall(c cluster.TestCluster) {
 
 		// check the metadata to make sure everything went well
 		if len(postUninstallStatus.Deployments) != 2 {
-			c.Fatal("Expected two deployments, got %d", len(postUninstallStatus.Deployments))
+			c.Fatal("Expected two deployments, got %d", 2, len(postUninstallStatus.Deployments))
```
miabbott: Missing another `%d` here?

platform/conf/conf.go
```diff
@@ -292,6 +295,78 @@ func (c *Conf) Bytes() []byte {
...
+	c.ignitionV21.Storage.Files = append(c.ignitionV21.Storage.Files, v21types.File{
+		Node: v21types.Node{
+			Filesystem: filesystem,
+			Path:       path,
+		},
+		FileEmbedded1: v21types.FileEmbedded1{
+			Contents: v21types.FileContents{
+				Source: dataurl.EncodeBytes([]byte(contents)),
+			},
+			Mode: mode,
```
miabbott: Should this be `&mode`?

kola/tests/rpmostree/deployments.go
```diff
@@ -134,9 +135,10 @@ func rpmOstreeUpgradeRollback(c cluster.TestCluster) {
 			c.Fatal(err)
 		}
 
-		// should still have 2 deployments
+		// should have 2 deployments, the previously booted deployment and the test deployment due to rpm-ostree pruning
 		if len(postRebootStatus.Deployments) != 2 {
-			c.Fatalf("Expected two deployments; found %d deployment", len(postRebootStatus.Deployments))
+			platform.Manhole(m)
```
miabbott: leftover Manhole

platform/conf/conf.go
arithx: in `v2.1.0` `mode` is an `int` https://github.com/coreos/mantle/blob/master/vendor/github.com/coreos/ignition/config/v2_1/types/schema.go#L60

kola/tests/rpmostree/deployments.go
arithx: Missed removing the `2, `. Had originally changed this to be using the initial length + 1 but that didn't work because of pruning.

platform/conf/conf.go
miabbott: I figured it was a difference in the spec, but didn't know where to look.  Thanks for the pointer!

cmd/kola/options.go
```diff
@@ -27,6 +27,7 @@ import (
 
 var (
 	outputDir          string
+	oscontainer        string
```
bgilbert: Unused.

platform/cluster.go
```diff
@@ -196,6 +207,10 @@ func (bc *BaseCluster) GetDiscoveryURL(size int) (string, error) {
 	return result, err
 }
 
+func (bc *BaseCluster) OSContainer() string {
```
bgilbert: I'd be inclined to omit the accessor since nothing outside this file needs it.

platform/cluster.go
```diff
@@ -162,6 +162,17 @@ func (bc *BaseCluster) RenderUserData(userdata *conf.UserData, ignitionVars map[
 		conf.CopyKeys(keys)
 	}
 
+	if bc.OSContainer() != "" {
+		if bc.Distribution() != "rhcos" {
+			return nil, fmt.Errorf("oscontainer is only supported on the rhcos distribution")
+		}
+		conf.AddSystemdUnitDropin("pivot.service", "00-before-sshd.conf", `[Unit]
+Before=sshd.service`)
+		conf.AddSystemdUnit("pivot.service", "", true)
```
bgilbert: Missing body?

platform/cluster.go
```diff
@@ -162,6 +162,17 @@ func (bc *BaseCluster) RenderUserData(userdata *conf.UserData, ignitionVars map[
...
+		conf.AddSystemdUnit("pivot-write-reboot-needed.service", "[Unit]\nDescription=Touch /run/pivot/reboot-needed\nConditionFirstBoot=true\n\n[Service]\nType=oneshot\nExecStart=/usr/bin/mkdir /run/pivot\nExecStart=/usr/bin/touch /run/pivot/reboot-needed\n\n[Install]\nWantedBy=multi-user.target\n", true)
+		conf.AddFile("/etc/pivot/image-pullspec", "root", bc.OSContainer(), 420)
```
bgilbert: `0644` would be clearer.

platform/cluster.go
```diff
@@ -162,6 +162,17 @@ func (bc *BaseCluster) RenderUserData(userdata *conf.UserData, ignitionVars map[
...
 	}
 
+	if bc.OSContainer() != "" {
+		if bc.Distribution() != "rhcos" {
+			return nil, fmt.Errorf("oscontainer is only supported on the rhcos distribution")
+		}
+		conf.AddSystemdUnitDropin("pivot.service", "00-before-sshd.conf", `[Unit]
+Before=sshd.service`)
+		conf.AddSystemdUnit("pivot.service", "", true)
+		conf.AddSystemdUnit("pivot-write-reboot-needed.service", "[Unit]\nDescription=Touch /run/pivot/reboot-needed\nConditionFirstBoot=true\n\n[Service]\nType=oneshot\nExecStart=/usr/bin/mkdir /run/pivot\nExecStart=/usr/bin/touch /run/pivot/reboot-needed\n\n[Install]\nWantedBy=multi-user.target\n", true)
```
bgilbert: A multi-line string would be clearer.

`mkdir -p` would be more robust.

platform/conf/conf.go
```diff
@@ -292,6 +295,78 @@ func (c *Conf) Bytes() []byte {
 	return []byte(c.String())
 }
 
+func (c *Conf) addFileV2(path, filesystem, contents string, mode int) {
+	u, err := url.Parse(dataurl.EncodeBytes([]byte(contents)))
+	if err != nil {
+		return
```
bgilbert: We should at least log the error.

platform/conf/conf.go
```diff
@@ -292,6 +295,78 @@ func (c *Conf) Bytes() []byte {
...
+func (c *Conf) AddFile(path, filesystem, contents string, mode int) {
+	if c.ignitionV2 != nil {
+		c.addFileV2(path, filesystem, contents, mode)
+	} else if c.ignitionV21 != nil {
+		c.addFileV21(path, filesystem, contents, mode)
+	} else if c.ignitionV22 != nil {
+		c.addFileV22(path, filesystem, contents, mode)
+	} else if c.ignitionV23 != nil {
+		c.addFileV23(path, filesystem, contents, mode)
+	}
```
bgilbert: For Ignition v1 and cloud-config, I'd rather we panic than silently do nothing.

platform/platform.go
```diff
@@ -143,6 +143,7 @@ type SystemdDropin struct {
 type Options struct {
 	BaseName       string
 	Distribution   string
+	OSContainer    string
```
bgilbert: Could use a comment describing the contents.

platform/cluster.go
arithx: This is intentional, I'm just enabling the `pivot.service` unit

platform/cluster.go
arithx: Will do, this unit was pulled from https://github.com/openshift/machine-config-operator/pull/363#issuecomment-463397373

platform/conf/conf.go
arithx: As a result of this I've dropped the `rhcos.ignition.v1.users` test. RHCOS shouldn't expect to ever receive Ignition v1 content and keeping the test would require a lot more runner plumbing than seems to be worth it to disallow that test specifically if an `oscontainer` is specified.

arithx: cc @miabbott

arithx:
Pushed an update with a few more test disables. This one converts all Ignition v1 tests to be CL specific as neither RHCOS nor FCOS should ever be receiving these configs.

PR #993, opened 2019-04-10
kola/tests/rpmostree: use different pkg for 'rpm-ostree (un)install'

miabbott:
Move away from `fpaste` as the package to test with because it
requires Python2 on the host and that is not available in RHCOS.

I made the name of the package and the path to the binary variables
for easier reuse.  I also dropped the check on running the binary.

The tests pass the latest RHCOS and should pass FCOS, but I was
unable to run `kola` against FCOS.

### DIFF
--- a/kola/tests/rpmostree/deployments.go
+++ b/kola/tests/rpmostree/deployments.go
@@ -195,10 +195,17 @@ func rpmOstreeUpgradeRollback(c cluster.TestCluster) {
 
 // rpmOstreeInstallUninstall verifies that we can install a package
 // and then uninstall it
-// we've hardcoded 'fpaste' throughout the test because it should be
-// a package that a) will be around a long time and b) is small +
-// well-defined
+//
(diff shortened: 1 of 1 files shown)


kola/tests/rpmostree/deployments.go
```diff
@@ -199,6 +199,10 @@ func rpmOstreeUpgradeRollback(c cluster.TestCluster) {
 // a package that a) will be around a long time and b) is small +
 // well-defined
 func rpmOstreeInstallUninstall(c cluster.TestCluster) {
+	// bcrypt is available in EPEL and installs on Fedora 29 and RHEL 8
```
yuqi-zhang: I just noticed that the comment right above refereces `fpaste` still, so maybe delete that and move this up?

miabbott: ack, will do

miabbott:
Good to merge

PR #995, opened 2019-04-29
*: add Ignition v2s3 support

### LINKED ISSUE #974: Ignition 3.0.0 Support
Ignition 3.0.0 is being actively worked on and is something that we should be running through additional testing via `kola`.

As the 3.0.0 spec will not be backwards compatible we will need to have a mechanism for declaring what Ignition version the distribution under test is running that can change out both the underlying config generation as well as some test level user-data to supported configuration versions.

Initially it might make sense to at least get a branch setup that drops Ignition 1.x & 2.x support and updates existing test configurations to 3.0.0 specs. Although that will create a large maintenance burden.

arithx:
Change List:
- Switch to go modules (NOTE: This ends up bumping the vendored AWS version as well as a few stdlibs)
- Add Ignition v2s3 to the vendor
- Add new cmdline flag `--ignition-version` which can be used to override distribution defaults to send specific versions of Ignition. Accepted values are `v2` to send Ignition spec version 2 configs or `v3` to send v2s3 configs.
- Adds v2s3 configs to tests

Closes #974 
Closes coreos/fedora-coreos-tracker#172


arithx:
NOTE: currently untested against RHCOS, tested against CL & FCOS.

For the AWS bump I've tested kola & `ore aws upload`

cmd/kola/options.go
```diff
@@ -58,6 +63,7 @@ func init() {
 	sv(&kola.Options.BaseName, "basename", "kola", "Cluster name prefix")
 	ss("debug-systemd-unit", []string{}, "full-unit-name.service to enable SYSTEMD_LOG_LEVEL=debug on. Specify multiple times for multiple units.")
 	sv(&kola.UpdatePayloadFile, "update-payload", "", "Path to an update payload that should be made available to tests")
+	sv(&kola.Options.IgnitionVersion, "ignition-version", "", "Ignition version override: v2, v3")
```
arithx: Just threw up `v2` & `v3` initially so we could bikeshed here.

Other ideas I tossed around were just using an int var and `s2/s3`

ajeddeloh: I'm fine with either. Strings have the advantage of being able to be more fine grained in the future if we need it (not sure why we would though)

arithx:
@ajeddeloh what do you think about creating a mini-config that has the unit / drop-in  / etc and then merging that with the current config (I think with the mini-config being the child config so it has precedent on overwrites). In addition to adding config validation before launching.

ajeddeloh:
SGTM

arithx:
Pushed with changes to switch to merging v3 configs rather than directly inserting into the objects & validating all Ignition configs at the `RenderUserData` stage (which happens for all machines)

PR #1003, opened 2019-05-20
platform/qemu: disable auto-read-only on qemu disks

arithx:
Disable auto-read-only. A patch was added in qemu[1] `v3.1.0-rc0` & `v4.0.0`
that returns `-EACCESS` if the node is read-write and `BDRV_O_AUTO_RD_ONLY`
is not set or `bdrv_can_set_read_only()` forbids making the node
read-only.

[1] https://github.com/qemu/qemu/commit/eaa2410f1ea864609090c0a5fda9e0ce9499da79

### DIFF
--- a/platform/qemu.go
+++ b/platform/qemu.go
@@ -20,9 +20,12 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


arithx:
Updated to add a check so this patch is conditionaly enabled only on QEMU versions 3.1.0 & greater

PR #1006, opened 2019-05-30
DNM: platform/api/aws: add additional s3 wrappers

arithx:
Adds three new wrappers to the AWS API for S3 operations. These
functions were identified as being likely to be used in future plume
work.


arithx:
(meant to open this as a draft DNM)

PR #1008, opened 2019-06-05
plume: add FCOS

arithx:


cmd/plume/fcos.go
```diff
@@ -0,0 +1,49 @@
...
+		Region:  "us-east-1",
+	}
+)
+
+func AddFcosSpecFlags(flags *pflag.FlagSet) {
+	flags.StringVar(&specPolicy, "policy", "public-read", "Canned ACL policy")
+	flags.StringVar(&specCommitId, "commit-id", "", "OSTree Commit ID")
+}
+
+func FcosChannelSpec() fcosChannelSpec {
```
jlebon: This function is confusing me. So, it returns a statically initialized struct, *but* verifies some runtime vars, some of which don't seem to be defined in this file? Can we move those checks to `runFcosRelease` ?

cmd/plume/release.go
```diff
@@ -445,3 +469,72 @@ func doAWS(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
...
+	api, err := aws.New(&aws.Options{
+		CredentialsFile: awsCredentialsFile,
+		Profile:         spec.Profile,
+		Region:          spec.Region,
+	})
+	if err != nil {
+		plog.Fatalf("creating aws client: %v", err)
+	}
+
+	err = api.UpdateBucketObjectsACL(spec.Bucket, filepath.Join("prod", "streams", specChannel, "builds", specVersion), specPolicy)
```
jlebon: Maybe throw a link to https://github.com/coreos/fedora-coreos-tracker/issues/189 as a comment for now? Once we're happy with it, we can add it to the design doc or something and have a better reference link.

cmd/plume/release.go
```diff
@@ -445,3 +469,72 @@ func doAWS(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
...
+	if err != nil {
+		plog.Fatalf("unmarshaling release metadata json: %v", err)
+	}
+
+	url, err := url.Parse(fmt.Sprintf("https://%v.s3.amazonaws.com/prod/streams/%v/builds/%v/release.json", spec.Bucket, specChannel, specVersion))
+	if err != nil {
+		plog.Fatalf("creating metadata url: %v", err)
+	}
+
+	m.Releases = append(m.Releases, BuildMetadata{
```
jlebon: Hmm, this isn't idempotent though right? Shouldn't we check if the release is already in the list first?

cmd/plume/release.go
```diff
@@ -445,3 +469,72 @@ func doAWS(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
...
+		CommitHash: specCommitId,
+		Version:    specVersion,
+		Endpoint:   url.String(),
+	})
+
+	m.Timestamp = time.Now().Format("2006-01-02T15:04:05Z")
```
jlebon: /me reads up on `Format()`

> Format returns a textual representation of the time value formatted according to layout, which defines the format by showing how the reference time, defined to be
> ```
> Mon Jan 2 15:04:05 -0700 MST 2006
> ```

That's... an interesting approach.

cmd/plume/release.go
```diff
@@ -445,3 +469,72 @@ func doAWS(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
...
+
```
jlebon: Missing e.g.:

```go
m.Observation = "Not for direct consumption; do not depend on this file or its contents."
```

:)

cmd/plume/release.go
arithx: Yeah, golang has an `RFC3339` constant in the `time` package, but @bgilbert had mentioned wanting an ISO-8601 compliant timestamp in this format (which slightly differs from `RFC3339` which does implement ISO-8601 but is variant at the end).

Looking at this again I think I need to modify this line to `time.Now().UTC().Format()` though.

cmd/plume/release.go
arithx: Ah yeah, I forgot that I needed to switch to doing that. Before the code was using hash based keys rather than a list of objects (which can contain duplicates) so the overwrite was taking care of that.

cmd/plume/fcos.go
arithx: This is how it's implemented for other distros (see CL here: https://github.com/coreos/mantle/blob/master/cmd/plume/containerlinux.go#L350-L411). Part of that reasoning is because we want that check ran for both prerelease & release (which would require duplication rather than handling it in the `channelSpec` functions as well as some runtime vars being required for fetching the correct `channelSpec`).

Given FCOS is pretty divergent WDYT about creating another function specifically for checking these runtime vars and calling it separately inside of `runFcosRelease`?

cmd/plume/release.go
arithx: My thinking was that that line would be inside of the initial document and then is just persisted rather than us overwriting the value each time `plume` is ran.

cmd/plume/types.go
```diff
@@ -78,3 +78,21 @@ type channelSpec struct {
...
 }
+
+type fcosChannelSpec struct {
+	Bucket  string
+	Profile string
+	Region  string
+}
+
+type ReleaseMetadata struct {
+	Observation string          `json:"observation"` // used to note to users not to consume the release metadata index
```
arithx: Feel free to bikeshed this name :)

cmd/plume/types.go
```diff
@@ -78,3 +78,21 @@ type channelSpec struct {
...
+	Releases    []BuildMetadata `json:"releases"`
+	Timestamp   string          `json:"timestamp"`
```
arithx: Probably going to change to:
```json
{
    "metadata": {
        "last-modified": "<timestamp>"
    }
}
```

per https://github.com/coreos/fedora-coreos-tracker/issues/98#issuecomment-499291779

cmd/plume/types.go
bgilbert: `info`?  `note`?  `caution`?

arithx: I like `note` but I could be swayed to `caution` if we feel it better conveys that users should pay attention to it :)

arithx:
Updates pushed ⬆️

cmd/plume/fcos.go
jlebon: Yup, I like the new approach!

cmd/plume/release.go
```diff
@@ -445,3 +470,84 @@ func doAWS(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
...
+	newRel := BuildMetadata{
+		CommitHash: specCommitId,
+		Version:    specVersion,
+		Endpoint:   url.String(),
+	}
+
+	for _, rel := range m.Releases {
+		if rel == newRel {
+			// this BuildMetadata already exists, exit
```
jlebon: Should we sanity check here that if it already exists, that it's the last element in the array? Given that we're relying on ordering.

arithx: I don't like the idea of moving entries in the array around (nor do I really like the fact that something is relying on array ordering but that's another topic), so I guess I can check if the last entry is exactly what we're wanting to post and otherwise just add another new entry (potentially creating a duplicate but making this release the latest ordered one while not retroactively modifying things).

arithx:
Updated

cmd/plume/release.go
```diff
@@ -445,3 +470,88 @@ func doAWS(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
...
+
+	path := filepath.Join("prod", "streams", specChannel, "releases.json")
+
+	f, err := api.DownloadFile(spec.Bucket, path)
+	if err != nil {
+		plog.Fatalf("downloading release metadata index: %v", err)
+	}
+	defer f.Close()
+
+	data, err := ioutil.ReadAll(f)
```
ajeddeloh: I'm fine with this as well, but the json API has a steaming mode you can use with file objects (any `io.Reader`) (https://golang.org/pkg/encoding/json/#Decoder)

cmd/plume/release.go
```diff
@@ -445,3 +470,88 @@ func doAWS(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
...
+		plog.Fatalf("reading release metadata index: %v", err)
+	}
+
+	var m ReleaseMetadata
+	err = json.Unmarshal(data, &m)
+	if err != nil {
+		plog.Fatalf("unmarshaling release metadata json: %v", err)
+	}
+
+	url, err := url.Parse(fmt.Sprintf("https://%v.s3.amazonaws.com/prod/streams/%v/builds/%v/release.json", spec.Bucket, specChannel, specVersion))
```
ajeddeloh: lets use %s instead of %v since they're all strings.

PR #1009, opened 2019-06-11
kola/tests/coretest/core: add RHCOS checks for disabled services

miabbott:
In RHCOS, we are explicitly disabling certain services.  These
additional tests will verify that we don't regress.

### DIFF
--- a/kola/tests/coretest/core.go
+++ b/kola/tests/coretest/core.go
@@ -23,6 +23,26 @@ const (
 	UpdateEnginePubKeyV2 = "a76a22e6afcdfbc55dd2953aa950c7ec93b254774fca02d13ec52c59672e5982"
 )
 
+// RHCOS services we expect disabled/inactive
+var offServices = []string{
+	"dnsmasq.service",
+	"iscsid.service",
+	"iscsid.socket",
+	"iscsiuio.service",
+	"nfs-blkmap.service",
+	"nfs-idmapd.service",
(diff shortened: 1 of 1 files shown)


kola/tests/coretest/core.go
```diff
@@ -275,6 +296,49 @@ func servicesActive(units []string) error {
 	return nil
 }
 
+func TestServicesDisabledRHCOS() error {
+	inactiveErr := servicesInactive(offServices)
```
arithx: No need to use separate error variables as you're not returning both, just re-use a single `err` variable.

kola/tests/coretest/core.go
```diff
@@ -275,6 +296,49 @@ func servicesActive(units []string) error {
...
+		}
+	}
+	return nil
+}
+
+func servicesDisabled(units []string) error {
+	for _, unit := range units {
+		c := exec.Command("systemctl", "is-enabled", unit)
+		var out bytes.Buffer
+		c.Stdout = &out
```
arithx: Any reason to do this rather than `out, err := c.Output()`?

miabbott: Just learning my way around golang...the example from the docs had this way, but I like the one-line approach much better.

miabbott:
Updated based on review :arrow_up:

miabbott:
Good to merge

Thanks for the feedback @arithx

PR #1012, opened 2019-06-13
platform: make machine check errors more useful

ajeddeloh:
Do not claim the machine is not reachable when it actually is and is
just booting. If it times out while booting, also log active jobs.

### DIFF
--- a/platform/platform.go
+++ b/platform/platform.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


platform/platform.go
```diff
@@ -326,16 +327,25 @@ func CheckMachine(ctx context.Context, m Machine) error {
 		}
 		out, stderr, err := m.SSH("systemctl is-system-running")
 		if !bytes.Contains([]byte("initializing starting running stopping"), out) {
-			return nil // stop retrying if the system went haywire
+			return fmt.Errorf("machine reached a bad state: %s, %v, %s", out, err, stderr)
```
arithx: We'll need to build a mechanism for detecting if we're running `kola spawn`. As this code currently is we cannot spawn into machines in a degraded state any longer with this patch.

PR #1015, opened 2019-06-25
cmd/plume/fcos: modify build metadata structure

arithx:
Per discussions in the [fedora-coreos-tracker](https://github.com/coreos/fedora-coreos-tracker/issues/98#issuecomment-505104633), modify the build metadata structure to better support multi-arch.

### DIFF
--- a/cmd/plume/release.go
+++ b/cmd/plume/release.go
@@ -518,25 +518,62 @@ func modifyReleaseMetadataIndex(spec *fcosChannelSpec, commitId string) {
 		plog.Fatalf("unmarshaling release metadata json: %v", err)
 	}
 
-	url, err := url.Parse(fmt.Sprintf("https://%s.s3.amazonaws.com/prod/streams/%s/builds/%s/release.json", spec.Bucket, specChannel, specVersion))
+	releasePath := filepath.Join("prod", "streams", specChannel, "builds", specVersion, "release.json")
+	url, err := url.Parse(fmt.Sprintf("https://%s.s3.amazonaws.com/%s", spec.Bucket, releasePath))
 	if err != nil {
 		plog.Fatalf("creating metadata url: %v", err)
 	}
 
+	releaseFile, err := api.DownloadFile(spec.Bucket, releasePath)
+	if err != nil {
+		plog.Fatalf("downloading release metadata: %v", err)
+	}
+	defer releaseFile.Close()
+
+	releaseData, err := ioutil.ReadAll(releaseFile)
+	if err != nil {
+		plog.Fatalf("reading release metadata: %v", err)
+	}
+
+	var im IndividualReleaseMetadata
+	err = json.Unmarshal(releaseData, &im)
+	if err != nil {
+		plog.Fatalf("unmarshaling release metadata: %v", err)
+	}
+
+	var commits []Commit
+	for arch, vals := range im.Architectures {
+		commits = append(commits, Commit{
+			Architecture: arch,
+			Checksum:     vals.Commit,
+		})
+	}
+
 	newRel := BuildMetadata{
-		CommitHash: specCommitId,
+		CommitHash: commits,
 		Version:    specVersion,
 		Endpoint:   url.String(),
 	}
 
 	for i, rel := range m.Releases {
-		if rel == newRel {
+		if compareStaticReleaseInfo(rel, newRel) {
 			if i != (len(m.Releases) - 1) {
 				plog.Fatalf("build is already present and is not the latest release")
 			}
 
-			// the build is already the latest release, exit
-			return
+			comp := compareCommits(rel.CommitHash, newRel.CommitHash)
+			if comp == 0 {
+				// the build is already the latest release, exit
+				return
+			} else if comp == -1 {
+				// the build is present and contains a subset of the new release data,
+				// pop the old entry and add the new version
+				m.Releases = m.Releases[:len(m.Releases)-1]
+				break
+			} else {
+				// the commit hash of the new build is not a superset of the current release
+				plog.Fatalf("build is present but commit hashes are not a superset of latest release")
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


arithx:
cc @sinnykumari @lucab @jlebon

cmd/plume/release.go
```diff
@@ -518,25 +518,62 @@ func modifyReleaseMetadataIndex(spec *fcosChannelSpec, commitId string) {
...
+	err = json.Unmarshal(releaseData, &im)
+	if err != nil {
+		plog.Fatalf("unmarshaling release metadata: %v", err)
+	}
+
+	var commits []Commit
+	for arch, vals := range im.Architectures {
+		commits = append(commits, Commit{
+			Architecture: arch,
+			Checksum:     vals.Commit,
```
lucab: @jlebon just double-checking: is it ok if we call this field (the ostree commit SHA) `commit.checksum`, or is there some possible confusion with `ostree-content-checksum` (which I'm not sure what it is)?

cmd/plume/types.go
```diff
@@ -92,11 +92,24 @@ type ReleaseMetadata struct {
...
+	CommitHash []Commit `json:"commits"`
+	Version    string   `json:"version"`
+	Endpoint   string   `json:"endpoint"`
 }
 
 type Metadata struct {
 	LastModified string `json:"last-modified"`
 }
+
+type IndividualReleaseMetadata struct {
```
lucab: I think at some later point (even in another PR) it would make sense to split structures belonging to different JSON-documents in different source files. It took me a bit to realize where this one was coming from.

cmd/plume/release.go
```diff
@@ -518,25 +518,62 @@ func modifyReleaseMetadataIndex(spec *fcosChannelSpec, commitId string) {
...
+		CommitHash: commits,
 		Version:    specVersion,
 		Endpoint:   url.String(),
 	}
 
 	for i, rel := range m.Releases {
-		if rel == newRel {
+		if compareStaticReleaseInfo(rel, newRel) {
 			if i != (len(m.Releases) - 1) {
 				plog.Fatalf("build is already present and is not the latest release")
```
lucab: Was this supposed to disappear too?

cmd/plume/release.go
```diff
@@ -555,3 +592,34 @@ func modifyReleaseMetadataIndex(spec *fcosChannelSpec, commitId string) {
...
+	}
+	return true
+}
+
+// returns -1 if a is a subset of b, 0 if equal, 1 if a is not a subset of b
+func compareCommits(a, b []Commit) int {
+	if len(a) > len(b) {
+		return 1
+	}
+	equal := len(a) == len(b)
```
lucab: s/equal/sameLenght/ (or something similar, just for clarity)

arithx:
> wondering should we keep "note" or remove it from ReleaseMetadata https://github.com/coreos/mantle/blob/master/cmd/plume/types.go#L89 ?

It was decided upstream to have the `note` field. This is just an implementation of that design.

cmd/plume/release.go
arithx: This is intended imo. I believe this falls into a case that Plume should not be handling which is essentially the case of republishing an older build as the most recent. In scenarios where that needs to occur it should be a deliberate manual action.

cmd/plume/types.go
arithx: I feel like that might cause confusion though if the FCOS metadata types are defined in separate files while every other struct used in plume is inside of this file though. Is this maybe more of an issue with the individual names of these structs rather than the location?

cmd/plume/release.go
```diff
@@ -555,3 +592,34 @@ func modifyReleaseMetadataIndex(spec *fcosChannelSpec, commitId string) {
...
+func compareStaticReleaseInfo(a, b BuildMetadata) bool {
+	if a.Version != b.Version || a.Endpoint != b.Endpoint {
+		return false
+	}
+	return true
+}
+
+// returns -1 if a is a subset of b, 0 if equal, 1 if a is not a subset of b
+func compareCommits(a, b []Commit) int {
+	if len(a) > len(b) {
```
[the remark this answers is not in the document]
arithx: I can see it being superfluous but I don't think it's inherently wrong. If a is larger than b it cannot be a subset so this exists as an early exit hatch rather than checking the individual entries.

sinnykumari:
> > wondering should we keep "note" or remove it from ReleaseMetadata https://github.com/coreos/mantle/blob/master/cmd/plume/types.go#L89 ?
> 
> It was decided upstream to have the `note` field. This is just an implementation of that design.

Ok, should we get it added into the release-index https://github.com/coreos/fedora-coreos-tracker/issues/98#issuecomment-505104633 ?

arithx:
> Ok, should we get it added into the release-index coreos/fedora-coreos-tracker#98 (comment) ?

I don't think it's strictly necessary as this is mostly just present in the case that something other than tools stumbles upon the file to dissuade usage. Any tool (e.x. the stream metadata generator) doesn't need to know or care about this field when parsing the release metadata index.

sinnykumari:
> > Ok, should we get it added into the release-index [coreos/fedora-coreos-tracker#98](https://github.com/coreos/fedora-coreos-tracker/issues/98) (comment) ?
> 
> I don't think it's strictly necessary as this is mostly just present in the case that something other than tools stumbles upon the file to dissuade usage. Any tool (e.x. the stream metadata generator) doesn't need to know or care about this field when parsing the release metadata index.

sure, mentioned it to make sure we didn't miss anything :)

cmd/plume/release.go
jlebon: Yup, checksum is fine. In any context where we're talking about the content checksum, we'd explicitly preface it with "content". For more information of what it is, see https://github.com/ostreedev/ostree/pull/1449.

cmd/plume/release.go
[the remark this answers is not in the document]
lucab: Don't mind me, it's hot today. I misread that as an equality check. The actual code is correct, my brain a bit less :)

cmd/plume/types.go
lucab: I'm fine either ways (and also keeping as-is). It was just a minor sentiment when scanning through the source.

arithx:
Updated addressing comments.

lucab:
LGTM. If you are not in a hurry I'd suggest leaving this hanging till tomorrow so @sinnykumari has a last chance for a pass.

arithx:
@lucab yup, was planning on waiting for everyone involved to have the chance to take a peek.

cmd/plume/release.go
```diff
@@ -518,25 +518,62 @@ func modifyReleaseMetadataIndex(spec *fcosChannelSpec, commitId string) {
 		plog.Fatalf("unmarshaling release metadata json: %v", err)
 	}
 
-	url, err := url.Parse(fmt.Sprintf("https://%s.s3.amazonaws.com/prod/streams/%s/builds/%s/release.json", spec.Bucket, specChannel, specVersion))
+	releasePath := filepath.Join("prod", "streams", specChannel, "builds", specVersion, "release.json")
```
jlebon: Note we'll need to tweak this soon once we [fix the layout](https://github.com/coreos/coreos-assembler/issues/463#issuecomment-506011148).

arithx: I'm not sure I understand, are we planning on moving the `release.json` from `s3://fcos-builds/prod/streams/$stream/builds/$version/`?

jlebon: Sorry, I was just confused. Don't mind me!

Issue #1004, opened 2019-05-22
[RHCOS] running kola in qemu_unpriv mode causes crio.service failure to start


miabbott:
When running `kola` against RHCOS using `qemu_unpriv` is causing all the tests to fail because the `crio.service` is failing to start:

```
$ bin/kola run --debug --qemu-image rhcos-410.8.20190520.0-qemu.qcow2 -b rhcos -p qemu-unpriv --ignition-version v2

<snip>

=== RUN   coreos.selinux.enforce
=== RUN   coreos.ignition.sethostname
=== RUN   rhcos.selinux.boolean.persist
=== RUN   rpmostree.upgrade-rollback
=== RUN   coreos.selinux.boolean
=== RUN   coreos.ignition.once
=== RUN   rhcos.basic
=== RUN   podman.network
=== RUN   rhcos.selinux.manage
=== RUN   ostree.basic
=== RUN   podman.base
=== RUN   rpmostree.status
=== RUN   coreos.auth.verify
=== RUN   coreos.ignition.v2.users
=== RUN   coreos.ignition.groups
2019-05-22T15:46:50Z platform: disabling auto-read-only for QEMU drives
2019-05-22T15:46:50Z platform/machine/qemu: NewMachine: ["qemu-system-x86_64" "-machine" "accel=kvm" "-cpu" "host" "-m" "1024" "-bios" "bios-256k.bin" "-smp" "1" "-uuid" "ba98fd19-8e2a-499c-ad1e-683684c66a4e" "-di
splay" "none" "-chardev" "file,id=log,path=_kola_temp/qemu-unpriv-2019-05-22-1146-19764/coreos.selinux.enforce/ba98fd19-8e2a-499c-ad1e-683684c66a4e/console.txt" "-serial" "chardev:log" "-fw_cfg" "name=opt/com.core
os/config,file=_kola_temp/qemu-unpriv-2019-05-22-1146-19764/coreos.selinux.enforce/ba98fd19-8e2a-499c-ad1e-683684c66a4e/ignition.json" "-add-fd" "fd=3,set=1" "-drive" "if=none,id=d3,format=qcow2,file=/dev/fdset/1,
auto-read-only=off" "-device" "virtio-blk-pci,drive=d3,serial=primary-disk" "-netdev" "user,id=eth0,restrict=yes,hostfwd=tcp:127.0.0.1:0-:22" "-device" "virtio-net-pci,netdev=eth0"]
[...]

PR #1016, opened 2019-07-01
kola: Don't check for crio.service started on RHCOS

cgwalters:
I plan to disable it by default, see this commit message:

    preset: Don't enable crio.service by default

    First, we actually need to reconfigure crio before using
    it in primary OpenShift use cases to use the pause image from
    the release payload; see
    https://github.com/openshift/installer/pull/1761
    And also the followup
    https://github.com/openshift/installer/pull/1768

    Second, having it enabled by default breaks our kola tests
    https://github.com/coreos/mantle/issues/1004
    because the way it handles networking is broken:
    https://github.com/cri-o/cri-o/issues/2478

    We don't need to enable it by default; rather, have it
    be pulled in by `kubelet.service`.

### DIFF
--- a/kola/tests/coretest/core.go
+++ b/kola/tests/coretest/core.go
@@ -70,7 +70,7 @@ func init() {
 		NativeFuncs: map[string]func() error{
 			"PortSSH":          TestPortSsh,
 			"DbusPerms":        TestDbusPerms,
-			"ServicesActive":   TestServicesActiveRHCOS,
+			"ServicesActive":   TestServicesActiveCoreOS,
 			"ServicesDisabled": TestServicesDisabledRHCOS,
 			"ReadOnly":         TestReadOnlyFs,
 			"Useradd":          TestUseradd,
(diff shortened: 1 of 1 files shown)


kola/tests/coretest/core.go
```diff
@@ -70,7 +70,7 @@ func init() {
 		NativeFuncs: map[string]func() error{
 			"PortSSH":          TestPortSsh,
 			"DbusPerms":        TestDbusPerms,
-			"ServicesActive":   TestServicesActiveRHCOS,
+			"ServicesActive":   TestServicesActiveFCOS,
```
arithx: Let's rename the actual test function from `TestServicesActiveFCOS` to `TestServicesActiveCoreOS` since it's targeting both OS's now.

arithx:
Will merge when travis reports green, thanks for the PR

kola/tests/coretest/core.go
cgwalters: done

PR #1018, opened 2019-07-08
platform: disable pinger & zincati by default on FCOS

### LINKED ISSUE #1017: Consider disabling Zincati/pinger in kola tests by default
On Container Linux, update_engine doesn't check into CoreUpdate until 7 minutes after startup, so it's unlikely to do so during a kola run.  On FCOS, if Zincati or pinger want to check in much more quickly than that, we should disable them in kola tests to avoid spurious checkins.

cc @lucab @rfairley

arithx:
Defaults to disabling both fedora-coreos-pinger & Zincati on the FCOS
distribution. Both are using a `90-` config so that it can still be
overridden by individual tests.

Closes #1017

### DIFF
--- a/platform/cluster.go
+++ b/platform/cluster.go
@@ -169,6 +169,14 @@ func (bc *BaseCluster) RenderUserData(userdata *conf.UserData, ignitionVars map[
 		conf.CopyKeys(keys)
 	}
 
+	// disable Zincati & Pinger by default
+	if bc.Distribution() == "fcos" {
+		conf.AddFile("/etc/fedora-coreos-pinger/config.d/90-disable-reporting.toml", "root", `[reporting]
+enabled = false`, 0644)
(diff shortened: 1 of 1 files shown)


arithx:
cc @rfairley @lucab

platform/cluster.go
```diff
@@ -169,6 +169,14 @@ func (bc *BaseCluster) RenderUserData(userdata *conf.UserData, ignitionVars map[
 		conf.CopyKeys(keys)
 	}
 
+	// disable Zincati & Pinger by default
+	if bc.Distribution() == "fcos" {
+		conf.AddFile("/etc/fedora-coreos-pinger/config.d/90-disable-reporting.toml", "root", `[reporting]
+enabled = false`, 0644)
+		conf.AddFile("/etc/zincati/config.d/90-disable-auto-updates.toml", "root", `[updates]
```
lucab: FYI `update.enabled = true` (as a single line) is also valid in TOML as an equivalent format, if you prefer that syntax.

Issue #1025, opened 2019-07-13
plume: FCOS: specify content-type of releases JSON


arithx:
The `releases.json` file built by `plume release --distro fcos` should be uploaded with the `application/json` content-type.

PR #1026, opened 2019-07-15
plume/s3: use proper content-type for releases.json

lucab:
This fixes plume and the underlying library to use "application/json"
as the proper content-type for `releases.json`.

Closes: https://github.com/coreos/mantle/issues/1025


jlebon:
https://github.com/coreos/coreos-assembler/pull/614

PR #1032, opened 2019-07-24
vendor: bump ignition to v2.0.1

ajeddeloh:


go.sum
```diff
@@ -40,6 +40,8 @@ github.com/coreos/ignition v0.31.0 h1:e6ktS2Flcxq6SIKc1JN9NV1gQxmH2gaVwpdgtJIGep
 github.com/coreos/ignition v0.31.0/go.mod h1:WJQapxzEn9DE0ryxsGvm8QnBajm/XsS/PkrDqSpz+bA=
 github.com/coreos/ignition/v2 v2.0.0 h1:2nIU6rQIh2bCSF0PNJKuNNa23L13G031CoxbI4GFojc=
 github.com/coreos/ignition/v2 v2.0.0/go.mod h1:EJP9Gk/u21BPwWa5nT6yYbm5mO0u8JQAAzFqaxgH7Fg=
```
jlebon: Hmm, shouldn't the older one drop out?

ajeddeloh: Doesn't hurt either way, the `go.sum` is just for verification. I can drop it if you want. Normally `go mod tidy` would fix this but running that appears to want to update a bunch of stuff to incompatible versions (to the point of not compiling).

ajeddeloh:
Agreed. Going to merge as is for now and file a bug.

PR #1036, opened 2019-08-01
kola: skip tests incompatible with FCOS

jlebon:
With this, we get a clean FCOS run on AWS! 🎉

### DIFF
--- a/kola/tests/misc/tls.go
+++ b/kola/tests/misc/tls.go
@@ -35,7 +35,7 @@ func init() {
 		ClusterSize:    1,
 		Name:           "coreos.tls.fetch-urls",
 		Flags:          []register.Flag{register.RequiresInternetAccess}, // Networking outside cluster required
-		ExcludeDistros: []string{"rhcos"},                                // wget not included in RHCOS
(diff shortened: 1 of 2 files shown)


kola/tests/misc/tls.go
```diff
@@ -35,7 +35,7 @@ func init() {
 		ClusterSize:    1,
 		Name:           "coreos.tls.fetch-urls",
 		Flags:          []register.Flag{register.RequiresInternetAccess}, // Networking outside cluster required
-		ExcludeDistros: []string{"rhcos"},                                // wget not included in RHCOS
+		ExcludeDistros: []string{"rhcos", "fcos"},                        // wget not included in *COS
```
lucab: @arithx I feel that at this point this would be better split in a curl test and in a wget one, WDYT?

arithx: @lucab yeah, we should probably split that in the future. I'll open an issue for updating / creating tests on FCOS.

jlebon: Similarly for the NFS one, I think we should maybe rename that one to e.g. "nfs-server-enablement" and have another simpler `systemd.enable-service` that targets a different service (or maybe even just a dummy service that we drop in via Ignition).

Issue #1035, opened 2019-07-29
`ore aws upload` does not override existing AMIs/snapshots


miabbott:
Maybe I am misunderstanding what the `--force` flag should do, but it doesn't appear to function the way it is described. 

```
--force                    overwrite existing S3 object rather than reusing
```

Namely, if I use the `--force` flag when trying to upload a `vmdk` to AWS, if the name of the uploaded image is the same, the existing image is reused.  I would expect that with the `--force` flag, the image would always get re-uploaded/overwritten, but this does not appear to happen.

```
### Initial upload

$ coreos-assembler buildextend-aws --build 42.80.20190729.4 --region us-east-1 --bucket s3://rhcos-ci/testing/amis
...
+ ore aws upload --region us-east-1 --bucket s3://rhcos-ci/testing/amis --ami-name rhcos-42.80.20190729.4 --name rhcos-42.80.20190729.4 --ami-description "OpenShift 4 42.80.20190729.4" --file tmp/buildpost-aws/rht
2019-07-29 15:00:07.677076 I | platform/api/aws: uploading s3://rhcos-ci/testing/amis/rhcos-42.80.20190729.4.vmdk
2019-07-29 15:00:23.078125 I | platform/api/aws: created snapshot import task import-snap-0a90d36ee59fb59e0
2019-07-29 15:10:26.944972 I | platform/api/aws: deleting s3://rhcos-ci/testing/amis/rhcos-42.80.20190729.4.vmdk

### Use --force with same image name, AMI name, file name (expecting no reuse)

$ ore upload --force --region us-east-1 --bucket s3://rhcos-ci/testing/amis --ami-name rhcos-42.80.20190729.4 --name rhcos-42.80.20190729.4 --ami-description "OpenShift 4 42.80.20190729.4" --file rhcos-42.80.20190729.4-aws.vmdk --disk-size-inspect --delete-object
2019-07-29 11:50:05.305003 I | platform/api/aws: found existing snapshot snap-05c8783a3c27820b8
2019-07-29 11:50:05.305071 I | platform/api/aws: deleting s3://rhcos-ci/testing/amis/rhcos-42.80.20190729.4-aws.vmdk
2019-07-29 11:50:06.245980 I | platform/api/aws: found existing image ami-0ea3a0a1c565b986c, reusing
[...]

PR #1039, opened 2019-08-07
ore: Make upload --force deregister AMI too

cgwalters:
This matches the semantic intended for coreos-assembler.

Closes: https://github.com/coreos/mantle/issues/1035

### DIFF
--- a/platform/api/aws/images.go
+++ b/platform/api/aws/images.go
@@ -370,6 +370,63 @@ func (a *API) CreatePVImage(snapshotID string, diskSizeGiB uint, name string, de
 	return a.createImage(params)
 }
 
+func (a *API) deregisterImageIfExists(name string) error {
+	imageID, err := a.FindImage(name)
+	if err != nil {
+		return err
+	}
+	if imageID != "" {
+		_, err := a.ec2.DeregisterImage(&ec2.DeregisterImageInput{ImageId: &imageID})
+		if err != nil {
+			return err
+		}
+		plog.Infof("Deregistered existing image %s", imageID)
+	}
+	return nil
+}
+
+// Remove all uploaded data associated with an AMI.
+func (a *API) RemoveImage(name, s3BucketName, s3ObjectPath string) error {
+	err := a.DeleteObject(s3BucketName, s3ObjectPath)
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok {
+			if awsErr.Code() != "NoSuchKey" {
+				return err
+			}
+		} else {
+			return err
+		}
+	} else {
+		plog.Infof("Deleted existing S3 object bucket:%s path:%s", s3BucketName, s3ObjectPath)
+	}
+
+	err = a.deregisterImageIfExists(name + "-hvm")
+	if err != nil {
+		return err
+	}
+	err = a.deregisterImageIfExists(name)
+	if err != nil {
+		return err
+	}
+
+	snapshot, err := a.FindSnapshot(name)
+	if err != nil {
+		return err
+	}
+	if snapshot != nil {
+		// We explicitly ignore errors here in case somehow another AMI was based
+		// on that snapshot
+		_, err := a.ec2.DeleteSnapshot(&ec2.DeleteSnapshotInput{SnapshotId: &snapshot.SnapshotID})
+		if err != nil {
+			plog.Warningf("deleting snapshot %s: %v", snapshot.SnapshotID, err)
+		} else {
+			plog.Infof("Deleted existing snapshot %s", snapshot.SnapshotID)
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


cgwalters:
Only compile tested so far.

bgilbert:
Per https://github.com/coreos/mantle/issues/1035#issuecomment-516547129, this should try to delete the underlying snapshot as well (otherwise we're leaking storage), but continue if that fails (since something else may also hold a reference to the snapshot).  Also I'd tend to agree that deletion should be handled by ore calling primitives, rather than as a special path in `createImage`.

cgwalters:
Bigger picture, do we care about ore's existing behavior?  Any reason not to unconditionally do the equivalent of `--force` and optimize it for cosa?

bgilbert:
I don't think there are compatibility constraints for ore (unlike plume) but unconditionally deleting a random image with the same name seems a bit antisocial.  What about failing by default if the image already exists (by calling `FindImage` from ore), unless the user specifies either e.g. `--reuse-existing` or `--overwrite-existing`?

arithx:
As @bgilbert mentioned I'd prefer that we handle the `FindImage` call followed by a new `DeleteImage` (and `DeleteSnapshot`) call inside of `cmd/ore/aws/upload.go` directly rather than adding it inside of the `createImage` function.

In the case of `DeleteSnapshot` if it provides any error the behavior should be to warn & print the ID (there's no other easy way to fetch once the AMI is deleted as mentioned in https://github.com/coreos/mantle/issues/1035#issuecomment-516547129).

----

> What about failing by default if the image already exists (by calling FindImage from ore), unless the user specifies either e.g. --reuse-existing or --overwrite-existing?

I think this can lead to a can of worms, do these options refer to S3 objects, snapshots, or AMIs? I'd probably prefer to have `--force` & explicit flags like `--reuse-s3`, `--reuse-snapshot` & `--reuse-ami`.

We might want to punt this discussion to another issue / future PR so as to get this in faster.

----

Side note for the current code: the description for the `--force` flag needs to be updated to reflect the new functionality (github won't let me comment on lines you didn't change).

cgwalters:
> but unconditionally deleting a random image with the same name seems a bit antisocial

Hardly random, you'd have to be in the same AWS account and using the same OS name right?

bgilbert:
> Hardly random, you'd have to be in the same AWS account and using the same OS name right?

Well, yeah.  So the issue is shared dev (or prod?) accounts.

Note: I do have a preference here but I don't feel super-strongly about it.

cgwalters:
With https://github.com/coreos/coreos-assembler/pull/693 the snapshot won't get lost.

And I think garbage collection is best implemented in cosa after we land https://github.com/coreos/coreos-assembler/issues/668

bgilbert:
> With coreos/coreos-assembler#693 the snapshot won't get lost.

For uploads initiated by cosa, not those initiated by hand, which after all is the primary purpose of ore.  There's no point in deleting the AMI if we leave behind the part that costs money.

cgwalters:
Thought about this some more and I think the real fix has to be https://github.com/coreos/coreos-assembler/issues/668 - reusing version numbers as we are doing now is just too prone to confusion and failure.

Anyways, I redid the patch per request since I think the existing status quo is really bad - it took me a lot of confusing debugging to realize that the AMI had been reused from a previous failed build.

cgwalters:
(Still only compile tested though)

platform/api/aws/images.go
```diff
@@ -370,6 +370,43 @@ func (a *API) CreatePVImage(snapshotID string, diskSizeGiB uint, name string, de
...
+	}
+
+	snapshot, err := a.FindSnapshot(name)
+	if err != nil {
+		return err
+	}
+	if snapshot != nil {
+		// We explicitly ignore errors here in case somehow another AMI was based
+		// on that snapshot
+		a.ec2.DeleteSnapshot(&ec2.DeleteSnapshotInput{SnapshotId: &snapshot.SnapshotID})
```
arithx: Let's capture the error here and if it's not `nil` perform a `plog.Warnf("deleting snapshot %s: %v", snapshot.SnapshotID, err)`

cgwalters:
OK actually testing this it didn't work because I was missing the `-hvm` for the `FindImage()` call.  Fixed that.  Added some logging too.

Also it's a bit tempting to just change the code and flip that to suffixing with `-pv` for the PV image and not suffixing what everyone uses now...but eh, separate PR.

cgwalters:
See also https://github.com/coreos/coreos-assembler/pull/698

cgwalters:
I don't have merge to this repo, what is the procedure?

ajeddeloh:
Just say "ready for merge".

PR #1040, opened 2019-08-08
platform/api/aws/images: return snapshotID from CopyImage 

arithx:
Snapshots are created when copying images to new regions, these extra
resources cost money and should be communicated by the API. Add
additional code in `plume` to throw out said snapshot IDs as they are
not published to end-users.

The end result of this PR from the perspective of command interactions
is that `ore aws copy-image` will have the snapshot ID included in the
JSON output.

### DIFF
--- a/cmd/plume/prerelease.go
+++ b/cmd/plume/prerelease.go
@@ -610,9 +610,13 @@ func awsUploadToPartition(spec *channelSpec, part *awsPartitionSpec, imageName, 
 		amis := map[string]string{}
 		if len(destRegions) > 0 {
 			plog.Printf("Replicating AMI %v...", imageID)
-			amis, err = api.CopyImage(imageID, destRegions)
+			copiedAmis, err := api.CopyImage(imageID, destRegions)
 			if err != nil {
 				return nil, fmt.Errorf("couldn't copy image: %v", err)
+			}
+
+			for region, data := range copiedAmis {
+				amis[region] = data.AMI
 			}
 		}
 		amis[part.BucketRegion] = imageID
--- a/platform/api/aws/images.go
+++ b/platform/api/aws/images.go
@@ -100,6 +100,11 @@ type Snapshot struct {
 	SnapshotID string
 }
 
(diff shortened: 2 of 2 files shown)


cmd/ore/aws/copy-image.go
```diff
@@ -53,7 +63,49 @@ func runCopyImage(cmd *cobra.Command, args []string) error {
...
+				return fmt.Errorf("creating AWS client in %s region: %v", region, err)
+			}
+			imageName, err := tmpAPI.GetAMIName(ami)
+			if err != nil {
+				return err
+			}
+			// Mantle automatically adds a `-hvm` suffix to HVM images but not to their snapshots
+			// As such we need to trim it from the image name to allow the FindSnapshot function
+			// to find the underlying snapshot of the new AMI
+			imageName = strings.TrimSuffix(imageName, "-hvm")
```
arithx: @bgilbert I couldn't really decide whether or not I wanted to keep this inside of the `ore` code or directly move it into the `FindSnapshot` API code. Do you have a preference?

cmd/ore/aws/copy-image.go
```diff
@@ -53,7 +63,49 @@ func runCopyImage(cmd *cobra.Command, args []string) error {
 		os.Exit(1)
 	}
 
-	err = json.NewEncoder(os.Stdout).Encode(amis)
+	out := make(map[string]CopiedImage)
+	for region, ami := range amis {
+		var snapshotID string
+		err = util.RetryConditional(12, 10*time.Second, func(err error) bool {
```
jlebon: Hmm, 12 seems a bit high. Should we only be retrying on certain error conditions?
If there's context around this, let's throw a link to that if so.

arithx: Ah yeah, this was originally 6 times at 10 seconds in between and I bumped it while trying to debug what ended up being user error. I'll knock it back down.

arithx:
Updated to return the snapshot ID from `CopyImage`

PR #1041, opened 2019-08-09
kola/packages: add no python test for fcos

mike-nguyen:
There was a lot of effort put into removing python from  FCOS.
Lets have a simple test that will let us know if python gets
pulled into a build by accident.


mike-nguyen:
ok to merge

PR #1042, opened 2019-08-13
kola/tests/rpmostree: handle install path for fcos

mike-nguyen:
The rpm-ostree install/uninstall tests  were failing because bcrypt
is installed in /bin on RHEL CoreOS and /usr/bin in Fedora CoreOS.

### DIFF
--- a/kola/tests/rpmostree/deployments.go
+++ b/kola/tests/rpmostree/deployments.go
@@ -239,7 +239,13 @@ func rpmOstreeUpgradeRollback(c cluster.TestCluster) {
 // currently using.  We've already had to swap from `fpaste` to `bcrypt`
 func rpmOstreeInstallUninstall(c cluster.TestCluster) {
 	var installPkgName = "bcrypt"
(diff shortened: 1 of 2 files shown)


kola/tests/rpmostree/deployments.go
```diff
@@ -239,7 +239,7 @@ func rpmOstreeUpgradeRollback(c cluster.TestCluster) {
 // currently using.  We've already had to swap from `fpaste` to `bcrypt`
 func rpmOstreeInstallUninstall(c cluster.TestCluster) {
 	var installPkgName = "bcrypt"
-	var installPkgBin = "/bin/bcrypt"
+	installPkgBin := [2]string{"/bin/bcrypt", "/usr/bin/bcrypt"}
```
ajeddeloh: Lets instead set this based on which distro we're on. It would also keep the logic below simpler.

mike-nguyen:
pushed requested change  :arrow_up:

mike-nguyen:
ok to merge

PR #1043, opened 2019-08-14
kola: Add --blacklist-test switch

ashcrow:
When one or more --blacklist-test provided at run time any test that matches by name or pattern to the provided black list is not added to the list of tests to execute.

Example:
```
$ # Run all linux tests except the nfs ones
$ ./bin/kola --blacklist-test linux.nfs* run linux.*
$ # Run everything but linux.nfs.v3
$ ./bin/kola --blacklist-test linux.nfs.v3 run
```

/cc @miabbott

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -52,6 +52,7 @@ func init() {
 	sv := root.PersistentFlags().StringVar
 	bv := root.PersistentFlags().BoolVar
 	ss := root.PersistentFlags().StringSlice
+	ssv := root.PersistentFlags().StringSliceVar
 
 	// general options
 	sv(&outputDir, "output-dir", "", "Temporary output directory for test data and logs")
@@ -64,7 +65,7 @@ func init() {
 	ss("debug-systemd-unit", []string{}, "full-unit-name.service to enable SYSTEMD_LOG_LEVEL=debug on. Specify multiple times for multiple units.")
 	sv(&kola.UpdatePayloadFile, "update-payload", "", "Path to an update payload that should be made available to tests")
 	sv(&kola.Options.IgnitionVersion, "ignition-version", "", "Ignition version override: v2, v3")
-
+	ssv(&kola.BlacklistedTests, "blacklist-test", []string{}, "List of tests to blacklist")
 	// rhcos-specific options
 	sv(&kola.Options.OSContainer, "oscontainer", "", "oscontainer image pullspec for pivot (RHCOS only)")
 
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -76,6 +76,8 @@ var (
 
 	UpdatePayloadFile string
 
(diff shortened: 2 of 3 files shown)


kola/harness.go
```diff
@@ -194,7 +196,27 @@ func filterTests(tests map[string]*register.Test, pattern, pltfrm string, versio
...
+		// Drop anything which is blacklisted directly or by pattern
+		blacklisted = false
+		for _, bl := range BlacklistedTests {
+			match, err := filepath.Match(bl, t.Name)
+			if err != nil {
+				return nil, err
+			}
+			// If it matched the pattern or the name is exact denote that
+			// this test is blacklisted
+			if match || t.Name == bl {
```
arithx: Shouldn't need to do a direct comparison here, if it's directly equal `match` will already be true (https://play.golang.org/p/6p8UECMI4r4).

cmd/kola/options.go
```diff
@@ -64,7 +65,7 @@ func init() {
 	ss("debug-systemd-unit", []string{}, "full-unit-name.service to enable SYSTEMD_LOG_LEVEL=debug on. Specify multiple times for multiple units.")
 	sv(&kola.UpdatePayloadFile, "update-payload", "", "Path to an update payload that should be made available to tests")
 	sv(&kola.Options.IgnitionVersion, "ignition-version", "", "Ignition version override: v2, v3")
-
+	ssv(&kola.BlacklistedTests, "blacklist-test", []string{}, "Test to blacklist. Use multiple times to blacklist multiple tests")
```
arithx: For the description let's keep it simple with `List of tests to blacklist`.

`cobra` also allows comma-seperated values (https://godoc.org/github.com/spf13/pflag#StringSliceVar) so unless you feel it's necessary to spell out one possible way of using it we might as well keep it open.

arithx:
Would also be nice to update https://github.com/coreos/mantle#kola-run with your example of blacklisting a list of tests.

ashcrow:
Updated per review

ashcrow:
Once this gets a second review it should be good to merge.

kola/harness.go
```diff
@@ -194,7 +196,26 @@ func filterTests(tests map[string]*register.Test, pattern, pltfrm string, versio
...
+		// Drop anything which is blacklisted directly or by pattern
+		blacklisted = false
+		for _, bl := range BlacklistedTests {
+			match, err := filepath.Match(bl, t.Name)
+			if err != nil {
+				return nil, err
+			}
+			// If it matched the pattern this test is blacklisted
+			if match {
+				blacklisted = true
```
yuqi-zhang: Minor, probably should just break here

ashcrow: Good idea. Updating.

ashcrow:
Updated

arithx:
merging on green CI

Issue #1045, opened 2019-08-20
s390x (IBM Z) support to pass ignition file to CoreOS guest


barthy1:
fw_cfg is used to share igntion file with CoreOS guest
https://github.com/coreos/mantle/blob/master/platform/qemu.go#L283
However fw_cfg parameter is not supported for s390x (IBM Z) and ppc64le (Power)

The proposal is to use [guestfish](http://libguestfs.org) for architectures, which do not support fw_cfg, like it is done for coreos-assembler https://github.com/coreos/coreos-assembler/blob/master/src/cmd-run#L239

You can also see example of the implementation and more details here https://github.com/dmacvicar/terraform-provider-libvirt/pull/629

lucab:
Thanks for the report!

> fw_cfg parameter is not supported for s390x (IBM Z) and ppc64le (Power)

I don't think there is any technical limitation for not having support those architectures (but I may be wrong, as I'm not an expert in that area).
Do we have ticket to track adding support for those two? What's the current status? Perhaps we can increase the priority of those a bit, if we indeed plan to have consumers?

bgilbert:
@lucab The QEMU developers don't want fw_cfg used in this way and aren't excited about adding a generic userdata mechanism.  They've suggested alternatives, several of which won't work because they require waiting indefinitely for udev.  https://github.com/coreos/ignition/issues/666 and https://github.com/coreos/ignition/issues/825 have some of the context.  One suggestion was that we pass userdata via a virtual PCI device, which might work but we haven't pursued it yet.

Our official guidance for passing userdata on s390x shouldn't be to use guestfish, so we'll still need a proper metadata mechanism.  It might be okay for kola to use guestfs in the short term, to unblock testing, but we shouldn't expect that to be the long-term approach.

lucab:
Sorry, I was not aware of the additional discussion on `fw_cfg`. The backchannel discussion for `s390x.qemu` is already tracked at https://github.com/coreos/ignition/issues/825 then.

barthy1:
@lucab So would it be ok for you to use guestfish in the short term to unblock testing activities?

barthy1:
https://github.com/coreos/mantle/pull/1050

PR #1048, opened 2019-08-23
Add s390x support for kola in mantle

barthy1:
This is the starting PR to add possibility to build mantle utilities and run kola on s390x architecture.
The code adds s390x architecture to the list of other supported archs to avoid "not-supported" errors and add functionality to generate qemu command, working for s390x

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -167,7 +167,7 @@ func syncOptions() error {
 	}
 
 	image, ok := kolaDefaultImages[kola.QEMUOptions.Board]
-	if !ok {
+	if kola.QEMUOptions.Distribution == "cl" && !ok {
 		return fmt.Errorf("unsupport board %q", kola.QEMUOptions.Board)
 	}
 
--- a/platform/api/packet/api.go
+++ b/platform/api/packet/api.go
@@ -65,10 +65,12 @@ var (
 	defaultPlan = map[string]string{
 		"amd64-usr": "baremetal_0",
 		"arm64-usr": "baremetal_2a",
+		"s390x-usr": "baremetal_3a",
 	}
 	linuxConsole = map[string]string{
 		"amd64-usr": "ttyS1,115200",
 		"arm64-usr": "ttyAMA0,115200",
+		"s390x-usr": "ttysclp0,115200",
 	}
 )
 
--- a/platform/qemu.go
+++ b/platform/qemu.go
@@ -265,18 +265,29 @@ func CreateQEMUCommand(board, uuid, biosImage, consolePath, confPath, diskImageP
 			"-cpu", "host",
 			"-m", "2048",
 		}
+	case "s390x--s390x-usr":
+		qmBinary = "qemu-system-s390x"
+		qmCmd = []string{
+			"qemu-system-s390x",
+			"-machine", "s390-ccw-virtio,accel=kvm",
+			"-cpu", "host",
+			"-m", "2048",
+		}
 	default:
 		panic("host-guest combo not supported: " + combo)
 	}
 
 	qmCmd = append(qmCmd,
-		"-bios", biosImage,
 		"-smp", "1",
 		"-uuid", uuid,
 		"-display", "none",
 		"-chardev", "file,id=log,path="+consolePath,
 		"-serial", "chardev:log",
 	)
+
+	if board != "s390x-usr" {
+		qmCmd = append(qmCmd, "-bios", biosImage)
+	}
 
 	if isIgnition {
 		qmCmd = append(qmCmd,
@@ -346,6 +357,8 @@ func Virtio(board, device, args string) string {
 		suffix = "pci"
 	case "arm64-usr":
 		suffix = "device"
+	case "s390x-usr":
+		suffix = "ccw"
 	default:
 		panic(board)
 	}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@ properties([
 
     parameters([
         choice(name: 'GOARCH',
-               choices: "amd64\narm64",
+               choices: "amd64\narm64\ns390x",
                description: 'target architecture for building binaries')
     ]),
 
--- a/build
+++ b/build
@@ -26,7 +26,7 @@ host_build() {
 
 cross_build() {
 	local a
-	for a in amd64 arm64; do
+	for a in amd64 arm64 s390x; do
 		echo "Building $a/$1"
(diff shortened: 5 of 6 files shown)


barthy1:
@alicefr

cmd/kola/options.go
```diff
@@ -35,6 +35,7 @@ var (
 	kolaDefaultImages  = map[string]string{
 		"amd64-usr": sdk.BuildRoot() + "/images/amd64-usr/latest/coreos_production_image.bin",
 		"arm64-usr": sdk.BuildRoot() + "/images/arm64-usr/latest/coreos_production_image.bin",
+                "s390x-usr": sdk.BuildRoot() + "/images/s390x-usr/latest/coreos_production_image.bin",
```
arithx: You can drop this, CL doesn't produce s390x builds.

platform/api/packet/api.go
```diff
@@ -57,18 +57,22 @@ var (
 		// HTTPS causes iPXE to fail on a "permission denied" error
 		"amd64-usr": "http://stable.release.core-os.net/amd64-usr/current",
 		"arm64-usr": "http://beta.release.core-os.net/arm64-usr/current",
+		"s390x-usr": "http://beta.release.core-os.net/s390x-usr/current",
```
arithx: You can drop this, CL doesn't produce s390x builds.

platform/api/packet/api.go
```diff
@@ -57,18 +57,22 @@ var (
...
 	}
 	defaultImageURL = map[string]string{
 		"amd64-usr": "https://alpha.release.core-os.net/amd64-usr/current/coreos_production_packet_image.bin.bz2",
 		"arm64-usr": "https://alpha.release.core-os.net/arm64-usr/current/coreos_production_packet_image.bin.bz2",
+		"s390x-usr": "https://alpha.release.core-os.net/s390x-usr/current/coreos_production_packet_image.bin.bz2",
```
arithx: You can drop this, CL doesn't produce s390x builds.

platform/qemu.go
```diff
@@ -265,19 +265,30 @@ func CreateQEMUCommand(board, uuid, biosImage, consolePath, confPath, diskImageP
...
 		"-serial", "chardev:log",
 	)
 
+	if runtime.GOARCH != "s390x" {
+		qmCmd = append(qmCmd, "-bios", biosImage)
+	}
+
 	if isIgnition {
 		qmCmd = append(qmCmd,
 			"-fw_cfg", "name=opt/com.coreos/config,file="+confPath)
```
arithx: Does `s390x` support `fw_cfg`? If not it's probably necessary to modify this `if` block to `isIgnition && board != "s390x"`.

platform/qemu.go
```diff
@@ -265,19 +265,30 @@ func CreateQEMUCommand(board, uuid, biosImage, consolePath, confPath, diskImageP
...
 	qmCmd = append(qmCmd,
-		"-bios", biosImage,
 		"-smp", "1",
 		"-uuid", uuid,
 		"-display", "none",
 		"-chardev", "file,id=log,path="+consolePath,
 		"-serial", "chardev:log",
 	)
 
+	if runtime.GOARCH != "s390x" {
```
arithx: `runtime.GOARCH` is the architecture of the host, this should be checking the `board` variable.

cmd/kola/options.go
```diff
@@ -35,6 +35,7 @@ var (
 	kolaDefaultImages  = map[string]string{
 		"amd64-usr": sdk.BuildRoot() + "/images/amd64-usr/latest/coreos_production_image.bin",
 		"arm64-usr": sdk.BuildRoot() + "/images/arm64-usr/latest/coreos_production_image.bin",
+		"s390x-usr": sdk.BuildRoot() + "/images/s390x-usr/latest/coreos_production_image.bin",
```
arithx: You can drop this, CL doesn't produce s390x builds.

platform/qemu.go
barthy1: @arithx thank you to have a look. `s390x` does not support `fw_cfg`. My team was going to have separate PR with workaround for this problem, see https://github.com/coreos/mantle/issues/1045, because we still need to pass ignition file to the guest. Would you prefer for this PR just to exclude ignition config for s390x at all at this moment (as you suggested), waiting for another PR to fix `fw_cfg` first, or have all the fixes for x390x in this PR?

cmd/kola/options.go
ajeddeloh: for that matter we can drop arm too.

platform/qemu.go
arithx: @barthy1 that's fine, was just pointing it out. Feel free to do it in a separate PR.

cmd/kola/options.go
barthy1: There is a check https://github.com/coreos/mantle/blob/master/cmd/kola/options.go#L170 which fails the kola run, if there is no image for board - `Error: unsupport board "s390x-usr"`
I will add code to exclude s390x from the check. If you think it's necessary, I can add also arm to this list

arithx: That code should probably be modified to check if the distribution is CL, it doesn't make sense in the context of RHCOS/FCOS. I'd say for now, leave the ARM entry and modify the `if` block like so:

```golang
if kola.Options.Distribution == "cl" && !ok {
    ...
}
```

barthy1:
all the discussed updates were done

cmd/kola/options.go
```diff
@@ -167,7 +167,7 @@ func syncOptions() error {
 	}
 
 	image, ok := kolaDefaultImages[kola.QEMUOptions.Board]
-	if !ok {
+	if kola.QEMUOptions.Distribution == "cl" && !ok {
```
ashcrow: So this is unsupported only if it's Container Linux *and* the `kolaDefaultImages` doesn't have an entry?

arithx: The [`kolaDefaultImages` variable](https://github.com/coreos/mantle/blob/master/cmd/kola/options.go#L35-L38) is a hashmap of where the images would live inside of the CL SDK, it's not really relevant outside of the context of CL. We could implement a secondary check for unsupported boards on other platforms but I suggested that the check just be modified in a comment [here](https://github.com/coreos/mantle/pull/1048#discussion_r317183787).

arithx:
@barthy1 @alicefr is this ok to merge?

alicefr:
@arithx for me it's ok

PR #1050, opened 2019-08-28
Add guestfs utility usage for s390x, to replace unsupported fw_cfg

barthy1:
In case of x86 architecture fw_cfg is used to pass ignition config file to the KVM guest.
s390x does not support fw_cfg, so guestfs is option to use.
guestfs allows to modify qcow2 image before running it as guest.
Current implementation modifies not the base qcow2 image, but the temporary one, created based on the testing qcow2 image.
Code has implementation of several calls to guestfish utility, it is expected that libguestfs-tools is installed on host.
The according issue is https://github.com/coreos/mantle/issues/1045

The list of commands to execute and their output for s390x case. Filesystems are not hardcoded in the code, but found based on their labels
  - guestfish --listen -a $IMG
         -- > GUESTFISH_PID=$PID; export GUESTFISH_PID
  - guestfish --remote -- run
  - guestfish --remote -- findfs-label boot
         -- >  /dev/sda1
  - guestfish --remote -- findfs-label root
         -- >  /dev/sda2
  - guestfish --remote -- mount /dev/sda2 /
  - guestfish --remote -- mount /dev/sda1 /boot
  - guestfish --remote -- mkdir -p /boot/ignition
  - guestfish --remote -- upload $HOST_IGNITION_FILE $GUEST_IGNITION_FILE
  - guestfish --remote -- umount-all
  - guestfish --remote -- exit

### DIFF
--- a/platform/qemu.go
+++ b/platform/qemu.go
@@ -15,6 +15,7 @@
 package platform
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -39,6 +40,7 @@ type Disk struct {
 	Size        string   // disk image size in bytes, optional suffixes "K", "M", "G", "T" allowed. Incompatible with BackingFile
 	BackingFile string   // raw disk image to use. Incompatible with Size.
 	DeviceOpts  []string // extra options to pass to qemu. "serial=XXXX" makes disks show up as /dev/disk/by-id/virtio-<serial>
+	ConfPath    string   // path to ignition to be able to use it with guestfs for temporary qcow2 images
 }
 
 var (
@@ -167,14 +169,14 @@ func (d Disk) setupFile() (*os.File, error) {
 	}
 
 	if d.Size != "" {
-		return setupDisk(d.Size)
+		return setupDisk(d.Size, d.ConfPath)
 	} else {
-		return setupDiskFromFile(d.BackingFile)
+		return setupDiskFromFile(d.BackingFile, d.ConfPath)
 	}
 }
 
 // Create a nameless temporary qcow2 image file backed by a raw image.
-func setupDiskFromFile(imageFile string) (*os.File, error) {
+func setupDiskFromFile(imageFile string, confPath string) (*os.File, error) {
 	// a relative path would be interpreted relative to /tmp
 	backingFile, err := filepath.Abs(imageFile)
 	if err != nil {
@@ -191,10 +193,10 @@ func setupDiskFromFile(imageFile string) (*os.File, error) {
 	}
 
 	qcowOpts := fmt.Sprintf("backing_file=%s,lazy_refcounts=on", backingFile)
-	return setupDisk("-o", qcowOpts)
+	return setupDisk(confPath, "-o", qcowOpts)
 }
 
-func setupDisk(additionalOptions ...string) (*os.File, error) {
+func setupDisk(confPath string, additionalOptions ...string) (*os.File, error) {
 	dstFileName, err := mkpath("")
 	if err != nil {
 		return nil, err
@@ -210,7 +212,12 @@ func setupDisk(additionalOptions ...string) (*os.File, error) {
 	if err := qemuImg.Run(); err != nil {
 		return nil, err
 	}
-
+	if len(confPath) > 0 {
+		err = setupIgnition(confPath, dstFileName)
+		if err != nil {
+			return nil, fmt.Errorf("ignition injection with guestfs failed: %v", err)
+		}
+	}
 	return os.OpenFile(dstFileName, os.O_RDWR, 0)
 }
 
@@ -290,8 +297,11 @@ func CreateQEMUCommand(board, uuid, biosImage, consolePath, confPath, diskImageP
 	}
 
 	if isIgnition {
-		qmCmd = append(qmCmd,
-			"-fw_cfg", "name=opt/com.coreos/config,file="+confPath)
+		// -fw_cfg is not supported for s390x, instead guestfs utility is used
+		if board != "s390x-usr" {
+			qmCmd = append(qmCmd,
+				"-fw_cfg", "name=opt/com.coreos/config,file="+confPath)
+		}
 	} else {
 		qmCmd = append(qmCmd,
 			"-fsdev", "local,id=cfg,security_model=none,readonly,path="+confPath,
@@ -318,12 +328,17 @@ func CreateQEMUCommand(board, uuid, biosImage, consolePath, confPath, diskImageP
 		plog.Debugf("disabling auto-read-only for QEMU drives")
 	}
 
-	allDisks := append([]Disk{
-		{
-			BackingFile: diskImagePath,
-			DeviceOpts:  primaryDiskOptions,
-		},
-	}, options.AdditionalDisks...)
+	primaryDisk := Disk{
+		BackingFile: diskImagePath,
+		DeviceOpts:  primaryDiskOptions,
+		ConfPath:    "",
+	}
+
+	if board == "s390x-usr" {
+		primaryDisk.ConfPath = confPath
+	}
+
+	allDisks := append([]Disk{primaryDisk}, options.AdditionalDisks...)
 
 	var extraFiles []*os.File
 	fdnum := 3 // first additional file starts at position 3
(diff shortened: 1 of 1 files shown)


barthy1:
@alicefr

platform/qemu.go
```diff
@@ -351,3 +366,93 @@ func Virtio(board, device, args string) string {
...
+	if err := cmd.Wait(); err != nil {
+		return fmt.Errorf("waiting for guestfish response: %v", err)
+	}
+	//GUESTFISH_PID=$PID; export GUESTFISH_PID
+	gfVarPid := strings.Split(string(buf), ";")
+	gfVarPidArr := strings.Split(gfVarPid[0], "=")
+	os.Setenv(gfVarPidArr[0], gfVarPidArr[1])
+
+	if err := exec.Command("guestfish", "--remote", "--", "run").Run(); err != nil {
+		return fmt.Errorf("fuestfish launch failed: %v", err)
```
alicefr: @barth1 sorry just noticed the typo

alicefr:
@barthy1 on vacation, so I'll take over on this PR (I've also write access).
> Did you by chance look into using the [guestfs-golang library](http://libguestfs.org/guestfs-golang.3.html) instead of shelling out?

@arithx mantle is cross compiled with CGO_ENABLED=0 https://github.com/coreos/mantle/blob/bc773777f4a001edd863a793c21cc3e9ebbc34a9/build#L32, the guestfish libary requires this flag.

alicefr:
@barthy1 I think we need to set the guestfish backend. The default value is `libvirt` and we should change it to `direct`. If we leave it to `libvirt` we can get `Permission denied` if libvirt cannot have access to directory of the qcow

platform/qemu.go
```diff
@@ -351,3 +366,93 @@ func Virtio(board, device, args string) string {
...
 	return fmt.Sprintf("virtio-%s-%s,%s", device, suffix, args)
 }
+
+func setupIgnition(confPath string, diskImagePath string) error {
+	fileRemoteLocation := "/boot/ignition/config.ign"
+
+	plog.Debugf("START guestfs")
+
+	cmd := exec.Command("guestfish", "--listen", "-a", diskImagePath)
+	stdout, err := cmd.StdoutPipe()
```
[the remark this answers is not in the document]
alicefr: @arithx I tried to substitute `cmd.Output()` to `cmd.StdoutPipe()` and it didn't wokr. I think the right choice is `cmd.StdoutPipe()` because guestfish is started in background and the `cmd.Output` returns only when the comand finish.

alicefr:
@arithx I update the PR with an additional commit

platform/qemu.go
```diff
@@ -364,3 +378,89 @@ func Virtio(board, device, args string) string {
 	}
 	return fmt.Sprintf("virtio-%s-%s,%s", device, suffix, args)
 }
+
+func setupIgnition(confPath string, diskImagePath string) error {
+	fileRemoteLocation := "/boot/ignition/config.ign"
+
+	plog.Debugf("START guestfs")
```
arithx: The two start/success `plog.Debugf` lines in this function don't feel particularly useful.

alicefr:
@arithx I tested with our internal rhcos. We still do not have the backport upstream because of anaconda (see https://github.com/coreos/coreos-assembler/pull/707#issuecomment-528236199). We have an internal version of the backport and I tried to run the tests with kola for this version.

platform/qemu.go
alicefr: @ashcrow removed the 2 `plog.Debugf`

platform/qemu.go
```diff
@@ -210,7 +211,12 @@ func setupDisk(additionalOptions ...string) (*os.File, error) {
 	if err := qemuImg.Run(); err != nil {
 		return nil, err
 	}
-
+	if len(confPath) > 0 {
+		err = setupIgnition(confPath, dstFileName)
```
lucab: Nit: this is usually written as a single statement, like `if err := setupIgnition(...); err != nil { ... }`.

platform/qemu.go
```diff
@@ -364,3 +378,86 @@ func Virtio(board, device, args string) string {
...
+	gfVarPidArr := strings.Split(gfVarPid[0], "=")
+	os.Setenv(gfVarPidArr[0], gfVarPidArr[1])
+
+	if err := exec.Command("guestfish", "--remote", "--", "run").Run(); err != nil {
+		return fmt.Errorf("guestfish launch failed: %v", err)
+	}
+
+	bootfs, err := findLabel("boot")
+	if err != nil {
+		return fmt.Errorf("guestfish command failed: %v", err)
```
lucab: This should mention "boot" somewhere in the error string (either here, or in the function).

platform/qemu.go
```diff
@@ -364,3 +378,86 @@ func Virtio(board, device, args string) string {
...
+	}
+
+	rootfs, err := findLabel("root")
+	if err != nil {
+		return fmt.Errorf("guestfish command failed: %v", err)
```
lucab: This should mention "root" somewhere in the error string (either here, or in the function).

platform/qemu.go
```diff
@@ -364,3 +378,86 @@ func Virtio(board, device, args string) string {
...
+	if err != nil {
+		return fmt.Errorf("reading guestfish output: %v", err)
+	}
+	if err := cmd.Wait(); err != nil {
+		return fmt.Errorf("waiting for guestfish response: %v", err)
+	}
+	//GUESTFISH_PID=$PID; export GUESTFISH_PID
+	gfVarPid := strings.Split(string(buf), ";")
+	gfVarPidArr := strings.Split(gfVarPid[0], "=")
+	os.Setenv(gfVarPidArr[0], gfVarPidArr[1])
```
lucab: This is missing a length check on both `gfVarPid` and `gfVarPidArr`. It will otherwise panic if parsing the file does not go as expected.

platform/qemu.go
```diff
@@ -364,3 +378,86 @@ func Virtio(board, device, args string) string {
...
+
+	if err := exec.Command("guestfish", "--remote", "--", "umount-all").Run(); err != nil {
+		return fmt.Errorf("guestfish umount failed: %v", err)
+	}
+
+	return nil
+}
+
+func findLabel(label string) (string, error) {
+	cmd := exec.Command("guestfish", "--remote", "findfs-label", label)
```
lucab: It would be helpful to have a check for an empty `label` before this.

platform/qemu.go
```diff
@@ -318,12 +327,17 @@ func CreateQEMUCommand(board, uuid, biosImage, consolePath, confPath, diskImageP
...
-			DeviceOpts:  primaryDiskOptions,
-		},
-	}, options.AdditionalDisks...)
+	primaryDisk := Disk{
+		BackingFile: diskImagePath,
+		DeviceOpts:  primaryDiskOptions,
+		ConfPath:    "",
+	}
+
+	if board == "s390x-usr" {
```
lucab: @arithx I know the `*-usr` convention is what we had in ContainerLinux. Are we sticking to that also for the new *COS images?

alicefr:
@lucab I've update the PR accordingly. Also added a check if `GUESTFISH_PID` has been set, in `findLabel`

cgwalters:
I've been saying this a lot but...we really need to de-dup the qemu code between cosa and mantle.

Perhaps one thing we could do is move some of the "glue stuff" into mantle itself and have cosa exec it? 

The language thing is a real issue though...I dunno.  Possibly the low-tech way to address this would be to use golang in mantle, and have CLI wrappers for things that e.g. take CLI arguments and output JSON?

platform/qemu.go
ashcrow: Good question. I see [above in the code it's currently still in use](https://github.com/coreos/mantle/pull/1050/files#diff-da5387f5087a8b96b97f913fdefab572R241).

platform/qemu.go
```diff
@@ -364,3 +377,98 @@ func Virtio(board, device, args string) string {
 	}
 	return fmt.Sprintf("virtio-%s-%s,%s", device, suffix, args)
 }
+
+func setupIgnition(confPath string, diskImagePath string) error {
+	fileRemoteLocation := "/boot/ignition/config.ign"
```
ashcrow: nit: (does not need to be updated, maybe a future change) Seems like this should be a constant that is referenced.

platform/qemu.go
```diff
@@ -364,3 +377,98 @@ func Virtio(board, device, args string) string {
...
+	}
+
+	if err := exec.Command("guestfish", "--remote", "--", "umount-all").Run(); err != nil {
+		return fmt.Errorf("guestfish umount failed: %v", err)
+	}
+
+	return nil
+}
+
+func findLabel(label string) (string, error) {
```
ashcrow: nit: Missing function doc

platform/qemu.go
```diff
@@ -364,3 +377,98 @@ func Virtio(board, device, args string) string {
 	}
 	return fmt.Sprintf("virtio-%s-%s,%s", device, suffix, args)
 }
+
+func setupIgnition(confPath string, diskImagePath string) error {
```
ashcrow: nit: Missing function doc

platform/qemu.go
arithx: @lucab There isn't anything necessarily restricting us to that convention, however as it's just a user provided parameter to inform `kola` that it's targeting that board architecture I don't think that there's necessarily any harm in it.

bgilbert: I'd rather not perpetuate the `board` and `-usr` naming if we don't have to.

alicefr: Should we only use `s390x` here?

alicefr:
I updated the PR. Now, we're using `--remote=$pid` to pass the guestfish pid to each command. The env variable causes some conflicts if there're multiple instances of guestfish running at the same time.

platform/qemu.go
```diff
@@ -364,3 +377,100 @@ func Virtio(board, device, args string) string {
...
+	if err := exec.Command("guestfish", remote, "umount-all").Run(); err != nil {
+		return fmt.Errorf("guestfish umount failed: %v", err)
+	}
+	return nil
+}
+
+// findLabel finds the partition based on the label. The partition belongs to the image attached to the guestfish instance identified by pid.
+func findLabel(label, pid string) (string, error) {
+	if pid == "" {
+	 return "", fmt.Errorf("The pid cannot be empty")
```
arithx: This needs to be ran through `gofmt`

barthy1: @arithx fixed

alicefr:
> I've been saying this a lot but...we really need to de-dup the qemu code between cosa and mantle.
> 
> Perhaps one thing we could do is move some of the "glue stuff" into mantle itself and have cosa exec it?
> 
> The language thing is a real issue though...I dunno. Possibly the low-tech way to address this would be to use golang in mantle, and have CLI wrappers for things that e.g. take CLI arguments and output JSON?

@cgwalters do you think, it's something we can fix in this PR, or is it out of scope? I mean, would you like to modify the code accordingly and then add the guestfish part, or before merging this and have a second discussion about the code structure.

cgwalters:
> @cgwalters do you think, it's something we can fix in this PR, or is it out of scope? 

I'm certainly not saying you or anyone else has to do it as part of this PR.  Or to say it positively - I think this PR is fine (and thanks for working on this!).

platform/qemu.go
arithx: I talked with @bgilbert yesterday about this. I'm going to investigate some other approaches for *COS distributions outside of this issue and we won't block on it.

PR #1051, opened 2019-08-30
kola/tests: add test for gce agent

ajeddeloh:
Add a test to ensure the gce agent is running.

### DIFF
--- a/kola/tests/misc/gce.go
+++ b/kola/tests/misc/gce.go
@@ -1,0 +1,57 @@
+// Copyright 2019 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package misc
+
+import (
+	"encoding/json"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+)
+
+func init() {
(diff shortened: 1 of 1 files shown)


kola/tests/misc/gce.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2018 CoreOS, Inc.
```
arithx: 2019 Red Hat

kola/tests/misc/gce.go
```diff
@@ -0,0 +1,53 @@
...
+
+func gceVerifyAgentIsRunning(c cluster.TestCluster) {
+	list := c.MustSSH(c.Machines()[0], "rkt list --format json")
+	rktlist := []rktEntry{}
+	json.Unmarshal(list, &rktlist)
+	if len(rktlist) != 1 {
+		// either it didn't start or it failed and now there's two
+		c.Fatalf("gce agent is not running")
+	}
+	if rktlist[0].State != "running" {
```
arithx: Should we loop `rktlist` and only do this validation if it's the gce agent to prevent a potential future breakage if something else is running in `rkt`?

kola/tests/misc/gce.go
```diff
@@ -0,0 +1,53 @@
...
+
+func init() {
+	register.Register(&register.Test{
+		Name:        "coreos.misc.gce.agent",
+		Platforms:   []string{"gce"},
+		Run:         gceVerifyAgentIsRunning,
+		ClusterSize: 1,
+		// Previously broken on NVMe devices, see
+		// https://github.com/coreos/bugs/issues/2399
+		MinVersion: semver.Version{Major: 1828},
```
bgilbert: There's no need to carry a `MinVersion` with a major older than the current stable.

kola/tests/misc/gce.go
ajeddeloh: nothing else should be running in rkt.

bgilbert: We could at least check that `rktlist[0]` is actually the agent.

ajeddeloh: done

bgilbert:
https://github.com/coreos/bugs/issues/2608

PR #1052, opened 2019-09-04
Add feature template

ashcrow:
Adds a feature template per meeting.

/cc @darkmuggle @ajeddeloh @lucab @vrutkovs @arithx

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,22 @@
+---
+name: Bug report
+about: Report an issue
+---
+
+# Bug Report #
+
+## Environment ##
+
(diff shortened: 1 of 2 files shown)


ashcrow:
Travis failure is unrelated

arithx:
Testing a fix in #1053 to unblock this

bgilbert:
This will cause the "New issue" button to present a UI similar to [this](https://github.com/coreos/ignition/issues/new/choose), but with only one row.  So if we're adding a feature template, we should add a bug template too.

arithx:
#1053 is merged, if you rebase on top of current master the travis failure should go away (as it will now correctly use the cached `vendor/` directory instead of attempting to fetch vendored libraries)

ashcrow:
Rebased and added

PR #1058, opened 2019-09-16
platform/api/aws: add additional checks in ensureInstanceProfile

### LINKED ISSUE #1047: kola can get into weird state in aws account
If I initially run kola without the iam passrole permissions then I end up with a instance profile that doesn't have a role associated with it.

In this case I run kola on a fresh account (no `CreatedBy=mantle` resources) and it fails because I didn't have passrole permissions: 

```
[coreos-assembler]$ kola -p aws --aws-ami ami-0e884738127697eb9 --aws-region us-east-1 -b fcos run coreos.ignition.resource.s3
=== RUN   coreos.ignition.resource.s3
--- FAIL: coreos.ignition.resource.s3 (0.32s)
        harness.go:507: Cluster failed starting machines: error verifying IAM instance profile: adding role "kola" to instance profile "kola": AccessDenied: User: arn:aws:iam::013116697141:user/dusty-fcos is not authorized to perform: iam:PassRole on resource: role kola
        status code: 403, request id: 08080270-c449-11e9-a7a4-a589214ff8db
FAIL, output in _kola_temp/aws-2019-08-21-1922-321
harness: test suite failed
```

Subsequent runs of kola won't re-attempt to fix the error (i.e. a `kola` role exists so continue on):

```
[coreos-assembler]$ 
[coreos-assembler]$ kola -p aws --aws-ami ami-0e884738127697eb9 --aws-region us-east-1 -b fcos run coreos.ignition.resource.s3
=== RUN   coreos.ignition.resource.s3
--- FAIL: coreos.ignition.resource.s3 (351.12s)
        harness.go:507: Cluster failed starting machines: machine "i-0d3611d49fc18f878" failed to start: ssh journalctl failed: dial tcp 52.201.248.149:22: connect: connection refused
) on machine i-0d3611d49fc18f878 consolening (fs/kernfs/dir.c:1503 kernfs_remove_by_name_ns+0x83/0x90
FAIL, output in _kola_temp/aws-2019-08-21-1925-336
harness: test suite failed
```

This test eventually fails because the there is no role in the instance profile:

```
$ curl http://169.254.169.254/latest/meta-data/iam/info
{
  "Code" : "Success",
  "Message" : "Instance Profile does not contain a role.  Please see documentation at http://docs.amazonwebservices.com/IAM/latest/UserGuide/RolesTroubleshooting.html.",
  "LastUpdated" : "2019-08-21T18:43:09Z",
  "InstanceProfileArn" : "arn:aws:iam::00000000000:instance-profile/kola",
  "InstanceProfileId" : "AIPARGFOZ5J262XIR3ZOJ"
}
```

I'm guessing we should either do a check for passrole early and fail before we even try to create the `kola` role, or we should check the instance profile later to make sure it contains a role before continuing. We could do both :)

arithx:
Adds some additional checks in the `ensureInstanceProfile` function that
not only does the instance profile exist but it has the correct role
(with the correct policies) attached to it.

Closes #1047

### DIFF
--- a/platform/api/aws/iam.go
+++ b/platform/api/aws/iam.go
@@ -59,14 +59,50 @@ const (
 // AmazonS3RReadOnlyAccess permissions policy applied to allow fetches
 // of S3 objects that are not owned by the root account.
 func (a *API) ensureInstanceProfile(name string) error {
-	_, err := a.iam.GetInstanceProfile(&iam.GetInstanceProfileInput{
+	profileExists := false
+	policy := "AmazonS3ReadOnlyAccess"
+
+	profile, err := a.iam.GetInstanceProfile(&iam.GetInstanceProfileInput{
 		InstanceProfileName: &name,
 	})
-	if err == nil {
-		return nil
-	}
-	if awserr, ok := err.(awserr.Error); !ok || awserr.Code() != "NoSuchEntity" {
+	if awsErr, ok := err.(awserr.Error); !ok || awsErr.Code() != "NoSuchEntity" {
 		return fmt.Errorf("getting instance profile %q: %v", name, err)
+	} else if err == nil {
+		profileExists = true
+		// check if a role of the same name already exists and is attached
+		// to the instance profile
+		for _, role := range profile.InstanceProfile.Roles {
+			if role.RoleName != nil && *role.RoleName == name {
+				if role.AssumeRolePolicyDocument != nil && *role.AssumeRolePolicyDocument != ec2AssumeRolePolicy {
+					// the role exists but is missing the assume role policy
+					// present this as an error to the user requiring manual intervention
+					return fmt.Errorf("role %q exists but is not configured properly, manual intervention required", name)
+				}
+				// validate that the role has the correct policy
+				_, err := a.iam.GetRolePolicy(&iam.GetRolePolicyInput{
+					PolicyName: aws.String(policy),
+					RoleName:   aws.String(name),
+				})
+				if awserr, ok := err.(awserr.Error); !ok || awserr.Code() != "NoSuchEntity" {
+					return fmt.Errorf("getting role policy: %v", err)
+				} else if awserr.Code() == "NoSuchEntity" {
+					// the role does not contain the AmazonS3ReadOnlyAccess policy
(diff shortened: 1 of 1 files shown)


dustymabe:
LGTM but I'm no golang programmer :)

platform/api/aws/iam.go
```diff
@@ -59,14 +59,50 @@ const (
...
+	profile, err := a.iam.GetInstanceProfile(&iam.GetInstanceProfileInput{
 		InstanceProfileName: &name,
 	})
-	if err == nil {
-		return nil
-	}
-	if awserr, ok := err.(awserr.Error); !ok || awserr.Code() != "NoSuchEntity" {
+	if awsErr, ok := err.(awserr.Error); !ok || awsErr.Code() != "NoSuchEntity" {
 		return fmt.Errorf("getting instance profile %q: %v", name, err)
+	} else if err == nil {
```
jlebon: Can't we just fold back this whole block with the rest of the logic below and just be more idempotent? E.g.:

```
- CreateInstanceProfile, ignore EntityAlreadyExists
- CreateRole, ignore EntityAlreadyExists
- PutRolePolicy (seems like that API is already idempotent)
- AddRoleToInstanceProfile, ignore EntityAlreadyExists
```

arithx: I do like the approach, the main deviation from what you have listed is `AddRoleToInstanceProfile`, in testing this is actually not returning an `EntityAlreadyExists` but rather a `LimitExceeded` because the default limit for `InstanceSessionsPerInstanceProfile` is `1` (essentially each instance profile can only have one role).

I'm not sure I feel that error is sufficient towards ensuring that the correct `Role` is attached to the `InstanceProfile`. Perhaps I'll just add a check there in addition to ignoring the `EntityAlreadyExists` errors on `CreateInstanceProfile` & `CreateRole`

arithx:
Closing as this is extremely old

dustymabe:
no longer useful?

arithx:
@dustymabe not necessarily, the bug is still there but it's an extreme corner case that users aren't likely to hit. I still have the branch on my fork if we decide to go back in the future to fix it.

PR #1060, opened 2019-09-20
gcp: Add UEFI_COMPATIBLE and SECURE_BOOT

cgwalters:
For FCOS we support both.  Mainly I'm adding these because
they seem to be prerequisites for accessing the "vTPM" which
is part of the "Shielded Cloud" umbrella.  We want access to the vTPM
to enable support for LUKS bound to vTPM for example.  Also,
for OpenShift, I think at some point we want to enable using TPM
as a "strong identity" for nodes.

### DIFF
--- a/platform/api/gcloud/image.go
+++ b/platform/api/gcloud/image.go
@@ -42,7 +42,7 @@ type ImageSpec struct {
 // CreateImage creates an image on GCE and returns operation details and
 // a Pending. If overwrite is true, an existing image will be overwritten
 // if it exists.
-func (a *API) CreateImage(spec *ImageSpec, overwrite bool) (*compute.Operation, *Pending, error) {
+func (a *API) CreateImage(spec *ImageSpec, overwrite, fedora bool) (*compute.Operation, *Pending, error) {
 	licenses := make([]string, len(spec.Licenses))
 	for i, l := range spec.Licenses {
 		license, err := a.compute.Licenses.Get(a.options.Project, l).Do()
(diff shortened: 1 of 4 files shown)


cgwalters:
This gets me a vTPM, though for some reason it still doesn't seem to be booting in UEFI mode with Secure Boot.

platform/api/gcloud/image.go
```diff
@@ -76,9 +76,16 @@ func (a *API) CreateImage(spec *ImageSpec, overwrite bool) (*compute.Operation,
...
 		Licenses:    licenses,
 		GuestOsFeatures: []*compute.GuestOsFeature{
+			// https://cloud.google.com/compute/docs/images/create-delete-deprecate-private-images
 			&compute.GuestOsFeature{
 				Type: "VIRTIO_SCSI_MULTIQUEUE",
 			},
+			&compute.GuestOsFeature{
+				Type: "UEFI_COMPATIBLE",
+			},
+			&compute.GuestOsFeature{
```
ajeddeloh: @bgilbert do you know if CL+GCE supports this?

bgilbert: UEFI should work.  Secure Boot does not.

bgilbert:
Actually, let's condition both flags on the OS.  We shouldn't change the CL image configuration at this point.

cgwalters:
How do we condition on the OS?  Does that imply adding a new CLI argument here and passing it from cosa? 

(When can we create a `cl-stable` branch for mantle and not have to think about possibly breaking Cl in mantle PRs?)

bgilbert:
> How do we condition on the OS?  Does that imply adding a new CLI argument here and passing it from cosa?

plume, which we _should_ be using, already knows what OS it's releasing.  For ore we'd have to add a CLI argument.

cgwalters:
OK updated to pass an arg.

> plume, which we should be using, already knows what OS it's releasing.

Yeah...it just hurts my head to even try to think about adding a *fourth* case for plume.  It's really the opposite direction for RHCOS anyways, where the "release" is not directly to users but through the installer or a release image.

arithx:
> (When can we create a cl-stable branch for mantle and not have to think about possibly breaking CL in mantle PRs?)

We talked about this OOB and are going to start preparing to branch mantle for CL.

arithx:
We had some OOB discussions around whether this should be gated post branching for CL. There were some concerns raised about whether or not Fedora Cloud Base would have similar restrictions on those options. Going to merge this as is and we can revisit it at a later date.

PR #1062, opened 2019-09-26
plume: rename --system argument to --distro

dustymabe:
I'm not sure why release and pre-release have two separate option
names for the same thing, which means that the invocation between
release and pre-release differs in a non-intuitive way. This commit
makes them match.

### DIFF
--- a/cmd/plume/prerelease.go
+++ b/cmd/plume/prerelease.go
@@ -96,7 +96,7 @@ func init() {
 	sort.Sort(sort.StringSlice(platformList))
 
 	cmdPreRelease.Flags().StringSliceVar(&selectedPlatforms, "platform", platformList, "platform to pre-release")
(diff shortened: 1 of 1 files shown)


dustymabe:
note that we'll probably have to update any consumers that exist.

jlebon:
> note that we'll probably have to update any consumers that exist.

Is this used by CL though? Seems safer to just add a `--distro` flag and alias the old one (and hide it from `--help` if possible).

dustymabe:
> > note that we'll probably have to update any consumers that exist.
> 
> Is this used by CL though? Seems safer to just add a `--distro` flag and alias the old one (and hide it from `--help` if possible).

I was just preferring we update the one consumer (fcos doesn't use pre-release) that exists instead of doing that.

dustymabe:
> I was just preferring we update the one consumer (fcos doesn't use pre-release) that exists instead of doing that.

actually even CL probably doesn't use this option because "cl" is the default so we might not have to update any consumers :)

PR #1063, opened 2019-09-26
plume: lots of tiny fixups for Fedora distro 

dustymabe:
- plume: add release example to the README
- plume: fedora: update s3 bucket names
- plume: release: only try to publish PV AMIs for CL
- plume: release: add code to do fedora release
- plume: fedora: remove templating from spec.AWS.BaseDescription
- plume: fedora: remove short options
- plume: fedora: target prod/dev account based on `--environment`
- plume: remove Prefix from Fedora channelspecs

### DIFF
--- a/cmd/plume/fedora.go
+++ b/cmd/plume/fedora.go
@@ -30,7 +30,7 @@ var (
 		"x86_64",
 		"aarch64",
 	}
-	awsFedoraPartitions = []awsPartitionSpec{
+	awsFedoraProdAccountPartitions = []awsPartitionSpec{
 		awsPartitionSpec{
 			Name:         "AWS",
 			Profile:      "default",
@@ -58,7 +58,7 @@ var (
 			},
 		},
 	}
-	awsFedoraUserPartitions = []awsPartitionSpec{
+	awsFedoraDevAccountPartitions = []awsPartitionSpec{
 		awsPartitionSpec{
 			Name:         "AWS",
 			Profile:      "default",
@@ -82,7 +82,7 @@ var (
 				BaseName:        "Fedora",
 				BaseDescription: "Fedora {{.ImageType}} AMI",
 				Image:           "Fedora-{{.ImageType}}-{{.Version}}-{{.Timestamp}}.n.{{.Respin}}.{{.Arch}}.raw.xz",
-				Partitions:      awsFedoraPartitions,
+				Partitions:      awsFedoraProdAccountPartitions,
 			},
 		},
 		"branched": channelSpec{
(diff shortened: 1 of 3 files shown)


dustymabe:
cc @jdoss

cmd/plume/release.go
```diff
@@ -461,7 +461,9 @@ func doAWS(ctx context.Context, client *http.Client, src *storage.Bucket, spec *
 				}
 			}
 			if aws.RegionSupportsPV(region) {
-				publish(imageName)
+				if selectedDistro == "cl" {
```
arithx: I'd say put this as the first case in the outer `if` statement to prevent even checking that the region supports PV unless it's on CL.

dustymabe:
tested this out.. should be ready for review/merge

dustymabe:
rebased on top of latest master

cmd/plume/fedora.go
```diff
@@ -119,11 +119,11 @@ var (
 )
 
 func AddFedoraSpecFlags(flags *pflag.FlagSet) {
-	flags.StringVarP(&specEnv, "environment", "E", "prod", "instance environment")
-	flags.StringVarP(&specImageType, "image-type", "I", "Cloud-Base", "type of image")
-	flags.StringVarP(&specTimestamp, "timestamp", "T", "", "compose timestamp")
-	flags.StringVarP(&specRespin, "respin", "R", "0", "compose respin")
-	flags.StringVarP(&specComposeID, "compose-id", "O", "", "compose id")
+	flags.StringVar(&specEnv, "environment", "prod", "AMI upload environment")
```
arithx: Given there are only two modes (the default prod and developer) would it make sense to switch this to a switch and when present swap over to developer partitions?

cmd/plume/fedora.go
```diff
@@ -80,7 +80,7 @@ var (
 			Boards:  awsFedoraBoards,
 			AWS: awsSpec{
 				BaseName:        "Fedora",
-				BaseDescription: "Fedora {{.ImageType}} AMI",
+				BaseDescription: "Fedora Cloud Base AMI",
```
arithx: Does it make sense to drop the `ImageType` parameter and always hard code in the `Image` field as well?

jdoss:
LGTM

cmd/plume/fedora.go
dustymabe: I think it may, but I wasn't sure so I didn't touch it yet

cmd/plume/fedora.go
dustymabe: you mean make it a boolean? "true/false" ? I don't mind it like it is but I could see wanting it to be just a boolean.

arithx: Yeah; ultimately it's more up to you & @jdoss I don't mind leaving it as is.

cmd/plume/fedora.go
arithx: That's fine, we can always come back in a later PR and drop it if it's unnecessary.

dustymabe: cool. Thanks

cmd/plume/fedora.go
dustymabe: I'll leave it as is then :)

PR #1064, opened 2019-09-27
kola/tests/ignition/passwd: update passwd entry for core user on RHCOS

miabbott:
Now that RHCOS is inheriting some of its config from `fedora-coreos-config`,
it uses the same `base.ign` file that includes a `gecos` entry set to
`CoreOS Admin` for the `core` user.

### DIFF
--- a/kola/tests/ignition/passwd.go
+++ b/kola/tests/ignition/passwd.go
@@ -282,7 +282,7 @@ func usersRhcos(c cluster.TestCluster) {
 	tests := []userTest{
 		{
 			user:           "core",
-			passwdRecord:   "core:x:1000:1000::/home/core:/bin/bash",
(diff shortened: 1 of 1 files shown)


miabbott:
FCOS is nearly similar, but the `passwd` entry has `/var/home/core` instead of `/home/core`.

Didn't get a chance to look into that...

jlebon:
> FCOS is nearly similar, but the passwd entry has /var/home/core instead of /home/core. If the two were the same, we could enable this test on FCOS, too.
>
> Didn't get a chance to look into that...

It's because the default `core` user in RHCOS encodes `homeDir: "/home/core"`. I submitted a patch to just drop that so `useradd` uses the correct default (which will match FCOS' `/var/home/core`).

PR #1065, opened 2019-10-02
kola: Change user test for `core` to be /var/home on RHCOS

cgwalters:
We fixed RHCOS to use the "canonical" homedir.

### DIFF
--- a/kola/tests/ignition/passwd.go
+++ b/kola/tests/ignition/passwd.go
@@ -173,7 +173,7 @@ func init() {
 	})
 	register.Register(&register.Test{
 		Name:        "coreos.ignition.v2.users",
-		Run:         usersRhcos,
+		Run:         usersFcos,
 		ClusterSize: 1,
 		UserData: conf.Ignition(`{
 		             "ignition": { "version": "2.0.0" },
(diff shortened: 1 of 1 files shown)


kola/tests/ignition/passwd.go
```diff
@@ -282,7 +282,7 @@ func usersRhcos(c cluster.TestCluster) {
 	tests := []userTest{
 		{
 			user:           "core",
-			passwdRecord:   "core:x:1000:1000:CoreOS Admin:/home/core:/bin/bash",
+			passwdRecord:   "core:x:1000:1000:CoreOS Admin:/var/home/core:/bin/bash",
```
jlebon: How about also:

```diff
diff --git a/kola/tests/ignition/passwd.go b/kola/tests/ignition/passwd.go
index c5cae9ce..e138e29e 100644
--- a/kola/tests/ignition/passwd.go
+++ b/kola/tests/ignition/passwd.go
@@ -219,7 +219,7 @@ func init() {
                               ]
                             }
                           }`),
-               Distros: []string{"rhcos"},
+               Distros: []string{"rhcos", "fcos"},
        })
 }

```

?

Since we should be able to run this as is on FCOS too now.

cgwalters: Right...also just noting here as soon as we can drop CL this is another instance of tech debt that would just go away.

ashcrow:
LGTM as well. Will defer to @arithx per above comment.

PR #1066, opened 2019-10-02
*: add aliyun

arithx:


auth/aliyun.go
```diff
@@ -0,0 +1,66 @@
...
+// ReadAliyunConfig decodes an aliyun config file
+//
+// If path is empty, $HOME/.aliyun/config.json is read.
+func ReadAliyunConfig(path string) (map[string]AliyunProfile, error) {
+	if path == "" {
+		user, err := user.Current()
+		if err != nil {
+			return nil, err
+		}
+		path = filepath.Join(user.HomeDir, OpenStackConfigPath)
```
jlebon: `AliyunConfigPath` ?

arithx: Whoops, yup. Nice catch!

arithx:
Tested & marked ready for review

cmd/ore/aliyun/create.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2019 Red Hat
```
ashcrow: Looking at other `cmd/ore/$CLOUD/`'s it seems like this file is generally called `create-image.go`, though there isn't total consensus:

```
$ find . -name "create.go"
./openstack/create.go
$ find . -name "create-image.go"
./azure/create-image.go
./gcloud/create-image.go
./do/create-image.go
```

cmd/ore/aliyun/create.go
```diff
@@ -0,0 +1,101 @@
...
+		fmt.Fprintf(os.Stderr, "Uploading image to object storage: %v\n", err)
+		os.Exit(1)
+	}
+
+	id, err := API.ImportImage(format, bucket, name, diskSize, device, name, description, architecture)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Couldn't create image: %v\n", err)
+		os.Exit(1)
+	}
+	fmt.Println(id)
```
ashcrow: Is there an output format expected here or is it free form? Examples I see:

## gcloud
```
        fmt.Printf("Creating image in GCE: %v...\n", imageNameGCE)
```

## do
Doesn't look like it outpus an id

network/ssh.go
```diff
@@ -128,8 +128,9 @@ func ensurePortSuffix(host string, port int) string {
 
 func (a *SSHAgent) newClient(host string, user string, auth []ssh.AuthMethod) (*ssh.Client, error) {
 	sshcfg := ssh.ClientConfig{
-		User: user,
-		Auth: auth,
+		User:            user,
+		Auth:            auth,
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
```
ashcrow: Are we sure this is what we want here in all cases? https://godoc.org/golang.org/x/crypto/ssh#InsecureIgnoreHostKey

cmd/ore/aliyun/create.go
arithx: There is not currently an expected output between different `ore` platforms. IMO there are two options here, we either leave it as is or I can update it to be a JSON output (similar to what is output by `copy-image`) in the format of `{'<region>': '<id>'}`.

PR #1067, opened 2019-10-03
kola: initial ppc64le support

jcajka:
This allows kola to be build and run on ppc64le, although the tests will not yet pass.

### DIFF
--- a/platform/api/packet/api.go
+++ b/platform/api/packet/api.go
@@ -68,9 +68,10 @@ var (
 		"s390x-usr": "baremetal_3a",
 	}
 	linuxConsole = map[string]string{
-		"amd64-usr": "ttyS1,115200",
-		"arm64-usr": "ttyAMA0,115200",
-		"s390x-usr": "ttysclp0,115200",
+		"amd64-usr":   "ttyS1,115200",
+		"arm64-usr":   "ttyAMA0,115200",
+		"s390x-usr":   "ttysclp0,115200",
+		"ppc64le-usr": "hvc0,115200",
 	}
 )
 
--- a/build
+++ b/build
@@ -26,7 +26,7 @@ host_build() {
 
 cross_build() {
 	local a
-	for a in amd64 arm64 s390x; do
+	for a in amd64 arm64 s390x ppc64le; do
 		echo "Building $a/$1"
 		mkdir -p "bin/$a"
 		CGO_ENABLED=0 GOARCH=$a \
--- a/platform/qemu.go
+++ b/platform/qemu.go
@@ -278,6 +278,13 @@ func CreateQEMUCommand(board, uuid, biosImage, consolePath, confPath, diskImageP
 			"-cpu", "host",
 			"-m", "2048",
 		}
+	case "ppc64le--ppc64le-usr":
+		qmBinary = "qemu-system-ppc64"
+		qmCmd = []string{
+			"qemu-system-ppc64",
+			"-machine", "pseries,accel=kvm,kvm-type=HV",
+			"-m", "2048",
+		}
 	default:
 		panic("host-guest combo not supported: " + combo)
 	}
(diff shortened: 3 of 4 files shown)


jcajka:
@ajeddeloh would you be ok with merging this, although it doesn't yet fully work(```cosa kola run``` tests fail)? I think that this would unblock/allow to drop some workarounds for ppc64le not building the mantle/kola in cosa.

Prashanth684:
> @ajeddeloh would you be ok with merging this, although it doesn't yet fully work(`cosa kola run` tests fail)? I think that this would unblock/allow to drop some workarounds for ppc64le not building the mantle/kola in cosa.

Also, I recently bumped mantle on cosa branch 4.2 to get s390x working which broke ppc builds..so would be nice to have this in so i can bump mantle to fix ppc.

cmd/kola/options.go
```diff
@@ -33,8 +33,9 @@ var (
 	kolaPlatforms      = []string{"aws", "azure", "do", "esx", "gce", "openstack", "packet", "qemu", "qemu-unpriv"}
 	kolaDistros        = []string{"cl", "fcos", "rhcos"}
 	kolaDefaultImages  = map[string]string{
-		"amd64-usr": sdk.BuildRoot() + "/images/amd64-usr/latest/coreos_production_image.bin",
-		"arm64-usr": sdk.BuildRoot() + "/images/arm64-usr/latest/coreos_production_image.bin",
+		"amd64-usr":   sdk.BuildRoot() + "/images/amd64-usr/latest/coreos_production_image.bin",
+		"arm64-usr":   sdk.BuildRoot() + "/images/arm64-usr/latest/coreos_production_image.bin",
+		"ppc64le-usr": sdk.BuildRoot() + "/images/ppc64le-usr/latest/coreos_production_image.bin",
```
arithx: This can be dropped, these default images are a legacy remnant of Container Linux.

platform/api/packet/api.go
```diff
@@ -63,14 +63,16 @@ var (
 		"arm64-usr": "https://alpha.release.core-os.net/arm64-usr/current/coreos_production_packet_image.bin.bz2",
 	}
 	defaultPlan = map[string]string{
-		"amd64-usr": "baremetal_0",
-		"arm64-usr": "baremetal_2a",
-		"s390x-usr": "baremetal_3a",
+		"amd64-usr":   "baremetal_0",
+		"arm64-usr":   "baremetal_2a",
+		"s390x-usr":   "baremetal_3a",
+		"ppc64le-usr": "baremetal_4a",
```
arithx: Is this an actual packet slug?

ajeddeloh:
I'm fine with merging this without complete support (i.e. with kola run failing). Does it fail completely (i.e. the runner fails), or do the tests themselves fail?

jcajka:
@ajeddeloh sorry, VMs are spun up and from the logs it seems that the ignition is injected and executed correctly, but there seems to be most probably some issue with network/ssh config as ssh connections gets rejected. Still investigating and I'm not sure where the culprit lies, kola, cosa, ignition, config...

platform/api/packet/api.go
jcajka: could you elaborate?

cmd/kola/options.go
jcajka: ack

platform/api/packet/api.go
arithx: Is `baremetal_4a` actually a Packet machine type for a ppc64le machine?

jcajka: Honestly I have no idea.

cmd/kola/options.go
jcajka: Looking at the extend of the change needed to dorp it and my rather limited knowledge of the code base. I will drop the ppc64le addition there as it shouldn't be need as per your comment.

platform/api/packet/api.go
arithx: This can be dropped as well then. To my knowledge packet doesn't have a ppc64le machine type.

arithx:
@jcajka the config that `kola` is attempting to inject will be viewable at `_kola_temp/<platform>-<last_run_date>/<test_name>/<machine_id>/ignition.json` and the serial console log will be in the same folder named `console.txt`. Feel free to reach out if you need some assistance debugging.

platform/api/packet/api.go
jcajka: Ack, thanks.

jcajka:
@arithx I have seen the logs, it took me bit of time(still getting distracted by cosa ;) ) to figure out that it is actually lack of entropy in the machine that caused the failed/refused ssh connection. I will and virtio-rng device to the qemu invocation(as in cosa build/run). Are you ok with it added across all arches?

jcajka:
Just pushed with virtio-rng added across all arches. For the record ```cosa kola run -- -d -p qemu-unpriv``` passes on ppc64le FCOS.

Prashanth684:
> Just pushed with virtio-rng added across all arches. For the record `cosa kola run -- -d -p qemu-unpriv` passes on ppc64le FCOS.

passes on ppc64le RHCOS too..

PR #1070, opened 2019-10-03
kola/tests/crio/crio: Add pods and service restart test

ashcrow:
Adds a test that verifies pods continue when the crio service is restarted.

See: https://access.redhat.com/solutions/3825511

### DIFF
--- a/kola/tests/crio/crio.go
+++ b/kola/tests/crio/crio.go
@@ -81,7 +81,8 @@ var crioPodTemplate = `{
 }`
 
 // crioContainerTemplate is a simple string template required for running a container
-// It takes three strings: the name (which will be expanded), the image, and the argument to run
+// It takes three strings: the name (which will be expanded), the image, and the argument to run.
+// For mounts see: https://godoc.org/k8s.io/cri-api/pkg/apis/runtime/v1alpha2#Mount
 var crioContainerTemplate = `{
 	"metadata": {
 		"name": "rhcos-crio-container-%s",
@@ -142,13 +143,27 @@ var crioContainerTemplate = `{
 				]
 			}
 		}
-	}
+	},
+	"mounts": [{
+		"container_path": "/tmp/test",
+		"host_path": "/tmp/test",
+		"read_only": false,
+		"selinux_relabel": true,
+		"propagation": 2
+	}]
 }`
 
 // RHCOS has the crio service disabled by default, so use Ignition to enable it
 var enableCrioIgn = conf.Ignition(`{
   "ignition": {
     "version": "2.2.0"
+  },
+  "storage": {
+	"directories": [{
+		"filesystem": "root",
+		"path": "/tmp/test",
+		"mode": 511
+	}]
   },
   "systemd": {
     "units": [
@@ -163,6 +178,12 @@ var enableCrioIgn = conf.Ignition(`{
 var enableCrioIgnV3 = conf.Ignition(`{
   "ignition": {
     "version": "3.0.0"
+  },
+  "storage": {
+	"directories": [{
+		"path": "/tmp/test",
+		"mode": 511
+	}]
   },
   "systemd": {
     "units": [
@@ -200,6 +221,7 @@ func init() {
 // crioBaseTests executes multiple tests under the "base" name
 func crioBaseTests(c cluster.TestCluster) {
 	c.Run("crio-info", testCrioInfo)
+	c.Run("pod-continues-during-service-restart", crioPodContinuesDuringServiceRestart)
 	c.Run("networks-reliably", crioNetworksReliably)
 }
 
@@ -247,7 +269,6 @@ func genContainer(c cluster.TestCluster, m platform.Machine, podName, imageName 
 	if err = c.DropFile(configPathContainer); err != nil {
 		return "", "", err
 	}
-
 	// Create the crio image used for testing, only if it doesn't exist already
 	output := c.MustSSH(m, "sudo podman images -n --format '{{.Repository}}'")
 	if !strings.Contains(string(output), "localhost/"+imageName) {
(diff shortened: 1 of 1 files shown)


kola/tests/crio/crio.go
```diff
@@ -416,3 +417,28 @@ func testCrioInfo(c cluster.TestCluster) {
...
+		c, m, "restart-test", "sleep", []string{"sleep"}, []string{"sleep"})
+	if err != nil {
+		c.Fatal(err)
+	}
+	cmdCreatePod := fmt.Sprintf("sudo crictl runp %s", crioConfigPod)
+	podID := c.MustSSH(m, cmdCreatePod)
+	containerID := c.MustSSH(m, fmt.Sprintf("sudo crictl create %s %s %s",
+		podID, crioConfigContainer, crioConfigPod))
+	// Set the restart to occur in 15 seconds
+	c.MustSSH(m, "sleep 15 && sudo systemctl restart crio &")
```
arithx: I don't think this is working properly, I added some code to test it like so:

```golang
before := time.Now()
c.MustSSH(m, "sleep 15 && sudo systemctl restart crio &")
fmt.Println(time.Since(before))
```

which is printing that it waited for at least 15s before exiting `c.MustSSH`.

One alternative idea would be to build both `sleep` & `echo` into the container and have the container mount in a folder and write to it after the wait. Then the code inside of the test could be something like:

```golang
c.MustSSH(m, "sudo crictl exec %s <sleep & write to file>")
time.Sleep(15 * time.Second())
c.MustSSH(m, "sudo systemctl restart crio")
time.Sleep(45 * time.Second())
output := c.MustSSH(m, "cat <file>")
```

cgwalters:
Is there a specific bug that motivated adding this?  Would be good to link it if so.

ashcrow:
> Is there a specific bug that motivated adding this? Would be good to link it if so.

There is. Good call, I'll add it.

kola/tests/crio/crio.go
ashcrow: Thanks @arithx, I'll give it a go!

cgwalters: Yeah, that suggestion does sound simpler.

Another approach would be running `MustSSH` in a goroutine.  Yet another would be to use `systemd-run` to create a background task on the target host.

cgwalters:
Is this still WIP?  Offhand it seems OK to me, though long term I'd like us to run crio's upstream tests, and ensure that this test lives there.

And just a sanity check since we don't have any useful CI on this repo yet - did you run this locally against rhcos-4.3 on...qemu?

ashcrow:
> Is this still WIP? Offhand it seems OK to me, though long term I'd like us to run crio's upstream tests, and ensure that this test lives there.

It's WIP. I've been running into issues getting it to run due to no network in qemu.

> And just a sanity check since we don't have any useful CI on this repo yet - did you run this locally against rhcos-4.3 on...qemu?

I have not yet.

cgwalters:
Did you try rebasing on https://github.com/coreos/mantle/pull/1072 ?

ashcrow:
> Did you try rebasing on #1072 ?

Ah! No I haven't. But I will :slightly_smiling_face:

ashcrow:
> > Did you try rebasing on #1072 ?
> 
> Ah! No I haven't. But I will slightly_smiling_face

Rebasing was super helpful. The test fails, but now I can better debug why. Thank you @cgwalters .

ashcrow:
```
$ sudo ./bin/kola --platform qemu-unpriv --distro rhcos --ignition-version v2 --ssh-on-test-failure --qemu-image ./rhcos-42.80.20191008.0-qemu.qcow2 run crio.base
=== RUN   crio.base
=== RUN   crio.base/crio-info
=== RUN   crio.base/pod-continues-during-service-restart
=== RUN   crio.base/networks-reliably
--- PASS: crio.base (96.43s)
    --- PASS: crio.base/crio-info (0.12s)
    --- PASS: crio.base/pod-continues-during-service-restart (40.91s)
            cluster.go:122: STEP 1: FROM scratch
            cluster.go:122: STEP 2: COPY . /
            cluster.go:122: STEP 3: COMMIT localhost/sleep
            cluster.go:122: Getting image source signatures
            cluster.go:122: Copying blob sha256:c3ccafd1a83fab850dc9b4599017c171a8f8847518426adaa409ab6e5df89922
            cluster.go:122: Copying config sha256:7f37a57e6fa8f0d4fec7a418cf188843122e30bc53d6b8c02c4aca0cfd8a758f
            cluster.go:122: Writing manifest to image destination
            cluster.go:122: Storing signatures
    --- PASS: crio.base/networks-reliably (27.27s)
            cluster.go:122: STEP 1: FROM scratch
            cluster.go:122: STEP 2: COPY . /
            cluster.go:122: STEP 3: COMMIT localhost/ping
            cluster.go:122: Getting image source signatures
            cluster.go:122: Copying blob sha256:67737d345c27d045d7cc3ebb0c5f2a5131ced2f644dd2bd454fce214c46c92bd
            cluster.go:122: Copying config sha256:6a66872b1e682e0c6b38a7149f39776b0807473d4053ff9cee762d46ddb13413
            cluster.go:122: Writing manifest to image destination
            cluster.go:122: Storing signatures
PASS, output in _kola_temp/qemu-unpriv-2019-10-10-1309-27108
```

ashcrow:
Ready for review

kola/tests/crio/crio.go
```diff
@@ -142,14 +143,28 @@ var crioContainerTemplate = `{
...
 // RHCOS has the crio service disabled by default, so use Ignition to enable it
 var enableCrioIgn = conf.Ignition(`{
   "ignition": {
     "version": "2.2.0"
   },
+  "storage": {
+	"directories": [{
+		"filesystem": "root",
+		"path": "/tmp/test",
+		"mode": 511
```
arithx: Does this need to be `0777`?

cgwalters:
`sudo ./bin/kola --platform qemu-unpriv `

Side note; no reason to use `sudo` with `qemu-unpriv` right?  Also today, `cosa kola` when invoked as non-root will automatically [do the right thing](https://github.com/coreos/coreos-assembler/blob/a1b81c92dbb6d1ad11c634767d90edad16e89413/src/cmd-kola#L54).

kola/tests/crio/crio.go
ashcrow: It probably doesn't HAVE to be `0777` ... `0755` would probably be fine as well.

arithx: I'm okay with it either way. Feel free to change or just merge

PR #1072, opened 2019-10-03
platform/unprivqemu: Drop restrict=yes

cgwalters:
Denying external network access indirectly breaks the `podman.go`
tests because podman uses bridging by default, expecting `/etc/resolv.conf`
to exist on the host and dying messily if it isn't.

I don't know of a reason for us to do this by default - probably
what we should do is something like only use `restrict=yes` for
tests that aren't flagged as requiring network.

But for now, let's make unprivqemu work like the other platforms
by default.  (As much as it can; this uses usermode TCP which has
weird limitations like no ICMP, and bugs of course, but it does
mostly work for what we need)

### DIFF
--- a/platform/machine/unprivqemu/cluster.go
+++ b/platform/machine/unprivqemu/cluster.go
@@ -101,7 +101,7 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
 
 	qc.mu.Lock()
 
-	qmCmd = append(qmCmd, "-netdev", "user,id=eth0,restrict=yes,hostfwd=tcp:127.0.0.1:0-:22", "-device", platform.Virtio(qc.flight.opts.Board, "net", "netdev=eth0"))
+	qmCmd = append(qmCmd, "-netdev", "user,id=eth0,hostfwd=tcp:127.0.0.1:0-:22", "-device", platform.Virtio(qc.flight.opts.Board, "net", "netdev=eth0"))
 
 	plog.Debugf("NewMachine: %q", qmCmd)
 


miabbott:
:+1:  The rationale makes sense to me.  I defer to the SMEs if there are implications for CL or otherwise that I'm not aware of.

ajeddeloh:
Historical context: The qemu tests were always meant to be independent of network so they would be more reliable. If a qemu tests fails it's because something is actually wrong somewhere in our code, not because some registry fell over, your laptop hopped networks, etc. IMO this is valuable and we should preserve it.

That doesn't sound like this is about though. To clarify, the problem here is that with restrict=yes we don't get a dhcp lease with dns, so we don't get an `/etc/resolve.conf` so podman fails (despite not needing to actually access the internet?).

This doesn't allow the host under test to reach the outside internet, right? Assuming that's the case, this LGTM.

jlebon:
> This doesn't allow the host under test to reach the outside internet, right?

It does. The context you provided makes sense. Though at the same time, there's a use case for enabling it for `kola spawn` (could gate behind a switch). So I think there's three categories here:

1. no `resolv.conf`, no external access (most tests)
2. `resolv.conf`, no external access (a few tests)
3. `resolv.conf`, external access (`kola spawn`)

So this patch is flattening out (2) for `unprivqemu` into (3), but it's fixing those tests as well as making `kola spawn` more useful. Maybe let's track refinements on this in a ticket?

cgwalters:
Wouldn't the "run only non-networked tests" be better as `kola run --net=none` or something?

cgwalters:
(And such a flag could be enforced easily on qemu by passing `restrict=yes` or on non-clustered tests, not even using a network interface at all)

cgwalters:
Tangentially related - a really nice-to-have here would be support for caching any artifacts used external to qemu; e.g. support injecting container images from my podman store on the host, etc.

ajeddeloh:
Hmm, yeah we ought to decouple the "needs networking" and qemu. Disabling networking would imply qemu since otherwise we can't talk to clouds.

cgwalters:
Do you see a larger rework of the networking/test/platform stuff as needed for this, or can it land after this PR?

ajeddeloh:
After this is fine

cgwalters:
I don't have merge privs to this repo BTW.

Issue #1068, opened 2019-10-03
[rhcos 4.2][s390x] kola run errors out with " Cluster failed starting machines: didn't find an address"


Prashanth684:
I am running kola on my s390x zVM by using the coreos-assembler (rhcos-4.2 branch) and building an rhcos(4.2) image. The cosa build completes successfully and I can bring up the VM using cosa run. When i execute kola tests, all the tests error out with the error message:

``` 
[coreos-assembler]$ coreos-assembler kola run
+ exec kola -b rhcos --ignition-version v2 -p qemu-unpriv --output-dir tmp/kola --qemu-image builds/42.80.20190927.0/rhcos-42.80.20190927.0-qemu.qcow2 run
=== RUN   coreos.selinux.enforce
=== RUN   rhcos.basic
=== RUN   coreos.selinux.boolean
=== RUN   rhcos.selinux.manage
=== RUN   podman.base
=== RUN   rhcos.selinux.boolean.persist
=== RUN   coreos.ignition.once
=== RUN   podman.network
=== RUN   coreos.auth.verify
=== RUN   coreos.ignition.sethostname
=== RUN   coreos.ignition.groups
=== RUN   rpmostree.status
=== RUN   rpmostree.upgrade-rollback
=== RUN   coreos.ignition.v2.users
=== RUN   ostree.basic
--- FAIL: coreos.selinux.enforce (27.64s)
        harness.go:507: Cluster failed starting machines: didn't find an address
--- FAIL: coreos.ignition.sethostname (29.15s)
        harness.go:507: Cluster failed starting machines: didn't find an address
--- FAIL: rpmostree.upgrade-rollback (29.94s)
        harness.go:507: Cluster failed starting machines: didn't find an address
--- FAIL: rpmostree.status (30.08s)
        harness.go:507: Cluster failed starting machines: didn't find an address
--- FAIL: coreos.ignition.groups (31.68s)
        harness.go:507: Cluster failed starting machines: didn't find an address
--- FAIL: coreos.ignition.v2.users (30.04s)
        harness.go:507: Cluster failed starting machines: didn't find an address
--- FAIL: ostree.basic (31.23s)
        harness.go:507: Cluster failed starting machines: didn't find an address
--- FAIL: podman.network (32.06s)
        harness.go:507: Cluster failed starting machines: didn't find an address
```

It looks like the qemu port could not be found ? In  tmp/kola the journal logs are empty.

arithx:
[...]

PR #1073, opened 2019-10-04
Check endianness when getting the qemu port

Prashanth684:
When getting the qemu port, the list of connections in /proc/net/tcp is checked and the localhost address
is compares against. Ensure that the correct value is used for big and little endian arches

Potential fix for: #1068

### DIFF
--- a/platform/machine/unprivqemu/cluster.go
+++ b/platform/machine/unprivqemu/cluster.go
@@ -163,7 +163,13 @@ func getAddress(pid string) (string, error) {
 		remoteAddress := fields[2]
 		inode := fields[9]
 
-		isLocalPat := regexp.MustCompile("0100007F:[[:xdigit:]]{4}")
+		var isLocalPat *regexp.Regexp
+		if util.HostEndianness == util.LITTLE {
+			isLocalPat = regexp.MustCompile("0100007F:[[:xdigit:]]{4}")
+		} else {
+			isLocalPat = regexp.MustCompile("7F000001:[[:xdigit:]]{4}")
+		}
+
 		if !isLocalPat.MatchString(localAddress) || remoteAddress != "00000000:0000" {
 			continue
 		}
--- a/util/common.go
+++ b/util/common.go
@@ -14,6 +14,35 @@
 
 package util
 
+import (
+	"unsafe"
+)
+
+const (
+	LITTLE Endian = iota // little endian
+	BIG                  // big endian
+)
+
+// Endianness of the platform - big or little
+type Endian int
+
+var HostEndianness Endian
+
+func init() {
+	// Determine endianness - https://stackoverflow.com/questions/51332658/any-better-way-to-check-endianness-in-go
+	buf := [2]byte{}
+	*(*uint16)(unsafe.Pointer(&buf[0])) = uint16(0x0100)
+
+	switch buf {
+	case [2]byte{0x00, 0x01}:
+		HostEndianness = LITTLE
+	case [2]byte{0x01, 0x00}:
+		HostEndianness = BIG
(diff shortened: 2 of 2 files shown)


platform/flight.go
```diff
@@ -41,6 +51,32 @@ func NewBaseFlight(opts *Options, platform Name, ctPlatform string) (*BaseFlight
 	return NewBaseFlightWithDialer(opts, platform, ctPlatform, network.NewRetryDialer())
 }
 
+func GetEndianness() Endian {
```
cgwalters: https://stackoverflow.com/questions/51332658/any-better-way-to-check-endianness-in-go
One of those seems better to me than shelling out to `lscpu` and parsing it...

cgwalters: Also, rather than passing this value down the call stack, I think it'd be better to just determine it as a package `init()` function and export it.

Prashanth684: > https://stackoverflow.com/questions/51332658/any-better-way-to-check-endianness-in-go
> One of those seems better to me than shelling out to `lscpu` and parsing it...

I did try it yesterday but the first solution didn't work, so I just used `lscpu` and didn't look beyond. The solution from the tensor flow library seems to work. Will use that.

Prashanth684: > Also, rather than passing this value down the call stack, I think it'd be better to just determine it as a package `init()` function and export it.

Makes sense. Will do.

platform/platform.go
```diff
@@ -184,6 +194,21 @@ type sshPipe struct {
 	io.Reader
 }
 
+func init() {
+	// Determine endianness
```
cgwalters: I'd add a link to the stackoverflow post to credit it.

Not sure if `platform.go` is the most elegant place to put this but I don't know the code well enough to suggest somethign else.

Prashanth684: done. `platform.go` looked like a good place to put a basic platform specific value like endianness. But again, I am new to this project too so will let @arithx comment.

arithx: With the new implementation I'd probably lean towards having it in `util`. It doesn't really have anything to do with the platform that we're trying to launch but rather is determining something about the host machine, which happens to affects how we need to look for the port number in the `qemu-unpriv` platform.

Prashanth684: OK. in that case i will put it in `util/common.go`.So in that case even the machine specific properties like architecture which are needed to make decisions on selection will be in util in the new implementation?

arithx: Elaborating a little bit further, to me the `util` package isn't the perfect place for it (ideally `go` would have a built-in) but is probably the one that makes the most sense to me.

> So in that case even the machine specific properties like architecture which are needed to make decisions on selection will be in util in the new implementation?

I'm not sure I understand, this isn't a property of the machine being tested but rather the machine that the tests are being ran from. Properties of the machine under test are currently (and will still be) stored inside of the `Options` variable that is passed to the `Flight`.

For example you mentioned `architecture` in your post. There are two different architectures at play here, the host architecture (machine that `kola` is being ran from) and the target architecture (machine under test; currently the `board` variable). The host architecture is only relevant because there are different QEMU commands for different architectures and we need to know which one is available on the system and is currently determined by looking at the `runtime.GOARCH` built-in. The target architecture is a property of the test that we want to run, it can have implications on not only how we might need to spawn the test (e.x.: in `qemu` we need to specify the `cpu` type) but could also impact what tests are ran or how we pull information from the machine.

Does that answer your question?

Prashanth684: Ah...my understanding was off..I assumed that host=target arch. Thank you for the detailed explanation. Makes things much clearer.

PR #1076, opened 2019-10-04
podman: Only run podman network tests on fcos, use --net=none some

cgwalters:
First, use `--net=none|host` where applicable unless we actually
*intend* to test networking.  Podman will by default allocate
a bridge even if all you wanted is e.g. some stdout from a container.

Second, only run the network tests on FCOS by default because
right now RHCOS has both `podman` and `cri-o` installed and
this causes networking issues:
https://bugzilla.redhat.com/show_bug.cgi?id=1757572

We do want the "base" podman sanity tests though since
that's how the OpenShift installer works on the bootstrap node.

### DIFF
--- a/kola/tests/podman/podman.go
+++ b/kola/tests/podman/podman.go
@@ -40,19 +40,27 @@ func init() {
 		Name:        `podman.base`,
 		Distros:     []string{"fcos", "rhcos"},
 	})
+	// These remaining tests use networking, and hence don't work reliably on RHCOS
+	// right now due to due to https://bugzilla.redhat.com/show_bug.cgi?id=1757572
 	register.Register(&register.Test{
 		Run:         podmanWorkflow,
 		ClusterSize: 1,
 		Name:        `podman.workflow`,
 		Flags:       []register.Flag{register.RequiresInternetAccess}, // For pulling nginx
-		Distros:     []string{"fcos", "rhcos"},
+		Distros:     []string{"fcos"},
 		FailFast:    true,
+	})
+	register.Register(&register.Test{
+		Run:         podmanNetworksReliably,
+		ClusterSize: 1,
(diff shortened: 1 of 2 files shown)


kola/tests/podman/podman.go
```diff
@@ -45,14 +45,20 @@ func init() {
...
 		Name:        `podman.workflow`,
 		Flags:       []register.Flag{register.RequiresInternetAccess}, // For pulling nginx
-		Distros:     []string{"fcos", "rhcos"},
+		Distros:     []string{"fcos"}, // Not on rhcos due to https://bugzilla.redhat.com/show_bug.cgi?id=1757572
 		FailFast:    true,
 	})
+	register.Register(&register.Test{
+		Run:         podmanNetworksReliably,
+		ClusterSize: 1,
+		Name:        `podman.network1`,
```
miabbott: nit: Maybe a more descriptive name like `podman.network.simple`

kola/tests/podman/podman.go
```diff
@@ -45,14 +45,20 @@ func init() {
...
+		Distros:     []string{"fcos"}, // Not on rhcos due to https://bugzilla.redhat.com/show_bug.cgi?id=1757572
+	})
 	register.Register(&register.Test{
 		Run:         podmanNetworkTest,
 		ClusterSize: 2,
-		Name:        `podman.network`,
-		Distros:     []string{"fcos", "rhcos"},
+		Name:        `podman.network2`,
```
miabbott: nit:  more descriptive name like `podman.network.full`

miabbott:
Travis CI is failing on `gofmt` check.

I had two nits, but otherwise LGTM

cgwalters:
Renamed and `gofmt`'d.

PR #1080, opened 2019-10-08
kola: Remove podman network-multi test

cgwalters:
I thought this passed at one time, but it seems to just hang
here.  I tried to debug it a little, but the expected flow here
is opaque to me, and it just doesn't seem worth it.

Bigger picture anyways, rather than inventing our own tests for most
upstream projects, we should run *their* tests that have
already been run in *their* CI too.  This is what I plan to
do for ostree/rpm-ostree at least with FCOS.  Basically download
tests from a container built from upstream sources, pull content
onto host and exec there.

### DIFF
--- a/kola/tests/podman/podman.go
+++ b/kola/tests/podman/podman.go
@@ -56,12 +56,13 @@ func init() {
 		Name:        `podman.network-single`,
 		Distros:     []string{"fcos"},
 	})
-	register.Register(&register.Test{
-		Run:         podmanNetworkTest,
-		ClusterSize: 2,
-		Name:        `podman.network-multi`,
-		Distros:     []string{"fcos"},
-	})
+	// https://github.com/coreos/mantle/pull/1080
+	// register.Register(&register.Test{
+	// 	Run:         podmanNetworkTest,
+	// 	ClusterSize: 2,
+	// 	Name:        `podman.network-multi`,
+	// 	Distros:     []string{"fcos"},
+	// })
 }
 
 // simplifiedContainerPsInfo represents a container entry in podman ps -a


cgwalters:
> So to be clear, this isn't necessarily a problem with the test. It could be a podman regression, right?

Yeah, for sure. I just looked at this and I just don't think debugging it is the most valuable thing I could be doing right now.

bgilbert:
I don't have an opinion on this particular test.  But with CL, when we investigate mysterious test failures, we often find that something in the OS did actually break.

If we don't want to pursue this now, what about disabling just the test registration and filing a mantle bug?

arithx:
I'd tend to agree with @bgilbert; For now, if you just want to comment out the `register.Register` call in the `init()` with a link to a bug that would WFM.

I'll add a `Disable` test flag in a follow-up PR that we can use in the future until we nail down a proper way of utilizing version checking for `MinVersion` & `MaxVersion` on *COS.

arithx:
@cgwalters did you mean to close this?

cgwalters:
> @cgwalters did you mean to close this?

Nope, reopened!

Some more background on the podman tests - they were originally only targeted at `rhcos`, and did work for a while.  But as of recently we disabled them there due to the referenced crio conflict.

So one thing I did quickly check is that the problem reproduces even with older FCOS:

`$ kola run -b fcos -p qemu-unpriv --qemu-image /srv/walters/pubannex/machine-images/fedora-coreos-30.20190926.dev.1-qemu.qcow2 podman.network-multi`
also hangs.

I agree we should not jump to disable broken tests as a general rule - in *this* case this test has apparently never passed for FCOS.  Although, based on my git log I would swear I did run it against FCOS but it's totally possible I failed to do that correctly.

PR #1081, opened 2019-10-08
network: Honor MANTLE_SSH_DIR

cgwalters:
I plan to use this for cosa to make it a bit easier to log in
manually to instances created by `cosa kola`.  It will be set
to a predictable path rather than a dir in `/tmp` that one may
have to hunt down.


cgwalters:
> Would also be nice to add a note in the README about it.

Hmm...where?  What would it say?  Mind giving a suggestion?  Or feel free to force-push to this PR of course!

PR #1084, opened 2019-10-09
kola: Change distro default to fcos, drop cl

cgwalters:
Now that CL is branched, let's improve ergonomics here and
default to `fcos` for distro.  This has bit me a few times already
where I ran `kola` directly and forgot the `-b` argument.


cgwalters:
After this I'll do the "remove all the cl tests" bit, just wanted to test the waters and get in this smaller prep PR first.

PR #1087, opened 2019-10-10
kola: Drop torcx

cgwalters:
These only apply to `cl` which is on a branch now.


cgwalters:
Yeah I thought about that...ok, I'll see if I can disentangle torcx from that and see what happens if I just add `fcos` to that list.

PR #1088, opened 2019-10-10
README: update qemu-unpriv notes

arithx:


cgwalters:
In theory I think we could set up [vhost-net](https://www.redhat.com/en/blog/introduction-virtio-networking-and-vhost-net) for example, although since it's not TCP...app support is sporadic.

But anyways, LGTM.

PR #1089, opened 2019-10-10
Drop CL tests & machine modifications

arithx:


cgwalters:
Only skimmed this; I obviously don't have the history with this repo to have a strong review here, but it looks good to me.  There are a few things like the "check user shells" that I'd guess we could run, but eh.  (I think the "check something readonly on the OS" tests are crying out to run as a container in the OS, but that's another topic)

kola/tests/misc/ntp.go
```diff
@@ -1,68 +0,0 @@
...
-)
-
-func init() {
-	register.Register(&register.Test{
-		Run:              NTP,
-		ClusterSize:      0,
-		Name:             "linux.ntp",
-		Platforms:        []string{"qemu"},
-		ExcludePlatforms: []string{"qemu-unpriv"},
-		Distros:          []string{"cl"},
```
ajeddeloh: Is there a reason to not test this on fcos?

kola/tests/misc/nfs.go
```diff
@@ -48,17 +48,6 @@ systemd:
 )
 
 func init() {
-	register.Register(&register.Test{
-		Run:         NFSv3,
-		ClusterSize: 0,
-		Name:        "linux.nfs.v3",
-		Distros:     []string{"cl"},
```
ajeddeloh: Is there a reason to not test this on fcos?

kola/tests/systemd/journald.go
```diff
@@ -1,113 +0,0 @@
-// Copyright 2015 CoreOS, Inc.
```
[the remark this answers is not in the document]
arithx: A quick check converting the config to V3 has it failing due to `/usr/lib/systemd/systemd-journal-remote` not existing.

kola/tests/misc/ntp.go
arithx: FCOS just ships the client, not nfs-server (https://github.com/coreos/fedora-coreos-tracker/issues/121)

kola/tests/misc/nfs.go
arithx: FCOS just ships the client, not nfs-server (https://github.com/coreos/fedora-coreos-tracker/issues/121)

kola/tests/misc/ntp.go
arithx: Having a general NTP test would be good, however this specific NTP test isn't usable on FCOS and would require an entire re-write.

kola/tests/systemd/journald.go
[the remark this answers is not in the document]
jlebon: Ahh interesting. OK, I filed https://github.com/coreos/fedora-coreos-tracker/issues/294 to discuss this.

kola/tests/ignition/empty.go
```diff
@@ -1,58 +0,0 @@
...
-	"github.com/coreos/mantle/kola/register"
-	"github.com/coreos/mantle/platform/conf"
-)
-
-// These tests require the kola key to be passed to the instance via cloud
-// provider metadata since it will not be injected into the config. Platforms
-// where the cloud provider metadata system is not available have been excluded.
-func init() {
-	// Tests for https://github.com/coreos/bugs/issues/1184
-	register.Register(&register.Test{
```
bgilbert: Shouldn't we retain these tests?

kola/tests/metadata/contents.go
```diff
@@ -1,95 +0,0 @@
...
-		    {
-			"name": "metadata.target",
-			"enable": true,
-			"contents": "[Install]\nWantedBy=multi-user.target"
-		    }
-		]
-	    }
-	}`)
-
-	register.Register(&register.Test{
```
bgilbert: Shouldn't we convert these for Afterburn?

kola/tests/misc/files.go
```diff
@@ -1,212 +0,0 @@
...
-func init() {
-	register.Register(&register.Test{
-		Run:         Filesystem,
-		ClusterSize: 1,
-		Name:        "cl.filesystem",
-		Distros:     []string{"cl"},
-	})
-}
-
-func Filesystem(c cluster.TestCluster) {
```
bgilbert: These all seem like useful checks for FCOS/RHCOS as well.

kola/tests/misc/network.go
```diff
@@ -1,168 +0,0 @@
...
-	"github.com/coreos/go-semver/semver"
-
-	"github.com/coreos/mantle/kola/cluster"
-	"github.com/coreos/mantle/kola/register"
-	"github.com/coreos/mantle/platform/conf"
-	"github.com/coreos/mantle/util"
-)
-
-func init() {
-	register.Register(&register.Test{
```
bgilbert: These would both be good tests to carry forward.

kola/tests/misc/raid.go
```diff
@@ -1,202 +0,0 @@
...
-		Run:         RootOnRaid,
-		ClusterSize: 0,
-		Platforms:   []string{"qemu"},
-		Name:        "cl.disk.raid.root",
-		Distros:     []string{"cl"},
-	})
-	register.Register(&register.Test{
-		Run:         DataOnRaid,
-		ClusterSize: 1,
-		Name:        "cl.disk.raid.data",
```
bgilbert: We could still keep this test, no?

cgwalters:
I guess the question here is whether or not we keep tests that may be applicable, but need porting in the source tree or not.  I don't have a really strong opinion on this to be honest.

However, let's at least identify some more of the actual cruft that is deleted by this PR and get that merged?

arithx:
> However, let's at least identify some more of the actual cruft that is deleted by this PR and get that merged?

Yeah, I've been meaning to get back to this and it keeps slipping my mind. I'll try and make another pass soon.

arithx:
Rebased & updated; with the newest patch set there's still a couple of remnant tests that need to be fixed (FCOS NFS ended up being a time sync that I wasn't able to solve & the DockerPing/DockerEcho from cl.internet didn't plug & play nicely [swapped to podman temporarily and left the docker variants unused])

cgwalters:
I skimmed this - LGTM.  I totally get the feeling that we should try to preserve tests where we can.  On the flip side, I think the elimination of tech debt here and consequent ability to more easily write *new* tests weights in favor.

Concretely for example, if we fully stop thinking about "cl" in git master, we can e.g. make "fcos", "rhcos" the default for tests, and start writing "coreos" tests for things common between the two and unify more things, etc.

PR #1094, opened 2019-10-14
platform/aliyun: add force image deletion & multipart upload

arithx:
Specifying the `--force` flag to `ore aliyun create-image` will now also delete images (and their corresponding snapshots) rather than failing to a `DuplicateImage` error.

Image uploading is now also using a new `UploadFile` API function that performs a multipart upload (1000K part size & 10 coroutines) to reduce the upload time of a 1.6GB object from 30m to 8m.


platform/api/aliyun/api.go
```diff
@@ -152,16 +167,82 @@ func (a *API) ImportImage(format, bucket, object, image_size, device, name, desc
 	return response.ImageId, nil
 }
 
-func (a *API) DeleteImage(image string, force bool) error {
+func (a *API) GetImages(name string) (*ecs.DescribeImagesResponse, error) {
```
ashcrow: nit: Doc public function/method

platform/api/aliyun/api.go
```diff
@@ -152,16 +167,82 @@ func (a *API) ImportImage(format, bucket, object, image_size, device, name, desc
...
+}
+
+func (a *API) GetImagesByID(id string) (*ecs.DescribeImagesResponse, error) {
+	request := ecs.CreateDescribeImagesRequest()
+	request.Scheme = "https"
+	request.ImageId = id
+	return a.ecs.DescribeImages(request)
+}
+
+func (a *API) DeleteImage(id string, force bool) error {
```
ashcrow: nit: Doc public function/method

platform/api/aliyun/api.go
```diff
@@ -152,16 +167,82 @@ func (a *API) ImportImage(format, bucket, object, image_size, device, name, desc
...
+			err = a.DeleteSnapshot(mapping.SnapshotId, force)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("deleting snapshot %v: %v", mapping.SnapshotId, err))
+			}
+		}
+	}
+	return errs.AsError()
+}
+
+func (a *API) DeleteSnapshot(id string, force bool) error {
```
ashcrow: nit: Doc public function/method

arithx:
Pushed an update with docstrings for public methods in `platform/api/aliyun`, merging on green CI.

PR #1096, opened 2019-10-15
WIP: Rework architecture, remove cork and gangue, remove kola dependency on sdk

Prashanth684:
A WIP. 

- Remove reference to board in all places and replace with architecture (although there are some urls which use "board" or "-usr" which I am not sure should be changed or maybe they will be removed because they only apply to CL)

- Drop cork and gangue

- Remove sdk dependency from kola

-  Cleanup sdk directory and remove CL support from plume


cmd/plume/containerlinux.go
```diff
@@ -96,8 +97,8 @@ var (
 	}
 	specs = map[string]channelSpec{
 		"user": channelSpec{
-			BaseURL: "gs://users.developer.core-os.net/" + os.Getenv("USER") + "/boards",
-			Boards:  []string{"amd64-usr"},
+			BaseURL:       "gs://users.developer.core-os.net/" + os.Getenv("USER") + "/boards",
```
Prashanth684: how do we deal with the URLs which have `board` or `-usr` in them ? I assume some will go away because they are specific to CL?

bgilbert: I'd expect that they all will.  Anything at `core-os.net` is for Container Linux.  In this particular case, all of `containerlinux.go` can (hopefully) be dropped.

cgwalters:
Smaller, incremental PRs can be more work but are *much* more likely to land.  Mind splitting this up some?  For example, removing cork/gangue is a clearly distinct thing.

Prashanth684:
> Smaller, incremental PRs can be more work but are _much_ more likely to land. Mind splitting this up some? For example, removing cork/gangue is a clearly distinct thing.

right..just thought i could squeeze that in here while i'm at it...but i agree

PR #1097, opened 2019-10-15
kola/tests: exclude all coreos-cloudinit tests on Packet

bgilbert:
Avoid more tcsd.service flakes.

### DIFF
--- a/kola/tests/misc/grub.go
+++ b/kola/tests/misc/grub.go
@@ -147,13 +147,15 @@ coreos:
 
 func init() {
 	register.Register(&register.Test{
-		Run:              UpdateGrubNop,
-		ClusterSize:      1,
-		Name:             "cl.update.grubnop",
-		UserData:         grubUpdaterConf,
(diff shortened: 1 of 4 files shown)


bgilbert:
We should also rebase the temporary jenkins4 branch onto this.

kola/tests/misc/grub.go
```diff
@@ -147,13 +147,15 @@ coreos:
...
-		Name:             "cl.update.grubnop",
-		UserData:         grubUpdaterConf,
-		MinVersion:       semver.Version{Major: 1745},
-		Distros:          []string{"cl"},
-		ExcludePlatforms: []string{"qemu-unpriv"},
+		Run:         UpdateGrubNop,
+		ClusterSize: 1,
+		Name:        "cl.update.grubnop",
+		UserData:    grubUpdaterConf,
+		MinVersion:  semver.Version{Major: 1745},
```
arithx: Unrelated but can probably drop this.

PR #1099, opened 2019-10-17
cmd/ore: add command to list enabled AWS regions

bgilbert:
For use by `cosa aws-replicate` to avoid hardcoding the region list there.


cmd/ore/aws/list-regions.go
```diff
@@ -0,0 +1,49 @@
...
+
+func runListRegions(cmd *cobra.Command, args []string) error {
+	regions, err := API.ListRegions()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "could not list regions: %v\n", err)
+		os.Exit(1)
+	}
+
+	for _, region := range regions {
+		fmt.Println(region)
```
arithx: WDYT about producing a JSON list rather than one region per line?

platform/api/aws/ec2.go
```diff
@@ -27,6 +28,21 @@ import (
...
+func (a *API) ListRegions() ([]string, error) {
+	output, err := a.ec2.DescribeRegions(nil)
+	if err != nil {
+		return nil, fmt.Errorf("describing regions: %v", err)
+	}
+	ret := make([]string, 0, len(output.Regions))
+	for _, region := range output.Regions {
+		ret = append(ret, *region.RegionName)
+	}
+	sort.Strings(ret)
```
arithx: Hmm, any particular reason for sorting here rather than letting the caller do that if they need the list sorted?

bgilbert: It seemed odd to return a list in random order.  But I can drop it if desired.

bgilbert:
> What do you think about the idea of changing this code in copy-images to default to the output of API.ListRegions() if a list isn't given?

That feels a bit too footgun for me, but I'm +1 to adding an `--all-regions` argument.

arithx:
> I'm +1 to adding an `--all-regions` argument.

Fair, I was thinking from the non-scripted developer use case it'd be nice to have it in `copy-image` directly. That's a good point about defaulting to it being a potential foot gun.

cmd/ore/aws/list-regions.go
arithx: No, not necessarily. I was more thinking from the perspective of this tool (in my eyes) mostly being useful in the context of being ran by scripts where I'd generally personally prefer to have a JSON output than parsing string outputs. I'm fine with leaving it as is if you feel it's too weighty for this tool.

platform/api/aws/ec2.go
arithx: Is there any particular value derived from the regions being in a sorted order? Nothing immediately struck out to me so I figured I'd ask if it's worth not spending the extra cycles sorting them always; it should never really become too cumbersome (unless AWS gets far too many regions) so I'm fine with keeping it if you feel there's value.

PR #1105, opened 2019-10-22
auth/aliyun: fix parsing config file

jlebon:
The code is expecting `~/.aliyun/config.json` to be a list of profiles,
but the one I have (that was created by the `aliyun` CLI) is an object
where the `profiles` key is a list of profile objects:

```
{
  "current": "default",
  "profiles": [
    {
      "name": "default",
      "access_key_id": "...",
      "access_key_secret": "...",
      "region_id": "...",
      ...
    }
  ],
  "meta_path": ""
}
```

Fix the code to adapt to this.

### DIFF
--- a/auth/aliyun.go
+++ b/auth/aliyun.go
@@ -31,6 +31,10 @@ type AliyunProfile struct {
 	Region          string `json:"region_id"`
 }
 
+type AliyunConfig struct {
+	Profiles []AliyunProfile `json:"profiles"`
+}
+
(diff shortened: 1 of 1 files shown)


jlebon:
Not sure how this happened. Are there two different formats out there?

arithx:
> Not sure how this happened. Are there two different formats out there?

Ah, that'd be my fault, rather than generating one using the aliyun CLI I mocked it based on the format listed in the pip package.

Issue #1106, opened 2019-10-23
`ore aliyun create-image` missing `--delete-object` switch


jlebon:
We should match the AWS default of deleting the uploaded image after the import is complete.

PR #1107, opened 2019-10-23
ore/aliyun: add --delete-object and default to true

jlebon:
Match what `ore aws upload` does and delete the OSS file by default
after the image was successfully uploaded. Then the GC strategy for
pipelines is just about cleaning up old images, and we don't have to
worry about the bucket too.

Closes: #1106


jlebon:
Marking as draft; still testing this.

jlebon:
OK, tested working! :heavy_check_mark:

jlebon:
Rebased!

jlebon:
Green CI; shipping!

PR #1113, opened 2019-10-24
platform/api/aliyun: wait until import is complete

jlebon:
Similarly to the AWS analogue of this command, wait until the import
task is complete before returning. Otherwise, e.g. we'll end up deleting
the OSS object beforehand and the task will naturally fail.

### DIFF
--- a/platform/api/aliyun/api.go
+++ b/platform/api/aliyun/api.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io"
 	"sort"
+	"time"
 
 	"github.com/coreos/mantle/auth"
 	"github.com/coreos/mantle/platform"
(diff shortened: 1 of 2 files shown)


jlebon:
The tricky bit about this is that without it, everything appears fine. The API returns an image ID pretty much immediately. But you can't actually use it to create an instance because it's not ready yet. And the snapshot will be marked as "Failed" in the web UI.

jlebon:
OK, tested working well in RHCOS!

platform/api/aliyun/api.go
```diff
@@ -168,7 +169,54 @@ func (a *API) ImportImage(format, bucket, object, image_size, device, name, desc
...
+			return false, fmt.Errorf("task %v failed", taskId)
+		default:
+			return false, fmt.Errorf("unexpected status for task %v: %v", taskId, res.TaskSet.Task[0].TaskStatus)
+		}
+	}
+
+	for {
+		var done bool
+		var err error
+		done, err = importDone(importImageResponse.TaskId)
```
cgwalters: nonblocking:
Why not use `:=` here?

mike-nguyen:
LGTM

platform/api/aliyun/api.go
jlebon: So, this code is actually based on AWS' `finishSnapshotTask()`, which also does this. Though in their case, they have to use `=` because of variable in an outer scope. We don't, so I switched this one here to `:=`.

jlebon:
Added docstrings and cross-references between the two similar function in both Aliyun and AWS. Could maybe dedupe some things, though not sure it's worth it.

Merging after CI!

PR #1116, opened 2019-11-11
azure: support removing images in ore

### LINKED ISSUE #1115: Add ore command to remove Azure image
# Feature Request #

Pipelines require ore subcommand to remove developer / testing images from Azure. It appears `unreplicate-image` won't do sufficient

vrutkovs:
This PR adds a new subcommand to remove uploaded image on Azure

Fixes #1115

### DIFF
--- a/cmd/ore/azure/delete-image.go
+++ b/cmd/ore/azure/delete-image.go
@@ -1,0 +1,56 @@
+// Copyright 2019 Red Hat
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


cmd/ore/azure/delete-image.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	cmdDeleteImage = &cobra.Command{
+		Use:   "delete-image",
```
arithx: Should be `delete-image-arm` as it's using the ARM-based commands not the ASM ones.

vrutkovs: Fixed

PR #1118, opened 2019-11-12
Support LUKS devices when using guestfish to inject ignition

Prashanth684:
Add support for recognizing LUKS filesystem when setting up ignition.
While not necessary for x86 because it uses fw_cfg, ppc64 and s390x use guestfish
and need this change for kola tests to run.

### DIFF
--- a/platform/qemu.go
+++ b/platform/qemu.go
@@ -15,6 +15,7 @@
 package platform
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -395,7 +396,8 @@ func setupIgnition(confPath string, diskImagePath string) error {
 	// Using libvirt can lead to permission denied issues if it does not have access
 	// rights to the qcow image
 	os.Setenv("LIBGUESTFS_BACKEND", "direct")
-	cmd := exec.Command("guestfish", "--listen", "-a", diskImagePath)
+	cmd := exec.Command("guestfish", "--listen", "--keys-from-stdin", "--key", "/dev/sda4:file:/dev/null", "-a", diskImagePath)
+	cmd.Stdin = bytes.NewBuffer([]byte("nokey"))
 	stdout, err := cmd.StdoutPipe()
 	if err != nil {
 		return fmt.Errorf("getting stdout pipe: %v", err)
@@ -440,6 +442,17 @@ func setupIgnition(confPath string, diskImagePath string) error {
 		return fmt.Errorf("guestfish command failed to find boot label: %v", err)
 	}
 
+	fstype, err := findFsType("/dev/sda4", pid)
+	if err != nil {
+		return fmt.Errorf("guestfish command failed to find fs type: %v", err)
+	}
+	if fstype == "crypto_LUKS" {
+		cmd := exec.Command("guestfish", remote, "luks-open", "/dev/sda4", "luks-00000000-0000-4000-a000-000000000002")
+		if lukserr := cmd.Run(); lukserr != nil {
+			return fmt.Errorf("guestfish command failed to execute luks-open: %v", lukserr)
+		}
+	}
+
 	rootfs, err := findLabel("root", pid)
 	if err != nil {
 		return fmt.Errorf("guestfish command failed to find root label: %v", err)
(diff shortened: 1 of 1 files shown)


Prashanth684:
Could i get some reviews for this please ? and a lgtm hopefully ? :)

platform/qemu.go
```diff
@@ -395,7 +396,8 @@ func setupIgnition(confPath string, diskImagePath string) error {
 	// Using libvirt can lead to permission denied issues if it does not have access
 	// rights to the qcow image
 	os.Setenv("LIBGUESTFS_BACKEND", "direct")
-	cmd := exec.Command("guestfish", "--listen", "-a", diskImagePath)
+	cmd := exec.Command("guestfish", "--listen", "--keys-from-stdin", "--key", "/dev/sda4:file:/dev/null", "-a", diskImagePath)
```
arithx: Is hardcoding `/dev/sda4` correct?

platform/qemu.go
```diff
@@ -467,6 +480,24 @@ func setupIgnition(confPath string, diskImagePath string) error {
 	return nil
 }
 
+// findPartName finds the partition name to see if it is a luks partition or a regular root partition.
```
arithx: wrong function name in this comment.

jlebon:
Would be good to link to the analogous coreos-assembler change in the commit message: https://github.com/coreos/coreos-assembler/pull/826.

One thing here, do we need to bump memory here as well? https://github.com/coreos/coreos-assembler/pull/826/files#diff-b67c90828c71ffa6aaa6454fa26aefa4R9-R15

platform/qemu.go
jlebon: Yeah, see https://github.com/coreos/coreos-assembler/blob/92a8ee879b8a8cb94981e679b92be9af6c19158c/src/create_disk.sh#L88. A more robust approach though would be to look for the root label. But let's not write that code before discussing some way to share this between cosa and mantle. :)

Prashanth684: I can pass it in to the  `setupIgnition` function , but i guess it needs to be passed in someplace. the coreos-assembler has the root partition hardcoded as 4, so I guess we should be fine unless that changes.

Prashanth684:
> Would be good to link to the analogous coreos-assembler change in the commit message: [coreos/coreos-assembler#826](https://github.com/coreos/coreos-assembler/pull/826).
> 
> One thing here, do we need to bump memory here as well? https://github.com/coreos/coreos-assembler/pull/826/files#diff-b67c90828c71ffa6aaa6454fa26aefa4R9-R15

Will add the reference to the commit. I am not sure about the memory increase. I was able to run the kola tests without that change and seemed to work fine. @cgwalters is the memory increase for guestfs needed for the kola tests?

platform/qemu.go
Prashanth684: oops..will fix

platform/qemu.go
Prashanth684: @jlebon yes, my knowledge of guestfish is limited but i was also wondering if there is a way to do that..

cgwalters:
This seems OK to me, but I think it'd be simpler to just stop mounting the rootfs - we don't need it to inject Ignition into `/boot` right?

tuan-hoang1:
> This seems OK to me, but I think it'd be simpler to just stop mounting the rootfs - we don't need it to inject Ignition into `/boot` right?

Second this. So we can just remove the rootfs mounting after all - `/boot` is all we need.

Prashanth684:
> > This seems OK to me, but I think it'd be simpler to just stop mounting the rootfs - we don't need it to inject Ignition into `/boot` right?
> 
> Second this. So we can just remove the rootfs mounting after all - `/boot` is all we need.

Yes. we could mount /boot using the guestfish debug shell, but one roadblock I hit was a way to upload a file to the partition in the debug shell. I am not a guestfish expert by any means so if someone knows of a way to do this that would be great.

tuan-hoang1:
Using guestfish's mount command with boot partition (/dev/sda1) being as `/` instead of `/boot` if you don't want to mount rootfs (/dev/sda4) at `/`. Yeah it looks a little bit misleading but works:

```
guestfish --remote -- mount /dev/sda1 /
guestfish --remote -- mkdir-p /ignition
guestfish --remote -- upload ign.conf /ignition/ign.conf
```
Thinking on how to do `guestfish debug sh` in a nice way if you don't want above solution.

Prashanth684:
> Using guestfish's mount command with boot partition (/dev/sda1) being as `/` instead of `/boot` if you don't want to mount rootfs (/dev/sda4) at `/`. Yeah it looks a little bit misleading but works:
> 
> ```
> guestfish --remote -- mount /dev/sda1 /
> guestfish --remote -- mkdir-p /ignition
> guestfish --remote -- upload ign.conf /ignition/ign.conf
> ```
> 
> Thinking on how to do `guestfish debug sh` in a nice way if you don't want above solution.

i think this is much better. It is misleading in the sense that we are not putting it in "/", but it is a cleaner solution. i will close this PR and open a new one. thanks @tuan-hoang1 !

Prashanth684:
opened #1124.

PR #1120, opened 2019-11-19
kola/qemu: Add a swtpm by default

cgwalters:
The duplication in qemu code between cosa and kola needs
to end - this is the start of moving some bits into kola.
Eventually we'll have `cosa run` be a wrapper for something
like `kola qemuexec`.

This bit is the equivalent of
https://github.com/coreos/coreos-assembler/pull/878/commits/55c9b651e5ee94fb4de947a943df213ff3c601eb

We want to strongly emphasize use of TPM devices in CoreOS, so let's provision
one by default.

They can be used for disk encryption, secure secret storage, etc.

### DIFF
--- a/platform/machine/unprivqemu/cluster.go
+++ b/platform/machine/unprivqemu/cluster.go
@@ -91,6 +91,28 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
 		consolePath: filepath.Join(dir, "console.txt"),
 	}
 
+	qm.swtpmTmpd, err = ioutil.TempDir("", "kola-swtpm")
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


cgwalters:
We don't actually use `qemu` anymore do we?  I am increasingly thinking that that path should really be `libvirt`.

cgwalters:
Does anyone mind if I do a PR to make `qemu-unpriv` an alias for `qemu`?

cgwalters:
> No reason not to also do this for the qemu platform, right?

Is this considered a blocker?

jlebon:
> Is this considered a blocker?

It's not, though I'd like at least @ajeddeloh or @arithx to have a look too.

cgwalters:
I don't have merge access btw (should I?)

PR #1122, opened 2019-11-20
build-sys: Don't cross-build kolet by default

cgwalters:
We're not using this for FCOS, and the cross build for kolet
was slowing down local build iteration times.

### DIFF
--- a/build
+++ b/build
@@ -24,6 +24,7 @@ host_build() {
 		"${REPO_PATH}/cmd/$1"
 }
 
+# Unused now, but kept in case we want it in the future
 cross_build() {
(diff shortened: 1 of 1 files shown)


bgilbert:
> I suspect this was on by default because CI for CL would do cross-arch tests? We don't do that for FCOS because we don't support cross-composes (and ideally we'll never have to). Anyway, would be nicer to gate this behind some flag if we do bring it back for whatever reason.

I agree we don't need it at the moment.  Eventually it'll be nice to be able to e.g. test ARM on AWS or Packet without running kola from an ARM box.

PR #1123, opened 2019-11-20
 kola: Add a test for RHCOS with LUKS+TPM encryption

cgwalters:
Depends https://github.com/coreos/mantle/pull/1120

See https://github.com/openshift/enhancements/blob/b5e77b5a99dc19de9acfa27fb0758ca42d74f3ee/enhancements/automated-policy-based-disencryption.md

### DIFF
--- a/kola/registry/registry.go
+++ b/kola/registry/registry.go
@@ -16,6 +16,7 @@ import (
 	_ "github.com/coreos/mantle/kola/tests/ostree"
 	_ "github.com/coreos/mantle/kola/tests/packages"
 	_ "github.com/coreos/mantle/kola/tests/podman"
+	_ "github.com/coreos/mantle/kola/tests/rhcos"
 	_ "github.com/coreos/mantle/kola/tests/rkt"
 	_ "github.com/coreos/mantle/kola/tests/rpmostree"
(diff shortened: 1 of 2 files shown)


kola/tests/rhcos/luks.go
```diff
@@ -0,0 +1,76 @@
...
+	})
+}
+
+func mustMatch(c cluster.TestCluster, r string, output []byte) {
+	m, err := regexp.Match(r, output)
+	if err != nil {
+		c.Fatal("Failed to match regexp %s: %v", r, err)
+	}
+	if !m {
+		c.Fatal("Regexp %s did not match", r)
```
arithx: Might also be useful to include the output in the error message.

kola/registry/registry.go
```diff
@@ -7,6 +7,7 @@ import (
 	_ "github.com/coreos/mantle/kola/tests/docker"
 	_ "github.com/coreos/mantle/kola/tests/etcd"
 	_ "github.com/coreos/mantle/kola/tests/fips"
+	_ "github.com/coreos/mantle/kola/tests/rhcos"
```
arithx: gofmt requires imports to be alphabetically ordered

cgwalters:
Updated and comments addressed!

PR #1124, opened 2019-11-21
Do not mount root parition for injecting ignition

Prashanth684:
With the changes made recently: https://github.com/coreos/coreos-assembler/pull/826, we need to add
support for detecting LUKS root partitions which is really unnecessary as we are just injecting ignition
in the boot partition.

Based on suggestions by @cgwalters and @tuan-hoang1 in #1118


cgwalters:
Yep, a lot simpler!  LGTM, thanks!

PR #1125, opened 2019-11-21
Drop cork and gangue and Makefile tweaks

jlebon:
See individual commit messages.


Makefile
```diff
@@ -16,7 +16,7 @@ build:
 .PHONY: install
 install: build
 	install -D -t $(DESTDIR)$(PREFIX)/bin bin/{ore,kola,plume}
-	install -D -m 0755 -t $(DESTDIR)$(PREFIX)/lib/kola/$(GOARCH) bin/$(GOARCH)/kolet
+	install -D -m 0755 -t $(DESTDIR)$(PREFIX)/lib/kola/$(GOARCH) bin/kolet
```
cgwalters: Oh ugh, sorry - it looks like this worked for me because my build tree was dirty with the binary in the old place.

Makefile
```diff
@@ -14,7 +14,7 @@ build:
 	./build
 
 .PHONY: install
-install: build
+install: bin/ore bin/kola bin/plume bin/kolet
```
cgwalters: OK by me, though at this point we might consider extracting a
`COMPONENTS := kola plume kolet`
variable and using it in multiple places.

jlebon: Yeah makes sense. Let's do that as a follow-up!

bgilbert:
To provide a bit more context: `cork download-image` became `coreos-installer download`, the other `cork` commands are specific to CL, and we won't need `gangue` because it's really just a wrapper for downloading individual binpkgs from GCS and checking their signatures.

PR #1126, opened 2019-11-27
platform/qemu: Drop some CL disk customization bits

cgwalters:
Prep for further work in qemu.

### DIFF
--- a/cmd/kola/kola.go
+++ b/cmd/kola/kola.go
@@ -173,8 +173,7 @@ func writeProps() error {
 		ImageURL              string `json:"image"`
 	}
 	type QEMU struct {
-		Image   string `json:"image"`
-		Mangled bool   `json:"mangled"`
+		Image string `json:"image"`
 	}
 	return enc.Encode(&struct {
 		Cmdline         []string  `json:"cmdline"`
@@ -237,8 +236,7 @@ func writeProps() error {
 			ImageURL:              kola.PacketOptions.ImageURL,
 		},
 		QEMU: QEMU{
-			Image:   kola.QEMUOptions.DiskImage,
-			Mangled: !kola.QEMUOptions.UseVanillaImage,
+			Image: kola.QEMUOptions.DiskImage,
 		},
 	})
 }
--- a/platform/machine/qemu/flight.go
+++ b/platform/machine/qemu/flight.go
@@ -16,14 +16,10 @@
 package qemu
 
 import (
-	"fmt"
-	"os"
-
 	"github.com/coreos/pkg/capnslog"
 
 	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/platform/local"
-	"github.com/coreos/mantle/util"
 )
 
 const (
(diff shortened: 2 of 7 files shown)


cgwalters:
This is also part of https://github.com/coreos/mantle/pull/1089 but that one stalled.

platform/machine/qemu/flight.go
```diff
@@ -70,37 +65,6 @@ func NewFlight(opts *Options) (platform.Flight, error) {
...
-		if info.Format != "raw" {
-			// platform.MakeCLDiskTemplate() needs to be able to mount
-			// partitions
-			plog.Debug("disk image is in qcow format; not enabling console logging")
-			opts.UseVanillaImage = true
-		}
-	}
-	if !opts.UseVanillaImage {
-		plog.Debug("enabling console logging in base disk")
-		qf.diskImageFile, err = platform.MakeCLDiskTemplate(opts.DiskImage)
```
bgilbert: This is the only assignment to `diskImageFile`, so that member can be dropped.  It looks like it was copypasted into `unprivqemu`, also.

platform/machine/qemu/flight.go
```diff
@@ -70,37 +65,6 @@ func NewFlight(opts *Options) (platform.Flight, error) {
...
-		if err != nil {
-			qf.Destroy()
-			return nil, err
-		}
-		// The template file has already been deleted, ensuring that
-		// it will be cleaned up on exit.  Use a path to it that
-		// will remain stable for the lifetime of the flight without
-		// extra effort to pass FDs to subprocesses.
-		qf.diskImagePath = fmt.Sprintf("/proc/%d/fd/%d", os.Getpid(), qf.diskImageFile.Fd())
```
bgilbert: This means the `diskImagePath` member can be dropped in favor of direct access to `opts.DiskImage`.  Same in `unprivqemu`.

platform/machine/qemu/flight.go
cgwalters: All gone! :recycle:

platform/machine/unprivqemu/flight.go
```diff
@@ -73,7 +70,4 @@ func (qf *flight) NewCluster(rconf *platform.RuntimeConfig) (platform.Cluster, e
 }
 
 func (qf *flight) Destroy() {
```
bgilbert: There's no need for an empty method here; this can just be dropped.

platform/machine/qemu/flight.go
```diff
@@ -124,7 +85,4 @@ func (qf *flight) NewCluster(rconf *platform.RuntimeConfig) (platform.Cluster, e
 
 func (qf *flight) Destroy() {
```
bgilbert: I believe this method can be dropped and we'll inherit the embedded `LocalFlight`'s `Destroy` method.

cmd/kola/kola.go
```diff
@@ -237,8 +237,7 @@ func writeProps() error {
 			ImageURL:              kola.PacketOptions.ImageURL,
 		},
 		QEMU: QEMU{
-			Image:   kola.QEMUOptions.DiskImage,
-			Mangled: !kola.QEMUOptions.UseVanillaImage,
```
bgilbert: Whoops, just noticed this: the member definition should be dropped too.

bgilbert:
and gofmt is failing.

cmd/kola/kola.go
cgwalters: :recycle:

PR #1128, opened 2019-11-28
ignition/mount.go: add tests for mounting additional disks

zonggen:
Adds tests for mounting additional disks at /var/lib/containers
and /var/log. Instead of mounting new disks, tests add new partitions
under the disk `vda`. Currently, there are four partitions, i.e. vda1
-vda4, so new partitions will start from vda5. Particularly, vda5 ->
/var/lib/containers, and vda6 -> /var/log.

~Before running the test, add 2G more space to the disk `vda` since
`vda5` and `vda6` will each take 1G space. Detailed explanations can
be found at https://gist.github.com/zonggen/d888b1ece140f8592a70137d35050073~

Closes: https://jira.coreos.com/browse/GRPA-1203

EDIT: barred adding space part

### DIFF
--- a/kola/tests/ignition/mount.go
+++ b/kola/tests/ignition/mount.go
@@ -1,0 +1,437 @@
+// Copyright 2019 Red Hat
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ignition
+
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform"
+	"github.com/coreos/mantle/platform/conf"
+	"github.com/coreos/mantle/platform/machine/qemu"
+	"github.com/coreos/mantle/platform/machine/unprivqemu"
+)
+
+var (
+	v2IgnitionConfig = conf.Ignition(`{
+		"ignition": {
+			"version": "2.2.0"
+		},
+		"storage": {
+			"disks": [
+				{
+					"device": "/dev/disk/by-id/virtio-secondary-disk",
+					"wipeTable": true,
+					"partitions": [
+						{
+							"label": "CONTR",
+							"size": 1048576,
+							"start": 0
+						}
+					]
+				},
+				{
+					"device": "/dev/disk/by-id/virtio-tertiary-disk",
+					"wipeTable": true,
+					"partitions": [
+						{
+							"label": "LOG",
+							"size": 1048576,
+							"start": 0
+						}
+					]
+				}
+			],
+			"filesystems": [
+				{
+					"name": "CONTR",
+					"mount": {
+						"device": "/dev/disk/by-partlabel/CONTR",
+						"format": "xfs",
+						"wipeFilesystem": true
+					}
+				},
+				{
+					"name": "LOG",
+					"mount": {
+						"device": "/dev/disk/by-partlabel/LOG",
+						"format": "xfs",
+						"wipeFilesystem": true
+					}
+				}
+			],
+			"files": [
+				{
+					"filesystem": "CONTR",
+					"path": "/hello.txt",
+					"contents": {
+						"source": "data:,hello%20world%0A"
+					},
+					"mode": 420
+				},
+				{
+					"filesystem": "LOG",
+					"path": "/hello.txt",
+					"contents": {
+						"source": "data:,hello%20world%0A"
+					},
+					"mode": 420
+				}
+			]
+		},
+		"systemd": {
+			"units": [
+				{
+					"name": "var-lib-containers.mount",
+					"enabled": true,
+					"contents": "[Mount]\nWhat=/dev/disk/by-partlabel/CONTR\nWhere=/var/lib/containers\nType=xfs\nOptions=defaults\n[Install]\nWantedBy=local-fs.target"
+				},
+				{
+					"name": "var-log.mount",
+					"enabled": true,
+					"contents": "[Mount]\nWhat=/dev/disk/by-partlabel/LOG\nWhere=/var/log\nType=xfs\nOptions=defaults\n[Install]\nWantedBy=local-fs.target"
+				}
+			]
+		}
+	}`)
+	v3IgnitionConfig = conf.Ignition(`{
+		"ignition": {"version": "3.0.0"},
+		"storage": {
+			"disks": [
+				{
+					"device": "/dev/disk/by-id/virtio-secondary-disk",
+					"wipeTable": true,
+					"partitions": [
+						{
+							"label": "CONTR",
+							"sizeMiB": 512,
+							"startMiB": 0,
+							"wipePartitionEntry": true
+						}
+					]
+				},
+				{
+					"device": "/dev/disk/by-id/virtio-tertiary-disk",
+					"wipeTable": true,
+					"partitions": [
+						{
+							"label": "LOG",
+							"sizeMiB": 512,
+							"startMiB": 0,
+							"wipePartitionEntry": true
+						}
+					]
+				}
+			],
+			"filesystems": [
+				{
+					"path": "/var/lib/containers",
+					"device": "/dev/disk/by-partlabel/CONTR",
+					"format": "xfs",
+					"label": "CONTR",
+					"wipeFilesystem": true
+				},
+				{
+					"path": "/var/log",
+					"device": "/dev/disk/by-partlabel/LOG",
+					"format": "xfs",
+					"label": "LOG",
+					"wipeFilesystem": true
+				}
+			],
+			"files": [
+				{
+					"path": "/var/lib/containers/hello.txt",
+					"contents": {
+						"source": "data:,hello%20world%0A"
+					}
+				},
+				{
+					"path": "/var/log/hello.txt",
+					"contents": {
+						"source": "data:,hello%20world%0A"
+					}
+				}
+			]
+		},
+		"systemd": {
+			"units": [
+				{
+					"name": "var-lib-containers.mount",
+					"enabled": true,
+					"contents": "[Mount]\nWhat=/dev/disk/by-partlabel/CONTR\nWhere=/var/lib/containers\nType=xfs\nOptions=defaults\n[Install]\nWantedBy=local-fs.target"
+				},
+				{
+					"name": "var-log.mount",
+					"enabled": true,
+					"contents": "[Mount]\nWhat=/dev/disk/by-partlabel/LOG\nWhere=/var/log\nType=xfs\nOptions=defaults\n[Install]\nWantedBy=local-fs.target"
+				}]
+		}
+	}`)
+)
+
+func init() {
+	// mount disks to `/var/log` and `/var/lib/containers`
+	register.Register(&register.Test{
+		Name:        "coreos.ignition.mount.disks",
+		Run:         testMountDisks,
+		ClusterSize: 0,
+		Platforms:   []string{"qemu"},
+		Distros:     []string{"fcos", "rhcos"},
+	})
+	// create new partiitons with disk `vda`
+	register.Register(&register.Test{
+		Name: "coreos.ignition.mount.partitions",
+		Run:  testMountPartitions,
+		UserData: conf.Ignition(`{
+			"ignition": {
+				"version": "2.2.0"
+			},
+			"storage": {
+				"disks": [
+					{
+						"device": "/dev/vda",
+						"wipeTable": false,
+						"partitions": [
+							{
+								"label": "CONTR",
+								"size": 2097152,
+								"start": 0
+							},
+							{
+								"label": "LOG",
+								"size": 2097152,
+								"start": 0
+							}
+						]
+					}
+				],
+				"filesystems": [
+					{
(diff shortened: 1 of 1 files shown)


kola/tests/ignition/mount.go
```diff
@@ -0,0 +1,331 @@
...
+							"size": 0,
+							"start": 0
+						}
+						]
+					}
+				],
+				"filesystems": [
+					{
+						"mount": {
+							"device": "/dev/vda5",
```
jlebon: You can do `/dev/disk/by-partlabel/CONTR` here for example to avoid hardcoding a specific partition number.

kola/tests/ignition/mount.go
```diff
@@ -0,0 +1,331 @@
...
+						}
+					}
+				]
+			},
+			"systemd": {
+				"units": [
+					{
+						"name": "var-lib-containers.mount",
+						"enabled": true,
+						"contents": "[Mount]\nWhat=/dev/vda5\nWhere=/var/lib/containers\nType=xfs\nOptions=defaults\n[Install]\nWantedBy=local-fs.target"
```
jlebon: And here, `.../by-label/CONTR`

kola/tests/ignition/mount.go
```diff
@@ -0,0 +1,331 @@
...
+
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform/conf"
+)
+
+func init() {
+	// verify mount to `/var/lib/containers`
+	register.Register(&register.Test{
+		Name:        "coreos.ignition.mount.var-lib-containers",
```
jlebon: Hmm, do we actually need to split this out into three separate tests instead of just the combined one?

kola/tests/ignition/mount.go
```diff
@@ -0,0 +1,331 @@
...
+				]
+			},
+			"systemd": {
+				"units": [
+					{
+						"name": "var-lib-containers.mount",
+						"enabled": true,
+						"contents": "[Mount]\nWhat=/dev/vda5\nWhere=/var/lib/containers\nType=xfs\nOptions=defaults\n[Install]\nWantedBy=local-fs.target"
+					}]
+			}
```
jlebon: Something else that'd be nice to test here is writing a file in each of those mounts and making sure that they were written when we boot.

kola/tests/ignition/mount.go
zonggen: :+1 on this

kola/tests/ignition/mount.go
zonggen: The Jira card stated it as three different tests so I thought we wanted three separate ones here, though I can't see any reasons not to just make a combined one.

kola/tests/ignition/mount.go
zonggen: Sounds good, will update :+1:

zonggen:
> Hmm, I'm a bit surprised you need to add space actually. With coreos/coreos-assembler#924, there should be a lot more headroom for more partitions. Can you confirm the images you tested had this patch?

You're right! Just did a rebuild and there was a lot of space left for more partitions. Here are the results:
FCOS with one extra partition:
```bash
[core@coreos ~] $ lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
sr0     11:0    1 1024M  0 rom  
vda    252:0    0    8G  0 disk 
|-vda1 252:1    0  384M  0 part /boot
|-vda2 252:2    0  127M  0 part /boot/efi
|-vda3 252:3    0    1M  0 part 
|-vda4 252:4    0  2.2G  0 part /sysroot
`-vda5 252:5    0  5.3G  0 part /var/log
```

FCOS with two extra partitions:
```
[core@coreos ~]$ lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
sr0     11:0    1 1024M  0 rom  
vda    252:0    0    8G  0 disk 
|-vda1 252:1    0  384M  0 part /boot
|-vda2 252:2    0  127M  0 part /boot/efi
|-vda3 252:3    0    1M  0 part 
|-vda4 252:4    0  2.2G  0 part /sysroot
|-vda5 252:5    0    1G  0 part /var/lib/containers
`-vda6 252:6    0    1G  0 part /var/log
```

The impression of having to add more space comes from the first few attempts to create partitions where the newly created partitions only has ~100k space and the system refused to mount the partition to `/var/lib/containers` since not enough space. Fixed by cleaning containers with `podman system prune -a`.

zonggen:
> (Another, perhaps more realistic approach, is having kola attach another disk, but I think having tests for both cases is good).

Or adding support to COSA to attach another disk

jlebon:
> (Another, perhaps more realistic approach, is having kola attach another disk, but I think having tests for both cases is good)

So reading the initial requirements for this work, it does seem like it was explicitly about setting up the partitions on a secondary disk. Looking at the kola test, I think you can do this via `platform.MachineOptions`. See e.g. `tests/misc/raid.go`:

```go
	options := platform.MachineOptions{
		AdditionalDisks: []platform.Disk{
			{Size: "520M", DeviceOpts: []string{"serial=secondary"}},
		},
	}
```

Up to you whether you want to work on that here, or as a follow-up PR!

kola/tests/ignition/mount.go
jlebon: Yeah, I think it'd be cleaner. WDYT about having two tests:

- one where we create the two partitions on the primary disk
- one where we attach two additional disks, one for each partition

?

zonggen:
I think it makes sense to add it here. Thanks for the pointer!
Updating

kola/tests/ignition/mount.go
zonggen: SGTM :+1:

kola/tests/ignition/mount.go
```diff
@@ -0,0 +1,411 @@
...
+	out = c.MustSSH(m, "cat /var/log/hello.txt")
+	if string(out) != "hello world" {
+		c.Fatalf("Failed to write content to /var/log/hello.txt")
+	}
+}
+
+func mountMultiple(c cluster.TestCluster) {
+	m := c.Machines()[0]
+
+	out := c.MustSSH(m, "cat /proc/mounts")
```
arithx: With the way that you've written these tests the validation could be split out into a sub-function to reduce duplication, e.x.:
```golang
func mountValidate(c cluster.TestCluster, m platform.Machine, mountContents, path string) {
    if strings.Contains(mountContents, path) {
        c.Fatalf("No partition mounted to %s", path)
    }

    fPath := filepath.Join(path, "hello.txt")
    fileContents := c.MustSSH(m, fPath)
    if string(fileContents) != "hello world" {
        c.Fatalf("Failed to write content to %s", fPath)
    }
}

func mountMultiple(c cluster.TestCluster) {
    m := c.Machines()[0]
    mountContents := c.MustSSH(m, "cat /proc/mounts")
    mountValidate(c, m, mountContents, "/var/lib/containers")
    mountValidate(c, m, mountContents, "/var/lib/log")
}
```

kola/tests/ignition/mount.go
```diff
@@ -0,0 +1,411 @@
...
+)
+
+func init() {
+	// verify mount to `/var/lib/containers`
+	register.Register(&register.Test{
+		Name:        "coreos.ignition.mount.var-lib-containers",
+		Run:         mountVarLibContainers,
+		ClusterSize: 1,
+		UserData: conf.Ignition(`{
+			"ignition": {"version": "2.2.0"},
```
cgwalters: Before we add many more tests that use nontrivial Ignition, we're really going to need to migrate the Ignition translation from cosa to here, and only write it one time.  @ajeddeloh may have newer code.

zonggen: :+1 on this, this would be very helpful
So for now maybe only test ignition v3 on FCOS?

kola/tests/ignition/mount.go
zonggen: Thanks for sharing the code! I've updated with helper functions based on you comment.

jlebon:
@zonggen Is this ready to go? Does this pass on both FCOS and RHCOS?

zonggen:
The tests are passing FCOS. Only FCOS(3.0.0) cases are tested, since as @cgwalters suggested we might be adding translation logic later. Should we add RHCOS(2.2.0) cases here as well?

kola/tests/ignition/mount.go
jlebon: @ajeddeloh Do you indeed have code ready to go we could use for this?

I think we can just hackily have v2 and v3 Ignition for now so we can get this test in and circle back to deduping once we have a solid translator in place. OTOH, if there's code ready to go to accomplish this, then let's definitely leverage that.

zonggen:
Another issue is that when testing with v2.2.0, adding ignition files with COSA would work but running inside kola test suite wouldn't. 

For example, with [code_here](https://gist.github.com/zonggen/1b0e860d337838a9dec0e8f0ed157d9f), running `cosa run -i partitions.ign` would work without issue but running `kola run -b rhcos --qemu-image rhcos-43.81.201912021843.0-qemu.x86_64.qcow2 coreos.ignition.mount.partitions` would hang.

jlebon:
> Another issue is that when testing with v2.2.0, adding ignition files with COSA would work but running inside kola test suite wouldn't.

You should be able to debug kola issues by looking at the console and/or journal output in `_kola_temp/...`.

zonggen:
I was originally following [instructions_here](https://github.com/coreos/mantle/tree/master/kola), where `kola run -b rhcos ...` was used, but it turns out that `--distro rhcos --ignition-version v2` should be used instead of only `-b rhcos` since by default kola would inject v3 config which is not valid under RHCOS, maybe make a PR against the documentation?

After that, [new_partitions.ign](https://gist.github.com/zonggen/6cf1353b827e5291a617763171a44920#file-partitions-ign) would still fail when mounting the device ([console.txt](https://gist.github.com/zonggen/6cf1353b827e5291a617763171a44920#file-console-txt)), which I'm currently stuck on. Extra help is much appreciated..

jlebon:
Ohhh wow, this is the `40rhcos-fips` module failing to run `ignition-files`... Sigh, I knew my sins would come back to haunt me. I think I was operating in Ignition v2 mode, where the `files` stage *doesn't* care about filesystems. But in v0.x, it of course does. Thanks for this, will fix!

jlebon:
OK, so I was playing with your [partitions.ign](https://gist.github.com/zonggen/6cf1353b827e5291a617763171a44920#file-partitions-ign) Ignition config here, and even with the `40rhcos-fips` issue fixed, there seems to be a race condition that happens between the encryption stuff and the disks stage. Anyway, sent a fix for that too. Once it's in, this should work (it does locally for me!).

zonggen:
The tests will work now for both RHCOS and FCOS, though I didn't hit the regression.. ready for another look whenever you have time! @jlebon

kola/tests/ignition/mount.go
```diff
@@ -0,0 +1,420 @@
...
+		m, err = pc.NewMachineWithOptions(ignConfig, options, true)
+	default:
+		c.Fatal("unknown cluster type")
+	}
+
+	if err != nil {
+		c.Fatal(err)
+	}
+
+	// reboot it to make sure it comes up again
```
jlebon: How about we check the partitions were mounted before rebooting too?

kola/tests/ignition/mount.go
```diff
@@ -0,0 +1,420 @@
...
+		c.Fatalf("could not reboot machine: %v", err)
+	}
+
+	// make sure partitions are mounted and files are written
+	mountContents := c.MustSSH(m, "cat /proc/mounts")
+	mountValidate(c, m, string(mountContents), "/var/lib/containers")
+	mountValidate(c, m, string(mountContents), "/var/log")
+}
+
+func testMountPartitions(c cluster.TestCluster) {
```
jlebon: I like the reboot check. Let's do that in this test too?

kola/tests/ignition/mount.go
```diff
@@ -0,0 +1,420 @@
...
+	m := c.Machines()[0]
+	// make sure partitions are mounted and files are written
+	mountContents := c.MustSSH(m, "cat /proc/mounts")
+	mountValidate(c, m, string(mountContents), "/var/lib/containers")
+	mountValidate(c, m, string(mountContents), "/var/log")
+}
+
+// Validate partition is mounted to `path` and `path`/hello.txt is written
+func mountValidate(c cluster.TestCluster, m platform.Machine, mountContents, path string) {
+	if !strings.Contains(mountContents, path) {
```
jlebon: I'd add a space to the `path` here so that e.g. `/var/log/systemd` doesn't match.

kola/tests/ignition/mount.go
```diff
@@ -0,0 +1,420 @@
...
+	mountValidate(c, m, string(mountContents), "/var/lib/containers")
+	mountValidate(c, m, string(mountContents), "/var/log")
+}
+
+func testMountPartitions(c cluster.TestCluster) {
+	m := c.Machines()[0]
+	// make sure partitions are mounted and files are written
+	mountContents := c.MustSSH(m, "cat /proc/mounts")
+	mountValidate(c, m, string(mountContents), "/var/lib/containers")
+	mountValidate(c, m, string(mountContents), "/var/log")
```
jlebon: Maybe let's just factor out this bit too into a separate function?

kola/tests/ignition/mount.go
zonggen: Updated.

kola/tests/ignition/mount.go
zonggen: Updated.

kola/tests/ignition/mount.go
zonggen: SGTM! Updated.

kola/tests/ignition/mount.go
zonggen: Updated with helper function.

zonggen:
Ran the tests on RHCOS and FCOS, passing.
Rebased and squash'd the commits

PR #1129, opened 2019-12-03
Add a --qemu-swtpm option to kola to support running kola tests without TPM

Prashanth684:
Architectures like s390x and pcc64le do not seem to have support for the virtual
TPM device in qemu. The kola tests fail to run today because it requires TPM support
by default.

With the --qemu-swtpm option specified, kola tests can be run on qemu images without TPM when setting it to false. For the tests that require TPM such as rhcos.luks.tpm, they can be blacklisted in the config. By default TPM is enabled for qemu.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -142,6 +142,7 @@ func init() {
 	sv(&kola.QEMUOptions.Board, "board", defaultTargetBoard, "target board")
 	sv(&kola.QEMUOptions.DiskImage, "qemu-image", "", "path to CoreOS disk image")
 	sv(&kola.QEMUOptions.BIOSImage, "qemu-bios", "", "BIOS to use for QEMU vm")
+	bv(&kola.QEMUOptions.Swtpm, "qemu-swtpm", true, "Create temporary software TPM")
 }
 
 // Sync up the command line options if there is dependency
--- a/platform/machine/unprivqemu/cluster.go
+++ b/platform/machine/unprivqemu/cluster.go
@@ -91,26 +91,29 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
 		consolePath: filepath.Join(dir, "console.txt"),
 	}
 
-	qm.swtpmTmpd, err = ioutil.TempDir("", "kola-swtpm")
-	if err != nil {
-		return nil, err
-	}
-
-	swtpmSock := filepath.Join(qm.swtpmTmpd, "swtpm-sock")
-
-	qm.swtpm = exec.Command("swtpm", "socket", "--tpm2",
-		"--ctrl", fmt.Sprintf("type=unixio,path=%s", swtpmSock),
-		"--terminate", "--tpmstate", fmt.Sprintf("dir=%s", qm.swtpmTmpd))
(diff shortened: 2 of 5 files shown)


platform/machine/unprivqemu/cluster.go
```diff
@@ -126,16 +129,19 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
 
 	// Default to user mode networking
 	qmCmd = append(qmCmd, "-netdev", "user,id=eth0,hostfwd=tcp:127.0.0.1:0-:22", "-device", platform.Virtio(qc.flight.opts.Board, "net", "netdev=eth0"))
+
 	// Bind the TPM device
-	qmCmd = append(qmCmd, "-chardev", fmt.Sprintf("socket,id=chrtpm,path=%s", swtpmSock), "-tpmdev", "emulator,id=tpm0,chardev=chrtpm", "-device", "tpm-tis,tpmdev=tpm0")
+	if qc.flight.opts.NoSwtpm == false {
+		qmCmd = append(qmCmd, "-chardev", fmt.Sprintf("socket,id=chrtpm,path=%s", swtpmSock), "-tpmdev", "emulator,id=tpm0,chardev=chrtpm", "-device", "tpm-tis,tpmdev=tpm0")
```
Prashanth684: wanted to pull socket creation here too..but when i did that i was getting an error like this:
`qemu-system-x86_64: -chardev socket,id=chrtpm,path=/tmp/kola-swtpm887381870/swtpm-sock: Failed to connect socket /tmp/kola-swtpm887381870/swtpm-sock: No such file or directory`

Not sure if it is a timing issue in creating the socket

platform/machine/unprivqemu/machine.go
```diff
@@ -69,11 +69,13 @@ func (m *machine) Reboot() error {
 }
 
 func (m *machine) Destroy() {
-	if err := m.swtpm.Kill(); err != nil {
-		plog.Errorf("Error killing swtpm of %v: %v", m.ID(), err)
-	}
-	if err := os.RemoveAll(m.swtpmTmpd); err != nil {
-		plog.Errorf("Error removing swtpm dir: %v", err)
+	if m.swtpmTmpd != "" {
```
Prashanth684: couldn't find a better way...

cmd/kola/options.go
```diff
@@ -142,6 +142,7 @@ func init() {
 	sv(&kola.QEMUOptions.Board, "board", defaultTargetBoard, "target board")
 	sv(&kola.QEMUOptions.DiskImage, "qemu-image", "", "path to CoreOS disk image")
 	sv(&kola.QEMUOptions.BIOSImage, "qemu-bios", "", "BIOS to use for QEMU vm")
+	bv(&kola.QEMUOptions.NoSwtpm, "no-swtpm", false, "Don't create a temporary software TPM")
```
arithx: Hmm, might be better to not make this a negative flag but rather something like `qemu-swtpm` defaulting to `true`; this would also have the added benefit of directly showing it's tied to the `qemu` related platforms.

Prashanth684: will do.

Prashanth684:
@arithx @cgwalters do the changes look good to merge? any concerns ?

PR #1130, opened 2019-12-06
qemu: Clean up argv handling a bit

cgwalters:
I'm working on creating a more direct `QemuInstance` struct
so we can use it for a `cosa run` backend separate from the flight
stuff.  This is prep cleanup for it - extract the basics to
a separate helper, and clarify the memory bits in particular.


cgwalters:
Formatted.

PR #1131, opened 2019-12-10
Add ExcludeArchitectures option for kola tests to exclude tests on certain arches

Prashanth684:
Based on the comments in https://github.com/coreos/coreos-assembler/pull/977#issuecomment-562677361, it makes sense to have an option to hardcode excluding certain tests which might never be supported for certain architectures. One good example is the rhcos.luks.tpm test which will probably never by supported on s390x. For powerpc , there is no concrete plans although there is talk of some effort soon.

### DIFF
--- a/cmd/kola/kola.go
+++ b/cmd/kola/kola.go
@@ -251,6 +251,7 @@ func runList(cmd *cobra.Command, args []string) {
 			test.Platforms,
 			test.ExcludePlatforms,
 			test.Architectures,
+			test.ExcludeArchitectures,
 			test.Distros,
 			test.ExcludeDistros}
 		item.updateValues()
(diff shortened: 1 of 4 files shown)


kola/tests/fips/fips.go
```diff
@@ -13,6 +13,9 @@ func init() {
 		Name:        `fips.enable`,
 		Flags:       []register.Flag{},
 		Distros:     []string{"rhcos"},
+		// Dracut relies on BOOT_IMAGE parameter to find the kernel and the hmac. ON s390x, BOOT_IMAGE is not a path,
+		// but a number indicating the chosen record. This is discussed in https://bugzilla.redhat.com/show_bug.cgi?id=1415032
```
jlebon: Hmm, this one looks fixable though. I.e. it's not inherent in the architecture, but just requires some dracut patches, right? Let's keep this as a blacklist instead (and file an issue to close that gap).

jlebon: I'm surprised this doesn't work. Doesn't RHEL support FIPS mode on s390x today?
Re. the path resolution thing, I added [a comment](https://bugzilla.redhat.com/show_bug.cgi?id=1415032#c22) on that RHBZ.

Prashanth684: ok...will do ..i put it in here because i was not sure about the timeline...but this can be fixed.

Prashanth684: the problem is two things: Firstly, the BOOT_IMAGE is not set on s390x on rhcos/fcos. I am not sure who sets it ? grub ? in that case it will not be set for s390x. Secondly, even if it is set - it is set to a number which represents the chosen record, so we have to translate that to the path.

PR #1133, opened 2019-12-10
Dockerfile: Simple Dockerfile

ashcrow:
Meant to be a base image for [coreos-assembler](http://github.com/coreos/coreos-assembler) usage.

```
$ podman run --rm localhost/mantle:latest kola version
mantle/kola version 0.9.0+527-gb729d0ba
$
```
cc @darkmuggle


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM registry.fedoraproject.org/fedora:31
+
+WORKDIR /tmp/mantle
+COPY . /tmp/mantle
+
+RUN yum update -y --nodocs && \
+    yum install -y --nodocs make golang gcc && \
```
cgwalters: Multi-stage builds are the better way to do this.  Also, we should consider using the pre-built OpenShift golang builder images.

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+    make && \
+    make install && \
```
cgwalters: If using multi-stage this would be `make install DESTDIR=/tmp/mantle` and then `COPY --from=builder /tmp/mantle/usr/bin/* /usr/bin` or so.

ashcrow: Good point, will update!

ashcrow:
> I'm missing some of the bigger picture here - this would be a base image?

cosa refactor -- it denotes that mantle would be used as a layer: https://url.corp.redhat.com/fabdf2c

_However_, another card in the epic notes using this container directly instead. Either way having a container seems like a need.

> Can you file e.g. a cosa issue so we hash out some design there? Or does it exist already?

It exists within Jira as an epic with cards, but not within GitHub to my knowledge.

cgwalters:
Seems OK to me, though I think we need a larger discussion about e.g. where this container is built (I would vote for the `coreos/` namespace in api.ci, and if we go that route it also makes it more obvious to use the shared OpenShift golang builder image, etc.) and how it's used (I recall now we discussed having it as a separate container for tests, which makes sense but definitely needs preparatory work in the pipelines).

cgwalters:
Actually, I'm not sure this is really going to work for the high level goal of making it faster to land testing because we won't have `qemu` in the image.

If we want to iterate on tests faster, I think a both easier and bigger win would be extracting the tests from the infrastructure to run the tests.  For example, have a `kola/` directory in the config repository with shell scripts to run, and teach `cosa kola` how to iterate over those and load them into kola.

Relatedly I want to make it easy for `kola` to execute container images with tests.

cgwalters:
That said this likely would work for *non* QEMU testing...but eh.  I think qemu is really really useful.

PR #1135, opened 2019-12-12
Remove priv qemu

cgwalters:
Our architecture with e.g. coreos-assembler is all around
fully unprivileged operation.  Having two variants of qemu
is a maintenance burden and impediment to making improvements
around our qemu testing.

Bigger picture, I think "privileged qemu" would really better
be done as "libvirt", since that would get us a lot of other
things for free.


cgwalters:
This builds on:

- https://github.com/coreos/mantle/pull/1134
- https://github.com/coreos/mantle/pull/1089

I'll rebase once those merge.

ajeddeloh:
I'm still of the opinion that we should not build mantle mantle to be only used in cosa. In that context I don't think we should remove priv qemu. With regard to using libvirt, I think there's an advantage to using raw qemu in that the testing environment isn't dependent on how libvirt does things; it's explicit what is used/needed. Using libvirt also makes kola much less self contained as it needs libvirtd running (I might be misunderstanding something there). I have less of a horse in this race though, so that's just my two cents.

cgwalters:
All dependencies are merged, updated :new:

cgwalters:
> I'm still of the opinion that we should not build mantle mantle to be only used in cosa. In that context I don't think we should remove priv qemu. 

Who else would use it, and why would they want priv qemu?

cgwalters:
There's always a tension between "general" and "specific".  Mantle is a lot of things, some of them could be reused by other projects.  But - there's also a *ton* of other projects around uploading cloud images and running tests.  e.g. Terraform is pretty popular for talking to clouds.  For testing...there's too many things to even try to enumerate.

At some point we need to go from "general" to "specific" - there are things fairly specific to Fedora CoreOS like systems that we need to write.  

Sharing things with other projects is a question hard to answer in general; we'd really need to have some other stakeholder here active who specifically wanted to use mantle to know what to do I think.

jlebon:
This is tough to balance, but I think I agree overall with not unnecessarily burdening ourselves with maintaining the two until there's a more concrete reason to (and of course, we can always bring things back if there's really a need to). Though would be interested in @lucab's or @dustymabe's opinion as well.

arithx:
> I'm still of the opinion that we should not build mantle mantle to be only used in cosa. In that context I don't think we should remove priv qemu. With regard to using libvirt, I think there's an advantage to using raw qemu in that the testing environment isn't dependent on how libvirt does things; it's explicit what is used/needed. Using libvirt also makes kola much less self contained as it needs libvirtd running (I might be misunderstanding something there). I have less of a horse in this race though, so that's just my two cents.

`qemu-unpriv` still directly shells out to `qemu-<arch>` binaries not libvirt.

PR #1137, opened 2019-12-13
kola/fips: Add user/group to file

cgwalters:
The OpenShift installer generates this, and we regressed on handling it.


lucab:
Cross-reference: https://github.com/openshift/installer/pull/2777#issuecomment-565286110

ashcrow:
LGTM

PR #1138, opened 2019-12-13
kola: Remove --qemu-bios argument

cgwalters:
What we really need here is to port the uefi/uefi-secure stuff
from `cosa run`.  In the meantime this is mostly CL cruft; delete
it in prep for a qemu refactoring.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -41,11 +41,6 @@ var (
 		"fcos":  "v3",
 		"rhcos": "v3",
 	}
-
-	kolaDefaultBIOS = map[string]string{
-		"amd64-usr": "bios-256k.bin",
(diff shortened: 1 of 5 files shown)


jlebon:
That one needs a rebase.

ajeddeloh:
Misc thought on this: treat bios/uefi/secure as machine types like c2.large

cgwalters:
> Misc thought on this: treat bios/uefi/secure as machine types like c2.large

Mmm...I think they're more bios types (distinct from this which took a bios filename which wasn't good enough for SB testing).

AWS machine types typically encompass a bag of other factors like disk size etc. that I think we always want independently frobbable.  (For example, I want to add `--disk-channel nvme` to here like cosa has it)

cgwalters:
Rebased :surfing_man:

PR #1139, opened 2019-12-13
kola: Also check for fatal dracut errors

cgwalters:
I had the fips test hang, and this might have helped.  What
I really want to do is change things so we *stream* (or poll)
the console and more quickly detect fatal errors.  But in
the meantime, let's add this error since it's definitely fatal.


jlebon:
That's neat, I didn't realize kola checked the console for error strings!

PR #1141, opened 2019-12-13
kola: Add http-server subcommand

cgwalters:
For a while `python3 -m http.server` has been my go-to for
a quick static webserver but it's pretty suboptimal since it's
single-threaded etc. and apparently OSTree's heavy use of
concurrent requests breaks it.

See
https://gist.github.com/cgwalters/4e5eb36bd439f5995928e9232d865f67

Let's add a convenience command here for cases like serving
up a local ostree repo and other things.

### DIFF
--- a/cmd/kola/kola.go
+++ b/cmd/kola/kola.go
@@ -17,6 +17,8 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"log"
+	"net/http"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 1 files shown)


cmd/kola/kola.go
```diff
@@ -61,12 +63,23 @@ will be ignored.
 		Run:   runList,
 	}
 
+	cmdHttpServer = &cobra.Command{
+		Use:   "http-server",
+		Short: "Run a static webserver",
+		Long: `Run a simple static webserver listening on port 8000
+
+This can be useful for e.g. serving locally build OSTree repos to qemu.
```
jlebon: Minor: `built`

cmd/kola/kola.go
```diff
@@ -281,6 +294,16 @@ func runList(cmd *cobra.Command, args []string) {
 	}
 }
 
+func runHttpServer(cmd *cobra.Command, args []string) {
+	directory := "."
+	port := 8000
```
jlebon: Optional: let's parametrize this one? Fine as a follow-up too.

cgwalters:
Updated :new:

jlebon:
Ahh, this needs a good ol' `gofmt`

PR #1142, opened 2019-12-13
kola/fips: Add a partition to Ignition config

jlebon:
Just to make the Ignition config a bit more realistic. Covers a recent
regression we had in that area:

https://github.com/coreos/mantle/pull/1128#issuecomment-564766733


jlebon:
Though requires the race condition fix mentioned in https://github.com/coreos/mantle/pull/1128#issuecomment-565519695.

jlebon:
OK yup, we can merge this now that the race is fixed!

PR #1144, opened 2019-12-16
kola: disable PodmanPing and add PodmanWgetHead in fcos.internet

jlebon:
It doesn't play well with `cosa kola` in OpenShift. We should enable it
only when we're not on qemu-unpriv. We do want the `PodmanEcho` test
though.

### DIFF
--- a/kola/tests/coretest/core.go
+++ b/kola/tests/coretest/core.go
@@ -54,14 +54,15 @@ func init() {
 		},
 		Distros: []string{"fcos", "rhcos"},
 	})
-	// TODO: enable DockerPing/DockerEcho once fixed
+	// TODO: Enable DockerPing/DockerEcho once fixed
+	// TODO: Only enable PodmanPing on non qemu-unpriv. Needs:
+	// https://github.com/coreos/mantle/issues/1132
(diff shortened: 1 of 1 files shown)


lucab:
LGTM.
I'd still like to have some basic container networking test. What about turning this into a `busybox wget --spider http://fedoraproject.org/static/hotspot.txt` instead of a ping? (as a followup)

jlebon:
> What about turning this into a `busybox wget --spider http://fedoraproject.org/static/hotspot.txt` instead of a ping? (as a followup)

Good idea, done!

jlebon:
Hmm, although... feels like it'd belong better in `tests/podman/podman.go` instead? Ehhh... I can do this if folks prefer.

lucab:
@jlebon test seems fine there, I guess the idea is that we test 1) host network via pulling 2) namespace network via wget. A `go fmt` pass is missing though.

lucab:
@jlebon new logic sounds good too. Feel feel to adjust title/comments and merge on your own.

jlebon:
Thanks for the review and suggestion!

Issue #1132, opened 2019-12-10
Support arch exclusion for Kolet tests


Prashanth684:
I encountered one problematic case, where for s390x, the `TestNetworkScripts` test of rhcos was failing on s390x because some scripts were present in /etc/sysconfig/network-scripts. On further discussions with s390x folks, it looks like the scripts are needed by the udev machinery to bring up s390x network devices.

With  this being the case, I am looking to see if we can add support for arch exclusion for the tests which are run through kolet and wondering what is the best way to go about doing this.

We can have an arch exclusion list for each test and evaluate the list before starting to execute the tests. Any suggestions/thoughts?

jlebon:
Fixed in https://github.com/coreos/mantle/pull/1131.

Prashanth684:
@jlebon i opened this for making the native tests arch specific. today the overall test itself can be exclude per arch with #1131 but i want that control for individual kolet tests too . For example:

```
NativeFuncs: map[string]func() error{
			"PortSSH":        TestPortSsh,
			"DbusPerms":      TestDbusPerms,
			"NetworkScripts": TestNetworkScripts,
			"ServicesActive": TestServicesActiveCoreOS,
			"ReadOnly":       TestReadOnlyFs,
			"Useradd":        TestUseradd,
			"MachineID":      TestMachineID,
		},
```
I want  to be able to disable `TestNetworkScripts` for s390x alone.

jlebon:
Ahh gotcha, thanks for clarifying. Re-opening!

PR #1145, opened 2019-12-16
Support excluding and blacklisting basic tests' subtest per arch

Prashanth684:
Support to exclude/blacklist native tests run by kolet. Following up on #1132, this is a way to exclude a certain native test instead of excluding the whole set of tests

Closes: #1132

### DIFF
--- a/kola/harness.go
+++ b/kola/harness.go
@@ -206,6 +206,25 @@ func filterTests(tests map[string]*register.Test, pattern, pltfrm string, versio
 				blacklisted = true
 				break
 			}
+
+			// Check if any native tests are blacklisted. To exclude native tests, specify the high level
+			// test and a "/" and then the glob pattern.
+			// - basic/TestNetworkScripts: excludes only TestNetworkScripts
+			// - basic/* - excludes all
+			// - If no pattern is specified after / , excludes none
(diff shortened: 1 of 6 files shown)


kola/register/register.go
```diff
@@ -40,14 +40,24 @@ var (
 	}
 )
 
+type NativeFuncWrap struct {
```
lucab: Nit: missing docstring.

kola/harness.go
```diff
@@ -206,6 +206,22 @@ func filterTests(tests map[string]*register.Test, pattern, pltfrm string, versio
 				blacklisted = true
 				break
 			}
+
+			// Check if any native tests are blacklisted. To exclude native tests, specify the basic
+			// test and a "/" and then the glob pattern. Example: rhcos.basic/TestNetworkScripts
```
lucab: What's the intended behavior if there is no glob pattern after the `/`? I believe right now this logic would panic.

Prashanth684: in case of something like `basic/`, it would run all the tests because `bl[nativeblacklistindex+1:]` would return an empty string. I will call that out explicitly in the docstring.

lucab:
Thanks for the PR, LGTM!

PR #1146, opened 2019-12-19
Rework qemu to use a better API

cgwalters:
In preparation for much more extensive use of qemu options for testing
(e.g. supporting uefi-secure, nmve for disks, etc.) rework the internal
QEMU bits to use a "builder+instance" pattern.

The `CreateQEMUCommand` API was growing a lot of arguments; a builder
struct helps that.

We can now more fully encapsulate qemu implementation details;
for example, `Virtio` is no longer exported.  The higher level
code needs to do much less construction of qemu command lines.

One twist with all of this is handling the Ignition config is
pretty different depending on whether the "board" supports `-fw_cfg`.
I reworked the way we handle disks to more clearly distinguish
the "primary" disk (where we may need to inject Ignition) versus
others.

The `QemuInstance` struct better encapsulates the qemu and swtpm
bits too.

### DIFF
--- a/platform/qemu.go
+++ b/platform/qemu.go
@@ -15,16 +15,27 @@
 package platform
 
 import (
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 
 	"github.com/coreos/mantle/system/exec"
 )
+
+// boardSupportsFwCfg if the target system supports injecting
+// Ignition via the qemu -fw_cfg option.
+func boardSupportsFwCfg(board string) bool {
+	switch board {
+	case "s390x-usr":
+	case "ppc64le-usr":
+		return false
+	}
+	return true
+}
 
 type MachineOptions struct {
 	AdditionalDisks []Disk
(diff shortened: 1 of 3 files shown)


cgwalters:
This currently depends on https://github.com/coreos/mantle/pull/1135 but it's not a *hard* dependency, I just didn't want to rewrite both qemu/qemuunpriv at once to use the new API as I was designing it.  But, if there's still objections to merging 1135 then I can rework this to support priv qemu too.

platform/qemu.go
```diff
@@ -34,80 +45,208 @@ type Disk struct {
 	Size        string   // disk image size in bytes, optional suffixes "K", "M", "G", "T" allowed. Incompatible with BackingFile
 	BackingFile string   // raw disk image to use. Incompatible with Size.
 	DeviceOpts  []string // extra options to pass to qemu. "serial=XXXX" makes disks show up as /dev/disk/by-id/virtio-<serial>
-	ConfPath    string   // path to ignition to be able to use it with guestfs for temporary qcow2 images
 }
 
-var (
-	ErrNeedSizeOrFile  = errors.New("Disks need either Size or BackingFile specified")
-	ErrBothSizeAndFile = errors.New("Only one of Size and BackingFile can be specified")
```
cgwalters: Also, there's no reason we can't support both a backing file *and* resizing (at least to a bigger size).

platform/qemu.go
```diff
@@ -15,17 +15,28 @@
...
 	"runtime"
 	"strings"
+	"syscall"
 
 	"github.com/coreos/mantle/system/exec"
 )
 
+// boardSupportsFwCfg if the target system supports injecting
+// Ignition via the qemu -fw_cfg option.
+func boardSupportsFwCfg(board string) bool {
```
arithx: Any reason to have this & `baseQemuArgs` be external functions rather than a class methods?

platform/qemu.go
```diff
@@ -34,80 +45,208 @@ type Disk struct {
...
+
+func (builder *QemuBuilder) EnableUsermodeNetworking(forwardedPort uint) {
+	var forward string
+	if forwardedPort != 0 {
+		forward = fmt.Sprintf(",hostfwd=tcp:127.0.0.1:0-:%d", forwardedPort)
+	}
+	builder.Append("-netdev", "user,id=eth0"+forward, "-device", virtio(builder.Board, "net", "netdev=eth0"))
+}
+
+func resolveBackingFile(backingFile string) (string, error) {
```
arithx: minor nit: I'd prefer to have the external function & class method definitions not be interleaved

cgwalters:
Rebased :surfing_man: now that #1135 merged!

platform/qemu.go
cgwalters: I think of methods as taking (in Rust terms) a `&mut self` - and really the same as in C++/Python/Java, where an instance method normally either reads or writes something from the struct.

But this code just operates on a string, not a builder instance.

platform/qemu.go
cgwalters: Hmm...what do you mean by "external"?  These aren't exported.  Is this more just like "order functions before methods"?

If so I think that generally makes sense...will do.

cgwalters: Hmm OK so I'll be honest, bouncing back and forth between shell/Golang/Rust/C/Python (not to mention Make, spec files etc.) so often makes it harder for me to internalize style.  I think my Golang, Rust and C have tended to intermix functions and methods.

What I think happens commonly is that a helper function gets extracted from a method.  And it's most natural to place that helper function just above the method.

Files often end up hence being ordered at least partially by their evolution over time.

From a quick survey of the mantle code and other Golang projects...it certainly doesn't seem like there's a consistent "functions before methods" - do you have a specific example or reference?  (Again, sorry if this is well known, I just don't *live* in Golang)

I think we could probably clean up the ordering here some though for sure...but note doing so would inflate the diff more.

platform/qemu.go
cgwalters: Reworked to be a method.

platform/qemu.go
cgwalters: OK, I added another commit which tries to clean things up here some more.

cgwalters: I think the guestfish injection stuff should probably be split to a separate file, but...trying to be incremental.

arithx: > I think the guestfish injection stuff should probably be split to a separate file, but...trying to be incremental.

That's fine!

PR #1147, opened 2020-01-08
qemuexec: New command

cgwalters:
This runs qemu directly as a subprocess, and supports providing
additional arguments to qemu.

Prep for rebasing `cosa run` on this, and draining more of the
qemu logic into mantle so we can share it between mantle/cosa.

### DIFF
--- a/cmd/kola/qemuexec.go
+++ b/cmd/kola/qemuexec.go
@@ -1,0 +1,84 @@
+// Copyright 2019 Red Hat, Inc.
+//
+// Run qemu directly as a subprocess.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/coreos/mantle/kola"
+	"github.com/coreos/mantle/platform"
+)
+
+var (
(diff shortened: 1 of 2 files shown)


cgwalters:
Mmm, let's hold this until I've actually tried rebasing `cosa run`.  Looking at that quickly we need to pass through the Ignition at least.

cgwalters:
WIP in https://github.com/coreos/coreos-assembler/pull/1027

cgwalters:
OK lifting WIP on this, tested with 

```
diff --git a/src/cmd-test-coreos-installer b/src/cmd-test-coreos-installer
index 705b8384..59f2892d 100644
--- a/src/cmd-test-coreos-installer
+++ b/src/cmd-test-coreos-installer
@@ -84,8 +84,7 @@ fi
 
 completionf=${statedir}/completion.txt
 touch "${completionf}"
-setpriv --pdeathsig SIGTERM -- qemu-system-x86_64 -accel kvm -m 1536 -nographic \
-       -object rng-random,filename=/dev/urandom,id=rng0 -device virtio-rng-pci,rng=rng0 \
+kola qemuexec -m 1524 -- \
        -boot once=n -option-rom /usr/share/qemu/pxe-rtl8139.rom \
        -device e1000,netdev=mynet0,mac=52:54:00:12:34:56 -netdev user,id=mynet0,net=192.168.76.0/24,dhcpstart=192.168.76.9,tftp=${tftpdir},bootfile=/pxelinu0\
        -drive if=virtio,file=${disk} \
```

cmd/kola/qemuexec.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := doQemuExec(cmd, args); err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		os.Exit(1)
+	}
+}
+
+func doQemuExec(cmd *cobra.Command, args []string) error {
+	var err error
+
+	builder := platform.NewBuilder(sdk.DefaultBoard(), "")
```
arithx: Probably worthwhile to move the board out to a parameter.

cgwalters: We already have a CLI option for that of course from kola; changed to use it.  All fixed another bug - I'd done a rebase but hadn't tested the rebase.

PR #1150, opened 2020-01-08
qemu: Add --qemu-firmware bios|uefi|uefi-secure

cgwalters:
More moving of code into mantle that was in coreos-assembler.
We want to test UEFI and Secure Boot for obvious reasons.

### DIFF
--- a/platform/machine/unprivqemu/cluster.go
+++ b/platform/machine/unprivqemu/cluster.go
@@ -94,6 +94,7 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
 	builder := platform.NewBuilder(board, confPath)
 	defer builder.Close()
 	builder.Uuid = qm.id
+	builder.Firmware = qc.flight.opts.Firmware
 	builder.ConsoleToFile(qm.consolePath)
 
 	primaryDisk := platform.Disk{
--- a/platform/qemu.go
+++ b/platform/qemu.go
@@ -16,6 +16,7 @@ package platform
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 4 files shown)


cgwalters:
See https://github.com/coreos/coreos-assembler/pull/570

platform/qemu.go
```diff
@@ -403,6 +406,35 @@ func baseQemuArgs(board string) []string {
 	}
 }
 
+func (builder *QemuBuilder) setupUefi(secureBoot bool) error {
+	varsVariant := ""
+	if secureBoot {
+		varsVariant = ".secboot"
+	}
+	varsSrc, err := os.Open(fmt.Sprintf("/usr/share/edk2/ovmf/OVMF_VARS%s.fd", varsVariant))
```
arithx: I'm not too familiar with this, is there a potential issue with having multiple spawned machines all touching this file?

cgwalters: We're making a private writable copy; this line is just opening it read-onyl.

cgwalters:
My test case for this was:

```
$ cosa kola run -- --qemu-firmware uefi-secure
$ grep lockdown tmp/kola/basic/8130194a-e7d8-4cde-a8a6-3c96153b4b66/console.txt
[    0.000000] Kernel is locked down from EFI Secure Boot mode; see man kernel_lockdown.7
...
```

platform/machine/unprivqemu/cluster.go
```diff
@@ -94,6 +94,7 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
 	builder := platform.NewBuilder(board, confPath)
 	defer builder.Close()
 	builder.Uuid = qm.id
+	builder.Firmware = qc.flight.opts.Firmware
```
arithx: Unrelated to this PR but are we missing the setting of `builder.Swtpm`? It's being defaulted to `true` in `platform.NewBuilder` and a cursory scan didn't see it being set to the value of `qc.flight.opts.Swtpm` anywhere.

cgwalters: I think we should nuke that option and just set it off if !x86_64 - will do it in a followup PR.

PR #1151, opened 2020-01-08
plume: drop CL-specific release code and deprecate --channel in favour of --stream

jlebon:
See individual commit messages.


jlebon:
I'm thinking of changing the interface so that the "distro" would instead be a positional arg, e.g.: `plume release <distro> [options]`. That way we don't have to expose all the flags for all the distros.

PR #1154, opened 2020-01-09
qemu: Automatically turn off swtpm on !x86_64 here

cgwalters:
I think we should have done this instead of
https://github.com/coreos/coreos-assembler/pull/973

This way invoking `kola` rather than `cosa kola` does the right thing.
Part of de-duping cosa and mantle.

### DIFF
--- a/platform/qemu.go
+++ b/platform/qemu.go
@@ -545,7 +545,8 @@ func (builder *QemuBuilder) Exec() (*QemuInstance, error) {
 		panic("Ignition specified but no primary disk")
 	}
 
-	if builder.Swtpm {
+	// TPM devices aren't on s390x/ppc64le at least
+	if builder.Swtpm && builder.Board == "amd64-usr" {
 		inst.swtpmTmpd, err = ioutil.TempDir("", "kola-swtpm")
 		if err != nil {
 			return nil, err
--- a/platform/machine/unprivqemu/cluster.go
+++ b/platform/machine/unprivqemu/cluster.go
@@ -91,6 +91,7 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
 	defer builder.Close()
 	builder.Uuid = qm.id
 	builder.Firmware = qc.flight.opts.Firmware
+	builder.Swtpm = qc.flight.opts.Swtpm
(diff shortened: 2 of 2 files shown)


platform/qemu.go
```diff
@@ -545,7 +545,8 @@ func (builder *QemuBuilder) Exec() (*QemuInstance, error) {
 		panic("Ignition specified but no primary disk")
 	}
 
-	if builder.Swtpm {
+	// TPM devices aren't on s390x/ppc64le at least
+	if builder.Swtpm && builder.Board == "amd64-usr" {
```
arithx: Should we also plumb through the setting of `builder.Swtpm` to the value of `qc.flight.opts.Swtpm` in [the cluster code](https://github.com/coreos/mantle/blob/master/platform/machine/unprivqemu/cluster.go#L93) to allow for not using swtpm on x86_64 machines?

Prashanth684: This looks fine. I am not sure if TPM is supported on ARM. But for ppc64le and s390x looks like it will not be supported anytime soon.

Prashanth684: i think the Swtpm option can be removed now that we have this

cgwalters: Right sorry, I hadn't forgotten your earlier comment and I meant to fix that too with this.  Done now!

cgwalters: Based on... https://www.kernel.org/doc/html/latest/security/tpm/tpm_ftpm_tee.html maybe?  But I think the swtpm we use is probably emulating an Intel one?  Not sure.

cgwalters: Anyways let's not block on this right now but I'll ask about it.

Prashanth684: agreed.

Prashanth684:
lgtm

platform/qemu.go
nullr0ute: I think software TPM2 should work just fine on ARM, if it doesn't let me know and I'll get it looked into. There's been a lot of work done with TPM2 on ARM so I suspect it should just work like on x86.

cgwalters: https://github.com/coreos/mantle/pull/1158

Issue #1157, opened 2020-01-10
kola --parallel broken with qemu


cgwalters:
In my test FCOS pipeline, one of the nodes ran OOM.  Looking at it, we're spawning too many qemu instances.  If I do `cosa kola run -- --parallel 2` locally, I get 8 qemu proceses.

Yet, I only see 1 concurrent if I don't specify `--parallel`.  So the bug seems to occur when parallel > 1?

I started reading the harness code around parallelism but my eyes glazed over - so many goroutines and barriers and mutexes and lions and tigers and bears...

cgwalters:
I don't think this is actually specific to qemu of course.  Probably we just don't notice it as much.

Hum...could this have something to do with subtests?

cgwalters:
Ah I see, this code came from https://github.com/coreos/mantle/commit/d5f50b5ab0886b1953359e1b3ef816d49132fe94 - it forked from the golang upstream, so of course they're going to be all "channels solve everything!".

Ah, and https://github.com/golang/go/commit/f04d5836181dec3ec1b7e427607f02fa7a204a2d#diff-70e628298261565d825f7199d13042f2 looks relevant here.

And it's Go so of course there's a pile of race conditions with unstructured goroutines, like
- https://github.com/golang/go/commit/7f31971f594edbacbdba5407aaee042850fbd220#diff-70e628298261565d825f7199d13042f2
- https://github.com/golang/go/commit/eff1e68528fc9052a7ff1ac7afe222696f85db8c#diff-70e628298261565d825f7199d13042f2

Blah.

arithx:
> If I do cosa kola run -- --parallel 2 locally, I get 8 qemu proceses.

Notably the `parallel` flag concerns test count not machine count so theoretically you could have a much larger machine count than the value of the parallel flag if the tests used require multiple machines (though in practice I don't believe we have a test with more than 3 machines, and it might be less now after dropping legacy CL tests).

> I started reading the harness code around parallelism but my eyes glazed over - so many goroutines and barriers and mutexes and lions and tigers and bears...

Yeah it's not pretty unfortunately, [the runner readme](https://github.com/coreos/mantle/blob/master/runner-readme.md) might help a bit.

cgwalters:
So...hum.  How about `--qemu-parallel N` which ensures that at most N qemu Flight instances are created at a time?

arithx:
> So...hum. How about --max-flights N which ensures that at most N Flight instances are created at a time?

[...]

PR #1161, opened 2020-01-14
kola: Add --max-machines

cgwalters:
This is only implemented for qemu at the moment, though it'd
be a mostly mechanical change to propagate it to the other
providers.

For our pipeline testing, we need to have a hard cap on the number
of qemu instances we spawn, otherwise we can go over the RAM
allocated to the pod.

Actually the FCOS pipeline today doesn't impose a hard cap, and
my test pipeline in the coreosci (nested GCP virt) ended up bringing
down the node via the OOM killer.

There were a few bugs here; first we were leaking the spawned
qemu instance.  We also need to invoke `Wait()` synchronously in
destruction.

Then, add a dependency on the `golang/x/semaphore` library, and
use it to implement a max limit.

Closes: https://github.com/coreos/mantle/issues/1157

### DIFF
--- a/platform/machine/unprivqemu/cluster.go
+++ b/platform/machine/unprivqemu/cluster.go
@@ -110,6 +110,13 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
 	}
 	builder.EnableUsermodeNetworking(22)
 
+	qc.flight.AcquireMachineReservation()
+	destroyMachine := true
+	defer func() {
+		if destroyMachine {
+			qm.Destroy()
+		}
+	}()
 	inst, err := builder.Exec()
 	if err != nil {
 		return nil, err
@@ -129,10 +136,10 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
 	}
 
 	if err := platform.StartMachine(qm, qm.journal); err != nil {
-		qm.Destroy()
 		return nil, err
 	}
 
+	destroyMachine = false
 	qc.AddMach(qm)
 
 	return qm, nil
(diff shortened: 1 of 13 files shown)


cgwalters:
Tweaked this to default `--max-machines` to the value of `--parallel`.

platform/machine/unprivqemu/cluster.go
```diff
@@ -110,10 +110,18 @@ func (qc *Cluster) NewMachineWithOptions(userdata *conf.UserData, options platfo
...
+	defer func() {
+		if destroyMachine {
+			qm.Destroy()
+		}
+	}()
 	inst, err := builder.Exec()
 	if err != nil {
 		return nil, err
 	}
+	qm.inst = *inst
```
cgwalters: This was the core leak.

arithx:
A couple questions:

What happens if:
 - The max-machine count is less than the required amount of machines for a given test (e.x. the test needs 3 machines and max-machines is 2
 - We're running tests serially with a max-machines count of 1 and we hit a test like [the NFS test](https://github.com/coreos/mantle/blob/master/kola/tests/misc/nfs.go) that spawns machines inside of the test.

jlebon:
Nice, good catch!

> What happens if:

Probably we should automatically skip tests that have a required machine count > `--max-machines`? And then have tests like the NFS one actually declare how many machines it intends to use. Seems good practice anyway for e.g. estimating resource needs when allocating the pod that will run kola.

I think we could at least get in the leaks and fixes for now, right? That alone should greatly reduce memory stress on the nodes we're allocated on. WDYT about splitting those out as a separate PR?

cgwalters:
> Probably we should automatically skip tests that have a required machine count > --max-machines? And then have tests like the NFS one actually declare how many machines it intends to use. 

Yeah, something like that.

> WDYT about splitting those out as a separate PR?

OK done in

https://github.com/coreos/mantle/pull/1163

But, I think we'll need this PR in order to make the pipeline truly reliable in the face of hard memory caps.

ashcrow:
> > Probably we should automatically skip tests that have a required machine count > --max-machines? And then have tests like the NFS one actually declare how many machines it intends to use.
> 
> Yeah, something like that.

Added a few reviewers, though it sounds like the above is needed so we avoid timeouts before this is ready. Is it fair to mark this WIP for now?

ashcrow:
And I think this might might sense to be redirected to the cosa repo.

PR #1162, opened 2020-01-14
ore/aws/copy-image: stream results back to stdout

jlebon:
Rather than accumulate all the results and then dump it all on stdout,
stream the output as line-separated JSON objects as the goroutines
complete.

This is technically a compatibility break since we change the format of
the output. However, I'm not aware of any clients which use `copy-image`
directly and not through cosa.

Prep for making use of this in cosa to speed up `cosa aws-replicate`.

### DIFF
--- a/platform/api/aws/images.go
+++ b/platform/api/aws/images.go
@@ -524,7 +524,7 @@ func (a *API) GrantLaunchPermission(imageID string, userIDs []string) error {
 	return nil
 }
 
-func (a *API) CopyImage(sourceImageID string, regions []string) (map[string]ImageData, error) {
+func (a *API) CopyImage(sourceImageID string, regions []string, cb func(string, ImageData)) error {
 	type result struct {
 		region string
 		data   ImageData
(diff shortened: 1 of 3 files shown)


jlebon:
Hmm, it seems like it's used by `fedora` too? Anyway, I just adapted it above! :arrow_up:

platform/api/aws/images.go
```diff
@@ -594,19 +594,21 @@ func (a *API) CopyImage(sourceImageID string, regions []string) (map[string]Imag
 			ch <- res
 		}()
 	}
-	wg.Wait()
-	close(ch)
+	go func() {
+		wg.Wait()
```
arithx: With the `wg.Wait` inside of a goroutine doesn't the other logic need to change to accommodate the fact that it's potentially being ran before the goroutine is done processing?

jlebon: Not sure which bit you're referring to as "the other logic"... I'm guessing you meant the `range ch` below? That will keep looping until we close the channel, right? And that'll only happen once we've waited for all the goroutines.

PR #1164, opened 2020-01-15
qemu: Silence errors from swtpm by default

cgwalters:
This will make failures harder to debug in the future, but
avoids spew on the normal case when qemu gets killed first today.


platform/qemu.go
```diff
@@ -572,7 +572,8 @@ func (builder *QemuBuilder) Exec() (*QemuInstance, error) {
 			"--ctrl", fmt.Sprintf("type=unixio,path=%s", swtpmSock),
 			"--terminate", "--tpmstate", fmt.Sprintf("dir=%s", inst.swtpmTmpd))
 		cmd := inst.swtpm.(*exec.ExecCmd)
-		cmd.Stderr = os.Stderr
+		// For now silence the swtpm stderr as it prints errors when
+		// disconnected, but that's normal.
```
jlebon: Could we toggle it based on `--verbose`? That way, we can still easily inspect it if something keeps failing.

cgwalters: Looked at that but the options aren't plumbed through, and that `--verbose` already defaults to true (oddly).  I don't think it's worth the time on this.

PR #1165, opened 2020-01-15
kola: Add a --cosa-build argument

cgwalters:
This starts the ball rolling on having mantle understand cosa,
which is key for long-term dedup of the projects.

The main motivation for this right now is that by decoding the
JSON into a Go struct, we are effectively doing a schema validation.

After this lands I'll make a separate PR to flesh out the struct
entirely, then change the decoder to error on unknown keys.

This also removes some CL cruft around the default disk image.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -15,27 +15,26 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/coreos/mantle/auth"
+	"github.com/coreos/mantle/cosa"
 	"github.com/coreos/mantle/kola"
 	"github.com/coreos/mantle/platform"
 	"github.com/coreos/mantle/sdk"
 )
 
 var (
-	outputDir          string
-	kolaPlatform       string
-	defaultTargetBoard = sdk.DefaultBoard()
-	kolaArchitectures  = []string{"amd64"}
-	kolaPlatforms      = []string{"aws", "azure", "do", "esx", "gce", "openstack", "packet", "qemu", "qemu-unpriv"}
-	kolaDistros        = []string{"fcos", "rhcos"}
-	kolaDefaultImages  = map[string]string{
-		"amd64-usr": sdk.BuildRoot() + "/images/amd64-usr/latest/coreos_production_image.bin",
-		"arm64-usr": sdk.BuildRoot() + "/images/arm64-usr/latest/coreos_production_image.bin",
-	}
+	outputDir                   string
+	kolaPlatform                string
+	defaultTargetBoard          = sdk.DefaultBoard()
+	kolaArchitectures           = []string{"amd64"}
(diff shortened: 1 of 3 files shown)


cgwalters:
And once we change `cosa kola` to use this by default in combination with making the decoding exhaustive, it would have blocked the unintentional schema regressions introduced recently.

jlebon:
Neat, I think this is mostly aligned with what I'm working on right now, which is: adding a `kola cosa-run` command which will run more expensive multi-artifact tests, such as installer tests and upgrade tests. (It would be workdir-aware though, rather than just `meta.json`-aware).

cmd/kola/options.go
```diff
@@ -62,6 +61,7 @@ func init() {
 	sv(&kola.Options.IgnitionVersion, "ignition-version", "", "Ignition version override: v2, v3")
 	ssv(&kola.BlacklistedTests, "blacklist-test", []string{}, "List of tests to blacklist")
 	bv(&kola.Options.SSHOnTestFailure, "ssh-on-test-failure", false, "SSH into a machine when tests fail")
+	sv(&kola.Options.CosaBuild, "cosa-build", "", "File path for coreos-assembler meta.json")
```
arithx: Would there be any advantage to having the parameters be the top level COSA dir & the build-id over a path to the meta.json for the build?

cosa/build.go
```diff
@@ -0,0 +1,46 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cosa
+
+// This code was copied from openshift-installer's
+// https://github.com/openshift/installer/blob/a0350404997b0493d7bb16aa2875e5c42879b069/pkg/rhcos/builds.go
+// For now, copy-paste updates there.  Later, maybe consider making this a public API?
```
arithx: I generally prefer to have the structs individually defined but this is fine for now given the context.

cmd/kola/options.go
cgwalters: That'd be better, it'd just involve moving a good bit more code, including parsing `builds.json` etc.

cgwalters:
OK, fixed the gofmt error.

cmd/kola/options.go
```diff
@@ -171,13 +171,29 @@ func syncOptions() error {
...
+		}
+		cosaBuild = &tmpBuild
 	}
 
 	if kola.QEMUOptions.DiskImage == "" {
-		kola.QEMUOptions.DiskImage = image
+		if cosaBuild != nil {
+			kola.QEMUOptions.DiskImage = filepath.Join(filepath.Dir(kola.Options.CosaBuild), cosaBuild.Images.QEMU.Path)
+		} else {
+			panic("No qemu disk image provided")
```
arithx: Rather than a `panic` you can just return an error and it will still directly exit (https://github.com/coreos/mantle/blob/master/cmd/kola/kola.go#L95).

cgwalters: Thanks, fixed!

PR #1171, opened 2020-01-28
ore/gcloud: Add --write-url argument

cgwalters:
coreos-assembler was trying to re-create the URL that ore computed
for the storage, and a recent refactoring broke that.

Add an option to write out the URL we used so we only need
to have its computation in one place.

Specifically, cosa started replacing `.` with `-` but ore wasn't.

xref https://github.com/coreos/coreos-assembler/issues/1078

### DIFF
--- a/cmd/ore/gcloud/upload.go
+++ b/cmd/ore/gcloud/upload.go
@@ -16,6 +16,7 @@ package gcloud
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


cgwalters:
Maybe we should try to consistently have `ore upload` output JSON or something at least, if we're not going to have mantle directly change `meta.json`.

PR #1175, opened 2020-02-07
kola/tests/rpmostree: relax regex on CoreOS version strings

miabbott:
Since ostree version strings can be completely arbitrary (even whitespace and emoji!), let's just relax the regex to be more accomodating.

### DIFF
--- a/kola/tests/rpmostree/status.go
+++ b/kola/tests/rpmostree/status.go
@@ -35,8 +35,8 @@ func init() {
 }
 
 var (
-	// Regex to extract version number from "rpm-ostree status"
-	rpmOstreeVersionRegex string = `^Version: ([0-9a-zA-Z.]+).*`
+	// rpmOstreeVersionRegex parses the text output of "rpm-ostree status"
+	rpmOstreeVersionRegex string = `^Version: *([^ ]*)`
 )
 
 // rpmOstreeCleanup calls 'rpm-ostree cleanup -rpmb' on a host and verifies
@@ -91,9 +91,9 @@ func rpmOstreeStatus(c cluster.TestCluster) {
 	for _, line := range statusArray {
 		versionMatch := reVersion.FindStringSubmatch(strings.Trim(line, " "))
 		if versionMatch != nil {
-			// versionMatch should be like `[Version: 420.8.20190711.0 (2019-07-11T09:00:04Z) 420.8.20190711.0]`
-			// i.e. the full match and the group we want
-			// `versionMatch[len(versionMatch)-1]` gets the last element in the array
+			// versionMatch should be like `Version: 44.81.202002052139-0 (2020-02-05T21:45:46Z) 44.81.202002052139-0`
+			// i.e. the full match and the groups we want
+			// `versionMatch[len(versionMatch)-1]` gets the last element (full version string) in the array
 			rpmOstreeVersion = versionMatch[len(versionMatch)-1]
 		}
 	}


kola/tests/rpmostree/status.go
```diff
@@ -36,7 +36,7 @@ func init() {
 
 var (
 	// Regex to extract version number from "rpm-ostree status"
-	rpmOstreeVersionRegex string = `^Version: ([0-9a-zA-Z.]+).*`
+	rpmOstreeVersionRegex string = `^Version: ([0-9a-zA-Z.]+(-[0-9])?).*`
```
cgwalters: Wouldn't it be simpler to just include `-` in the set of characters?  

Or in fact, allow any characters except whitespace?  OSTree version numbers can be arbitrary strings, I don't see why we should try to artificially constrain it here.  I mean...we *might* decide to use e.g. emoji someday :wink: 

This change lets the test pass for me:
```
diff --git a/kola/tests/rpmostree/status.go b/kola/tests/rpmostree/status.go
index 9c36ce7d..c521d831 100644
--- a/kola/tests/rpmostree/status.go
+++ b/kola/tests/rpmostree/status.go
@@ -35,8 +35,8 @@ func init() {
 }
 
 var (
-       // Regex to extract version number from "rpm-ostree status"
-       rpmOstreeVersionRegex string = `^Version: ([0-9a-zA-Z.]+).*`
+       // rpmOstreeVersionRegex parses the text output of "rpm-ostree status"
+       rpmOstreeVersionRegex string = `^Version: *([^ ]*)`
 )
 
 // rpmOstreeCleanup calls 'rpm-ostree cleanup -rpmb' on a host and verifies
```

cgwalters: Technically, OSTree version numbers could even include whitespace but even I'm not that evil...

cgwalters: The other reason to do it this way is that the test will continue to pass on older RHCOS versions too.

miabbott: > Or in fact, allow any characters except whitespace? OSTree version numbers can be arbitrary strings, I don't see why we should try to artificially constrain it here. I mean...we _might_ decide to use e.g. emoji someday wink

Ack, that's the easier, more sane approach.

One day we should switch to [emojisum](https://github.com/emojisum/emojisum/)  :smile:

miabbott:
Works against FCOS 31.20200118.3.0 and RHCOS 44.81.202002052139-0

kola/tests/rpmostree/status.go
jlebon: What is that subtest actually trying to test though? ISTM like it's about sanity-checking that the versioning fields in the treefile are set correctly by rpm-ostree? And in FCOS it'd be that versionary spit out a version that corresponds to the expected scheme.

If we make it a liberal regex, it's not really testing anything at all, other than just making sure there *is* a version, which is a pretty low bar. :)

cgwalters: TBH I dunno what the original intent was, but I think if we mess up our version numbers that'll be super obvious and not something we need a kola test for (booting a whole VM for it would be silly).

miabbott: > What is that subtest actually trying to test though? ISTM like it's about sanity-checking that the versioning fields in the treefile are set correctly by rpm-ostree? And in FCOS it'd be that versionary spit out a version that corresponds to the expected scheme.

The original intent goes back to the first `rpm-ostree` tests from AHT days...just making sure that stdout + JSON output are identical.

The criticisms about the test as it is now (and as it was then) are valid; we could probably safely drop this test entirely.

PR #1177, opened 2020-02-07
qemuexec: Add --ignition arg, propagate kola --qemu-* args

cgwalters:
Plumb through the rest of the bits the CLI needs to support
replacing the guts of `cosa run`.

### DIFF
--- a/cmd/kola/qemuexec.go
+++ b/cmd/kola/qemuexec.go
@@ -36,12 +36,15 @@ var (
 
 	memory  int
 	usernet bool
+
+	ignition string
 )
 
 func init() {
 	root.AddCommand(cmdQemuExec)
 	cmdQemuExec.Flags().BoolVarP(&usernet, "usernet", "U", false, "Enable usermode networking")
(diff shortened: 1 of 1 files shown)


cmd/kola/qemuexec.go
```diff
@@ -36,12 +36,15 @@ var (
...
 	usernet bool
+
+	ignition string
 )
 
 func init() {
 	root.AddCommand(cmdQemuExec)
 	cmdQemuExec.Flags().BoolVarP(&usernet, "usernet", "U", false, "Enable usermode networking")
 	cmdQemuExec.Flags().IntVarP(&memory, "memory", "m", 0, "Memory in MB")
+	cmdQemuExec.Flags().StringVarP(&ignition, "ignition", "i", "", "Path to ignition config")
```
arithx: Notably spawn uses userdata (partially because CL had to support cloudinit as well; https://github.com/coreos/mantle/blob/master/cmd/kola/spawn.go#L60)

PR #1180, opened 2020-02-11
platform/api/aws: Iterate over all subnets to run the instances

pothos:
The subnet determines the Availability Zone. If in that
zone no machines are available, running an instance
will fail. The other Availability Zones probably have
machines available and the code should try to run
instances there instead of giving up too early.

### DIFF
--- a/platform/api/aws/ec2.go
+++ b/platform/api/aws/ec2.go
@@ -106,57 +106,66 @@ func (a *API) CreateInstances(name, keyname, userdata string, count uint64) ([]*
 		return nil, fmt.Errorf("error resolving vpc: %v", err)
 	}
 
-	subnetId, err := a.getSubnetID(vpcId)
+	subnetIds, err := a.getSubnetIDs(vpcId)
 	if err != nil {
-		return nil, fmt.Errorf("error resolving subnet: %v", err)
+		return nil, fmt.Errorf("error resolving subnets: %v", err)
 	}
 
 	key := &keyname
 	if keyname == "" {
 		key = nil
 	}
-	inst := ec2.RunInstancesInput{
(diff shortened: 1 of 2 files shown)


cgwalters:
Seems rasonable offhand, though have thought about whether we're starting to grow too much "provisioning" logic there that may be better off farmed out to something else (terraform/ansible/whatever).  But clearly this is an incremental enough change to merge.

One procedural note - so far we've been removing CL support from git master, focusing on {F,RH}COS.  There's a [cl branch](https://github.com/coreos/mantle/tree/cl) that it may make sense for you to use/maintain?

If this continues we should likely have a discussion about how to share code going forward.

pothos:
Thanks for the notice, we'll have a look at the branch and we also have to think how we could keep in sync with your `master` in the future given all the change that will come.

cgwalters:
We never merged this, even though it was approved - this seems like a mistake.

But https://github.com/coreos/coreos-assembler/pull/1152 merged and I notice you have a mantle fork over at https://github.com/kinvolk/mantle - so things are going to diverge more.  The current plan is to archive this repository.

I'd like to keep in mind the possibility of continuing to share some code, but things like Ignition spec 2 vs 3, our increasing tie of "CoreOS" to mean (Ignition, OSTree) etc. will make that tricky.

bgilbert:
Note that we won't be archiving the repository until Container Linux goes EOL.  The `master` branch is no longer in use, but the `cl` branch will be maintained until May 26.

pothos:
Thanks again for writing. Our fork is located at https://github.com/flatcar-linux/mantle/ and we will give it a bit more care soon.

Issue #1179, opened 2020-02-10
kola run-ext: Add support for running an externally defined test


cgwalters:
I'd like to propose something like:

```
$ kola run-ext-script /path/to/binary
```

Takes a single binary at `/path/to/binary`, injects it and a generated basic systemd unit `kola-test-ext-$(basename binary).service` that runs it.  
kola connects via ssh and runs `systemctl --wait is-active` (does that work? if not find another command).

When the service exits successfully the test is considered successful.  When it exits with an EXITCODE error, the test has failed.  If it exits from SIGTERM, then we assume the host is rebooting and continue waiting.

Then we support a much bigger generalization of this,

```
kola run-ext /path/to/dir
```

which injects a full directory for dependencies, and which would have at least a `kola-test.service` unit, and maybe we support a container image?

This would allow us to more easily adapt e.g. the ostree test suite to run inside kola. So in ostree's upstream CI we could use that.  And importantly the reverse works - run the ostree tests downstream too, i.e. reverse dependency testing. e.g. in FCOS pipeline, when the kernel/systemd change, run ostree's tests.

arithx:
The proposal sounds like it has a lot of overlap to what we're already doing with `kolet` so it should be relatively straightforward to implement.

One idea is to also allow some specification of the binaries and whether or not the tests are entirely self-contained to allow for re-use of the same machine which could be valuable when running on non-qemu platforms.

cgwalters:
Another idea here is:

```
kola run-ext-container quay.io/coreos/ostree-tests:latest
```

We could define a simple schema for "test containers" - perhaps two cases:

- run unprivileged with the host filesystem mounted *readonly* at `/host`, can run multiple in parallel/sequence
- run privileged, assumed to be destructive and only one run per VM

PR #1182, opened 2020-02-13
kola/sssd.go: verifies pam configs correctly configured and no regressions are hit

zonggen:
This test verifies changes made in https://gitlab.cee.redhat.com/coreos/redhat-coreos/commit/03ba15aa429af7bf07531e0a533d691dbfd623c4
does not cause regression by checking `altfiles` is not deleted from /etc/nsswitch.conf,
which is the case when calling `authselect select sssd ...` with the `authselect` that comes with
the image.

In addition, also tests that `pam_sss.so` is present in pam config files, e.g.
/etc/pam.d/password-auth and /etc/pam.d/system-auth.

### DIFF
--- a/kola/tests/rhcos/sssd.go
+++ b/kola/tests/rhcos/sssd.go
@@ -1,0 +1,62 @@
+// Copyright 2020 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rhcos
+
+import (
+	"github.com/coreos/mantle/kola/cluster"
+	"github.com/coreos/mantle/kola/register"
+	"github.com/coreos/mantle/platform"
+	"github.com/coreos/mantle/platform/conf"
+)
+
+func init() {
+	register.RegisterTest(&register.Test{
+		Run:         verifySSSD,
+		ClusterSize: 1,
+		Name:        `rhcos.sssd`,
+		Flags:       []register.Flag{},
+		Distros:     []string{"rhcos"},
+		Platforms:   []string{"qemu"},
+		UserData: conf.Ignition(`{
+			"ignition": {
+				"version": "2.2.0"
+			}
+		}`),
+	})
+}
+
+// Verify changes made in https://gitlab.cee.redhat.com/coreos/redhat-coreos/commit/03ba15aa429af7bf07531e0a533d691dbfd623c4
+// do not cause regression by checking /etc/nsswitch.conf contains 'altfiles'.
+// For example, calling `sudo authselect select --force sssd with-custom-passwd with-custom-group` would
+// remove 'altfiles' in /etc/nsswitch.conf, which will fail sssd.service.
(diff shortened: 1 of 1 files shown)


zonggen:
Not entirely sure whether only checking presence of `altfiles` in `/etc/nsswitch.conf`  covers all of the potential regressions that could happen.

cc @jlebon @miabbott

zonggen:
Tested inside cosa container by running
`sudo /mantle/bin/kola run -b rhcos --ignition-version v2  --qemu-image `pwd`/rhcos-44.81.202002131810.0-qemu.x86_64.qcow2 rhcos.nss-altfiles`, passed the current test.

kola/tests/rhcos/nss-altfiles.go
```diff
@@ -0,0 +1,49 @@
...
+	})
+}
+
+// Verify changes made in https://gitlab.cee.redhat.com/coreos/redhat-coreos/commit/03ba15aa429af7bf07531e0a533d691dbfd623c4
+// do not cause regression by checking /etc/nsswitch.conf contains 'altfiles'.
+// For example, calling `sudo authselect select --force sssd with-custom-passwd with-custom-group` would
+// remove 'altfiles' in /etc/nsswitch.conf, which will fail sssd.service.
+func verifyNssAltfiles(c cluster.TestCluster) {
+	m := c.Machines()[0]
+	grepAltfiles := c.MustSSH(m, "cat /etc/nsswitch.conf | grep altfiles")
```
cgwalters: Minor nit...it's more efficient to do directly:
`grep altfiles /etc/nsswitch.conf`

Efficiency is obviously irrelevant here but forming the correct habit here can help when writing code that might process larger files in other places.

kola/tests/rhcos/nss-altfiles.go
```diff
@@ -0,0 +1,49 @@
...
+	if len(grepAltfiles) == 0 {
```
cgwalters: `MustSSH` already panics if the command exits non-zero, which `grep` will do by default if it doesn't find a match.  You might as well use `grep -q altfiles /etc/nsswitch.conf` to clarify we just care about the exit code.

kola/tests/rhcos/nss-altfiles.go
zonggen: Totally agree.. somehow I was so used to `cat file | grep sth` syntax.. Thanks for pointing that out!

zonggen:
> For now: OK, but let's avoid doing too much more of this until we fix the above issue. This could easily be e.g. a systemd-run --property=DynamicUser=yes -- grep -q altfiles /etc/nsswitch.conf for example that would take milliseconds to run inside an existing system nondestructively.

Makes a lot of sense to me. Will update the commit according to comments but I'm totally agree this PR could hold/closed until https://github.com/coreos/mantle/issues/1179 is finalized

kola/tests/rhcos/nss-altfiles.go
zonggen: Updated with -q option and removed stdout checking.

jlebon:
Something else worth checking here is that `/etc/pam.d/password-auth` and `/etc/pam.d/system-auth` have `pam_sss.so`. Though I guess that's not strictly about nss-altfiles. 

Maybe we can rename the test to `rhcos.sssd` instead? Then e.g. even once we switch to systemd-sysusers, we'll still want to make sure the PAM configs are set up for sssd.

zonggen:
Updated with two more test conditions:
- `grep -q pam_sss.so /etc/pam.d/password-auth`
- `grep -q pam_sss.so /etc/pam.d/system-auth`

jlebon:
Looks like the commit message title also needs to be updated?

zonggen:
Yes!

miabbott:
Perhaps as a follow-up or another test, but could we test user login when `sssd` is configured properly?

PR #1184, opened 2020-02-18
kola: Remove special casing of $discovery

cgwalters:
While working on a PR related to Ignition conversion I made
a `$discovery` that this code is only for CL and can go away
now.


bgilbert:
If we're not planning to have any built-in etcd discovery support in kola for FCOS/RHCOS, we should tear out `GetDiscoveryURL`, plus `SimpleEtcd` from `platform/local`.

cgwalters:
I guess etcd tooling could be useful for https://github.com/coreos/airlock testing?  But I dunno...seems like something where we'd have just the airlock tests script setting etcd from ignition rather than doing it via kola?

cgwalters:
OK, recycled :recycle: those bits too.

bgilbert:
Yeah, we're not shipping etcd in the OS image, so I think individual tests should handle it if needed.

bgilbert:
I believe we can also unvendor etcd and drop the loglevel overrides from `cli/cli.go`.

cgwalters:
> I believe we can also unvendor etcd and drop the loglevel overrides from cli/cli.go.

Ah yes, that's a substantial amount of code to delete.  Thanks!

Issue #1160, opened 2020-01-14
`kola list` should filter results according to `--platform` and `--distro`


jlebon:
Expected `kola list -p qemu-unpriv -b rhcos` to show just the results that match those filters, but it looks like it just prints all known tests.

zonggen:
Agreed, I think this would be a convenient feature to have.

zonggen:
Initial PR: https://github.com/coreos/mantle/pull/1185

PR #1185, opened 2020-02-19
kola/list: add show filter for `kola list`

zonggen:
Adds show filter for `kola list`, usage:
 - `kola list --platform qemu --distro rhcos`
 - `kola list -p qemu -b rhcos`

Closes: https://github.com/coreos/mantle/issues/1160

### DIFF
--- a/cmd/kola/kola.go
+++ b/cmd/kola/kola.go
@@ -103,6 +103,8 @@ This can be useful for e.g. serving locally built OSTree repos to qemu.
 	}
 
 	listJSON           bool
+	listPlatform       string
+	listDistro         string
 	httpPort           int
 	findParentImage    bool
 	qemuImageDir       string
@@ -114,6 +116,8 @@ func init() {
 
 	root.AddCommand(cmdList)
 	cmdList.Flags().BoolVar(&listJSON, "json", false, "format output in JSON")
+	cmdList.Flags().StringVarP(&listPlatform, "platform", "p", "all", "filter output by platform")
+	cmdList.Flags().StringVarP(&listDistro, "distro", "b", "all", "filter output by distro")
 
 	root.AddCommand(cmdHttpServer)
 	cmdHttpServer.Flags().IntVarP(&httpPort, "port", "P", 8000, "Listen on provided port")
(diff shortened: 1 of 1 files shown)


cmd/kola/kola.go
```diff
@@ -310,6 +315,21 @@ func runList(cmd *cobra.Command, args []string) {
 		fmt.Fprintln(w, "Test Name\tPlatforms\tArchitectures\tDistributions")
 		fmt.Fprintln(w, "\t")
 		for _, item := range testlist {
+			platforms := strings.Join(item.Platforms, " ")
```
arithx: Couldn't all of this be solved with two loops through `item.Platforms` & `item.Distros`? i.e.
```golang
for _, platform := range item.Platforms {
    if platform == "all" || platform == listPlatform {
        break
    }
}
for _, distro := range item.Distros {
    if distro == "all" || distro == listDistro {
        break
    }
}
fmt.Fprintf(w, "%v\n", item)
```

zonggen: Thank you for reviewing!
Initially I was trying to avoid using for loops so ended up with this awkward readability code. Will add a commit with the above changes.

cmd/kola/kola.go
```diff
@@ -102,6 +104,8 @@ func init() {
 
 	root.AddCommand(cmdList)
 	cmdList.Flags().BoolVar(&listJSON, "json", false, "format output in JSON")
+	cmdList.Flags().StringVarP(&listPlatform, "platform", "p", "all", "show output by platform")
+	cmdList.Flags().StringVarP(&listDistro, "distro", "b", "all", "show output by distro")
```
jlebon: Maybe `s/show/filter/` on those two?

cmd/kola/kola.go
```diff
@@ -310,7 +314,23 @@ func runList(cmd *cobra.Command, args []string) {
 		fmt.Fprintln(w, "Test Name\tPlatforms\tArchitectures\tDistributions")
 		fmt.Fprintln(w, "\t")
 		for _, item := range testlist {
-			fmt.Fprintf(w, "%v\n", item)
+			platformFound := false
+			for _, platform := range item.Platforms {
+				if listPlatform == "all" || platform == "all" || platform == listPlatform {
+					platformFound = true
+					break
```
jlebon: Hmm, I think we can check `listPlatform` upfront to see if it's `all`, right? That way we avoid iterating entirely.

Whenever I encounter nested for-loops like this and bool vars, I always check if it'd be cleaner to use functions. So e.g. here it'd be:

```go
for _, item := range testlist {
  if matchesFilter(item.Platforms, listPlatforms) && matchesFilter(item.Distros, listDistro) {
    fmt.Fprintf(...)
  }
}
```

That way, the functions can just `return true` when found and plus you dedupe the loop logic (and plus we don't try to filter by distro if the platform didn't match anyway).

arithx: That's fair, if only golang had a native slice.Contains method 😄

cmd/kola/kola.go
zonggen: Sure, updating..

cmd/kola/kola.go
zonggen: Updated with pre-checking `listPlatform` and `listDistro` :)

cgwalters: > That's fair, if only golang had a native slice.Contains method smile

And of course...generics and... :wink:

zonggen:
@cgwalters yes I do, squashing and merging

PR #1186, opened 2020-02-19
kola: Translate from Ignition spec v3 to v2 as needed

cgwalters:
Duplicating Ignition spec 3 and 2 code across our tests is
a huge demotivator for writing tests and is prone to error, etc.

Add the https://github.com/coreos/ign-converter code and
use it in the case where we're booted to an OS requesting spec 2,
and we only have spec3 in the test.

This is prep for working on install tests, where I didn't
want to write Ignition configs twice.


cgwalters:
Also forgot to mention I only deleted the duplicate spec 2 from [one test](https://github.com/coreos/mantle/pull/1186/files#diff-a1315e0e4bd52ed25fffafa768b0be28) but if agreed to merge this I'll go through and delete the rest too!

platform/cluster.go
```diff
@@ -151,7 +151,7 @@ func (bc *BaseCluster) RenderUserData(userdata *platformConf.UserData, ignitionV
 		}
 	}
 
-	conf, err := userdata.Render(bc.bf.ctPlatform)
+	conf, err := userdata.Render(bc.bf.ctPlatform, bc.IgnitionVersion() == "v2")
```
arithx: We should probably keep track of this in the future and make sure that if we start repeating this we provide a cleaner interface for determining if a cluster is expecting Ignition v2. Fine punting that to a later PR though.

cgwalters:
/hold
I'd tested this w/FCOS but not all of the RHCOS tests, doing that now.

cgwalters:
/hold cancel
It works fine as is, I just got tripped up by new kola tests and an old local RHCOS build.

PR #1187, opened 2020-02-21
cosa: use generated structs from schema

darkmuggle:
Use the canonical schema from COSA.


darkmuggle:
Dependant PR https://github.com/coreos/coreos-assembler/pull/1147

Throwing this out for early review before I get too far into this.

cgwalters:
Oh neat, I didn't know about schematyper.  
/approve

darkmuggle:
Okay, the latest push works. 

@arithx WDDYT about this?

generate
```diff
@@ -0,0 +1,24 @@
...
+hash schematyper 2>/dev/null || {
+    echo >&2 "schematyper must be installed and in the PATH to use this script: https://github.com/idubinskiy/schematyper"
+    exit 1
+}
+
+schema_version="v1"
+schema_commit="08725e679fbc3b923a8f7196a9ac505263db7c07"
+schema_url="https://raw.githubusercontent.com/coreos/coreos-assembler/${schema_commit}/src/schema/${schema_version}.json"
+echo "Generating COSA Schema ${schema_version} ${schema_commit}"
+curl -sL "${schema_url}" -o "cosa/${schema_version}.json"
```
arithx: Hmm, I wonder if there's value in just leaving the act of fetching the schema from the COSA github to a manual process rather than fetching it automatically each run of `generate`, especially given we're including the file in this repository (`cosa/v1.json`).

darkmuggle: Or, the question I've been asking -- and I have an answer that I like, per se -- do we:
- leave the schema, and generated GoLang in COSA itself
- vendor it in, ala `go get github.com/coreos/coreos-assembler/schema`

arithx: Either would work for me. At that point it's more of a matter of if we want to carry golang code in COSA.

jlebon: What makes this more confusing though is that mantle itself is a submodule of cosa. So we can end up in a situation where the baked mantle is using a different schema than the rest of cosa. (Of course, all this just hints at what we talking about re. either merging the two together, or alternatively separating them better.)

Hmm, maybe what we actually want here is to do all this at the `./build` step? E.g. `./build` in this repo defaults to using the schema from git master cosa, but when built as a submodule in cosa, we can pass the file to the schema to translate so that we're guaranteed they're always in sync.

cgwalters: Right, https://github.com/coreos/coreos-assembler/issues/163 and I also added this as an item for realtime cabal discussion tomorrow.

darkmuggle: I think the cabal topic is a marvelous idea.

cgwalters: We don't need to block this on that though IMO, it's just one of a number of things that would be changed/tweaked after a merge, and if we don't merge it's still worth it.

PR #1188, opened 2020-02-24
kola: Rework syncOptions

cgwalters:
Prep for adding more commands which need `syncOptions`; rather
than having `syncOptions` and `syncCosaOptions`, make the upgrade
test the exception that skips the latter.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -142,7 +142,7 @@ func init() {
 }
 
 // Sync up the command line options if there is dependency
-func syncOptions() error {
+func syncOptionsImpl(useCosa bool) error {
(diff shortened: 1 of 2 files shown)


cmd/kola/options.go
```diff
@@ -169,10 +169,8 @@ func syncOptions() error {
 		return err
 	}
 
-	if kola.Options.CosaBuild != "" {
-		var err error
-		kola.CosaBuild, err = cosa.ParseBuild(kola.Options.CosaBuild)
-		if err != nil {
+	if kola.Options.CosaBuild != "" && useCosa {
```
arithx: Maybe better to switch the order of the conditionals?

PR #1189, opened 2020-02-24
kola: Add helper to guess Ignition version from OS

cgwalters:
And use it if no version is specified, and we're
testing qemu and have a disk image path.

Part of draining bits of cosa into mantle so they can
be reused for more testing.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -215,5 +215,11 @@ func syncCosaOptions() error {
 		}
 	}
 
+	if kola.Options.IgnitionVersion == "" {
+		if kola.QEMUOptions.DiskImage != "" {
+			kola.Options.IgnitionVersion = sdk.TargetIgnitionVersionFromName(kola.QEMUOptions.DiskImage)
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


cmd/kola/options.go
```diff
@@ -215,5 +215,11 @@ func syncCosaOptions() error {
 		}
 	}
 
+	if kola.Options.IgnitionVersion == "" {
```
arithx: With how the code is currently laid out `kola.Options.IgnitionVersion` will never be an empty string when entering this function. It's auto-set to something in [`syncOptions`](https://github.com/coreos/mantle/blob/master/cmd/kola/options.go#L193-L199) based on the distribution & [`syncCosaOptions` is only called after `syncOptions`](https://github.com/coreos/mantle/blob/master/cmd/kola/kola.go#L119-L122). Should this PR be on top of #1188?

cgwalters: Yes, it is intended to be on top of #1188 - I split it out for clarity.  But I've now merged #1188 so this code is now correct, right?

PR #1190, opened 2020-02-24
kola: Add ign-convert2

cgwalters:
Intending to replace `coreos-assembler/src/incomplete-hack-ign-3to2`.

### DIFF
--- a/cmd/kola/kola.go
+++ b/cmd/kola/kola.go
@@ -26,6 +26,8 @@ import (
 	"strings"
 	"text/tabwriter"
 
+	ignconverter "github.com/coreos/ign-converter"
+	ignv3 "github.com/coreos/ignition/v2/config/v3_0"
 	"github.com/coreos/pkg/capnslog"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


arithx:
Is this just to have it in mantle so cosa doesn't have to vendor the `ign-converter` binary? It seems like that already covers the case of both up & down translating configs.

cgwalters:
This whole area is obviously a huge mess.  One question here is - are we intending to make `/usr/bin/ign-converter` a binary exposed in cosa, or even to users?  Or is it more a feature of `fcct`?  And in fact we're debating [fcct functionality](https://github.com/coreos/fcct/issues/76) in general still too.

This PR is just a tactical shortcut around all of that basically.

cgwalters:
Or to say it another way, the goal of this PR is:

> Intending to replace coreos-assembler/src/incomplete-hack-ign-3to2.

and not much else - so we'll at least delete code.

(And BTW this exposed a bug in ign-converter, `cosa run` fails around `overwrite` semantics for `/etc/motd` for me with this, need to dig into that)

If we vendor ign-converter in coreos-assembler too then we'll have *two* vendors of the code (well in addition to machine-config-operator), until of course we merge cosa and mantle (hey look that came up again!)

PR #1192, opened 2020-02-25
kola: Add testiso command

cgwalters:
This currently just tests the legacy installer.  I translated
the existing shell script into Go.  The next steps here
are to clean this up and refactor the code so it can test both
installation paths.

### DIFF
--- a/cmd/kola/testiso.go
+++ b/cmd/kola/testiso.go
@@ -1,0 +1,326 @@
+// Copyright 2015-2018 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/coreos/mantle/util"
+	"github.com/pkg/errors"
+
+	"github.com/coreos/mantle/sdk"
+
+	ignconverter "github.com/coreos/ign-converter"
+	ignv3types "github.com/coreos/ignition/v2/config/v3_0/types"
+	"github.com/spf13/cobra"
+
+	"github.com/coreos/mantle/kola"
+	"github.com/coreos/mantle/platform"
+)
+
+var (
+	cmdTestIso = &cobra.Command{
+		Run:    runTestIso,
+		PreRun: preRun,
+		Use:    "testiso",
+		Short:  "Test a CoreOS PXE boot or ISO install path",
+	}
+	// TODO expose this as an API that can be used by cosa too
+	consoleKernelArgument = map[string]string{
+		"amd64-usr":   "ttyS0",
+		"ppc64le-usr": "hvc0",
+		"arm64-usr":   "ttyAMA0",
+		"s390x-usr":   "ttysclp0",
+	}
+)
+
+func init() {
+	root.AddCommand(cmdTestIso)
+}
+
+func runTestIso(cmd *cobra.Command, args []string) {
+	if err := doTestIso(cmd, args); err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		os.Exit(1)
+	}
+}
+
+func doTestIso(cmd *cobra.Command, args []string) error {
+	//var err error
+
+	if kola.CosaBuild == nil {
+		return fmt.Errorf("Must provide --cosa-build")
+	}
+
+	if kola.CosaBuild.BuildArtifacts.Kernel.Path != "" {
+		return errors.Wrapf(testLegacyInstaller(), "testing legacy installer")
+	} else {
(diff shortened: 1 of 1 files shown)


cgwalters:
Tested w/RHCOS 4.4 on x86_64.  I didn't test the other architecture code and it's somewhat likely to have regressed.

cmd/kola/testiso.go
```diff
@@ -0,0 +1,314 @@
...
+	kargsStr := strings.Join(kargs, " ")
+
+	var bootfile string
+	switch boottype {
+	case "pxe":
+		pxeconfigdir := filepath.Join(tftpdir, "pxelinux.cfg")
+		if err := os.Mkdir(pxeconfigdir, 0777); err != nil {
+			return err
+		}
+		pxeimages := []string{"pxelinux.0", "ldlinux.c32"}
```
Prashanth684: for s390, the pxelinux image has to be built from the kernel and initrd.
ref:https://github.com/coreos/coreos-assembler/pull/1139/commits/074e32836bdb108e158efdc2aa066ab8dd3fe8c9#diff-4cf8ac28258bdbad16564f473da21a71R104

Prashanth684: and the pxelinux parameters are passed a little differently which is annoying. 

@cgwalters since i will be testing s390x and ppc anyway i can make those tweaks if necessary

Prashanth684:
Thanks @cgwalters ! I will test it on ppc64le and s390x

cmd/kola/testiso.go
cgwalters: OK yep I forgot to translate that part.  Let me hold a mutex on this for a bit and try to continue translation.

cgwalters: OK mutex :lock: released - do you want to just post a `git diff` from any changes you need here?

(Do we have any docs on getting CoreOS maintainers access to ppc64le/s390x kube/OpenShift clusters set up to run cosa?)

cgwalters: I went ahead and added you to [Co-authored-by](https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-github).

cmd/kola/testiso.go
```diff
@@ -0,0 +1,327 @@
+// Copyright 2015-2018 CoreOS, Inc.
```
arithx: date / company

cmd/kola/testiso.go
```diff
@@ -0,0 +1,327 @@
...
+
+func runTestIso(cmd *cobra.Command, args []string) {
+	if err := doTestIso(cmd, args); err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		os.Exit(1)
+	}
+}
+
+func doTestIso(cmd *cobra.Command, args []string) error {
+	//var err error
```
arithx: Can drop this

cmd/kola/testiso.go
```diff
@@ -0,0 +1,327 @@
...
+		boottype = "grub"
+		networkdevice = "virtio-net-pci"
+		break
+	case "s390x-usr":
+		boottype = "pxe"
+		networkdevice = "virtio-net-ccw"
+		tftpipaddr = "10.0.2.2"
+		bootindex = "1"
+	default:
+		return fmt.Errorf("Unsupported arch %s", kola.QEMUOptions.Board)
```
arithx: `arm64` is in the `consoleKernelArgument` map but would just error out. Should that just be dropped?

cmd/kola/testiso.go
Prashanth684: afaik there is just the one cluster for ppc64le/s390x which runs the rhcos jenkins pipeline. let me see if i can arrange something on this.

cmd/kola/testiso.go
```diff
@@ -0,0 +1,327 @@
...
+	defer os.RemoveAll(tftpdir)
+
+	if err := ioutil.WriteFile(filepath.Join(tftpdir, "config.ign"), serializedConfig, 0777); err != nil {
+		return err
+	}
+
+	metalIsCompressed := !strings.HasSuffix(metalimg, ".raw")
+	metalname := metalimg
+	if !metalIsCompressed {
+		fmt.Println("Compressing metal image")
```
arithx: For the `Println` log statements might be worthwhile importing `plog` & using that so it can be easily suppressed.

cmd/kola/testiso.go
Prashanth684: thanks. tested on power and x390. looks good. on s390 needed a minor change:
https://gist.github.com/Prashanth684/5e49fd514471a58e21b739f48535d26e

Prashanth684: also we definitely can give you guys access to the ppc64le/s390x pipeline systems. Please get in touch with Andy McCrae ..he should be able to help out

cgwalters:
I'm going to hold this one until we land https://github.com/coreos/coreos-assembler/pull/1152

cmd/kola/testiso.go
cgwalters: I kept it for completeness when copying the code from cosa.  I plan to de-dup that code soon, and we want to support arm64 anyways in the future right?

Prashanth684: right...i did look into arm...should be a matter of pointing to the correct boot images so definitely can be supported.. when  i get some time i'll do that

cmd/kola/testiso.go
cgwalters: I looked at this and spent 5 minutes banging my head against understanding the log levels and details like whether `\n` would be automatically emitted or not but...the bigger problem here is that we output the qemu console by default, and we kind of need to at the moment because we don't have the journal handling hooked up.

So...that can come next?

cgwalters:
Moved to https://github.com/coreos/coreos-assembler/pull/1168

PR #1193, opened 2020-02-26
qemu: Only set serial= if not already provided

cgwalters:
Regression from 73831285bf5f85af95bb202f0349545f45ac278f which
broke `coreos.ignition.mount.disks` because we were overriding
the provided serial.


cgwalters:
This is what's blocking https://github.com/coreos/coreos-assembler/pull/1152

PR #1194, opened 2020-02-26
kola: Use uuid instead of partlabel for ignition.mount disk test

Prashanth684:
Changed from by-partlabel to by-partuuid by assigning a random uuid. On s390x, the by-partlabel string is incorrect because GPT partitioning uses little endian by default and s390x is big endian. Refer
to: https://github.com/coreos/ignition-dracut/pull/100#issuecomment-529948403.

### DIFF
--- a/kola/tests/ignition/mount.go
+++ b/kola/tests/ignition/mount.go
@@ -39,7 +39,8 @@ var (
 						{
 							"label": "CONTR",
 							"size": 1048576,
-							"start": 0
+							"start": 0,
+							"guid": "63194b49-e4b7-43f9-9a8b-df0fd8279bb7"
 						}
 					]
 				},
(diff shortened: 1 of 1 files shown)


cgwalters:
Mind matching the commit message style here?  Use a "topic prefix" (and a shorter commit message title)  Something like:

> kola: Use uuid instead of partlabel for ignition.mount tests

Otherwise...it does seem to me that udev should be fixed to byteswap, though in theory someone could be depending on the non-byteswapped names, but I guess we could add symlinks to both?

Anyways commit message aside, seems OK to me.

Prashanth684:
> Mind matching the commit message style here? Use a "topic prefix" (and a shorter commit message title) Something like:
> 
> > kola: Use uuid instead of partlabel for ignition.mount tests
> 
> Otherwise...it does seem to me that udev should be fixed to byteswap, though in theory someone could be depending on the non-byteswapped names, but I guess we could add symlinks to both?
> 
> Anyways commit message aside, seems OK to me.

done. yes it does seem like udev needs to be fixed for s390x - can't think of who could be using partlabel for s390x because it's broken. But it would be good to get feedback from the IBM team and maybe open an issue with udev for this.

PR #1195, opened 2020-02-27
kola: Fix syncOptions

cgwalters:
Regression from d4b252bb7c8b77e73ad6926de70921e95eeb95e5
We need to keep parsing the cosa build, just not actually use it
to sync the target options in order to match how the upgrade
test works.

### DIFF
--- a/cmd/kola/options.go
+++ b/cmd/kola/options.go
@@ -143,6 +143,8 @@ func init() {
 
 // Sync up the command line options if there is dependency
 func syncOptionsImpl(useCosa bool) error {
+	var err error
+
 	kola.PacketOptions.Board = kola.QEMUOptions.Board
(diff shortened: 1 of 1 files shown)


cmd/kola/options.go
```diff
@@ -169,6 +171,11 @@ func syncOptionsImpl(useCosa bool) error {
 		return err
 	}
 
+	kola.CosaBuild, err = cosa.ParseBuild(kola.Options.CosaBuild)
```
jlebon: Shouldn't we check first here if `kola.Options.CosaBuild` is not empty?

cgwalters: Fixed, thanks!  (Imagine if golang had had `Option<T>`...)

